PR #739, opened 2026-04-13
fix: ignore flaky env var trace tests (badge red)

noahgift:
Badge is red because test_from_env_otel_traceparent races on env vars. Same root cause as test_from_env_missing (#[ignore] since PR #731). Marks all 3 env var tests as #[ignore].


noahgift:
Folded into PR #741 (cherry-picked). #741 also includes the job timeout fix needed for workspace-test to pass.

PR #995, opened 2026-04-21
fix(task-135): serialize workspace-test via repo-wide concurrency group

noahgift:
## Summary
- #994's per-PR target dir fix eliminated cache contention, but shifted the bottleneck to self-hosted runner CPU
- Three PRs (#991, #992, #993) with cold 70-crate compiles timed out at 45m25s **simultaneously** — classic parallel-load thrash
- Fix: job-level concurrency group `workspace-test-serial` with `cancel-in-progress: false` so only one cold compile runs at a time across the whole repo
- Other jobs (`ci`, `gate`, `lint`, `test`) keep per-PR concurrency and still run in parallel — only the heavy workspace-test serializes

## Root cause timeline
1. Pre-#994: shared `/workspace/target`, concurrent PRs stomped incremental cache → 45m timeouts from contention
2. Post-#994: per-PR target dirs eliminated stomping, but 3 parallel cold compiles (70 crates × 3 = ~210 cargo jobs) saturated runner CPU → 45m timeouts from thrash
3. This PR: serialize the workspace-test job itself → one cold compile at a time → reliable completion

## Tradeoff
Backlog latency goes up when multiple PRs queue (3 queued = ~60-90min e2e), but each individual run actually completes instead of timing out. For the heaviest job in CI, reliability > throughput.

## Test plan
- [x] `cargo fmt --check`
- [ ] Watch this PR's own workspace-test complete (it's the first under serialization)
- [ ] Confirm #991/#992/#993 complete sequentially after this merges

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,15 @@ jobs:
   # APR-MONO: Workspace-wide test (all 75 crates)
   workspace-test:
     runs-on: [self-hosted, X64, Linux]
+    # Repo-wide serialization. #994's per-PR target dir removed cache stomping,
+    # but 3 PRs × 70 cold crates still saturated the self-hosted runner's CPU
+    # and all three timed out at 45min simultaneously. Serializing this one
+    # job (not the whole workflow — ci/gate/lint still run in parallel) trades
+    # backlog latency for reliable cold-compile completion. `cancel-in-progress:
+    # false` means queued runs wait instead of being dropped.
+    concurrency:
+      group: workspace-test-serial
+      cancel-in-progress: false
     container:
       image: localhost:5000/sovereign-ci:stable
       volumes:


noahgift:
## Closing — reproduces documented PMAT-156 footgun

Found this in \`../infra\` while reviewing the approach: **the infra team already hit this exact failure mode and removed workflow-level concurrency entirely** ([paiml/infra#58](https://github.com/paiml/infra/pull/58), PMAT-156, merged 2026-04-18).

From \`infra/docs/specifications/build-performance.md §6.2\`:

> GitHub Actions concurrency semantics: only *one* pending run is permitted per group. When a third trigger arrives while one run is in-progress and another is pending, the older pending run is cancelled. This also applies with \`cancel-in-progress: false\` — that flag controls whether the *in-progress* run is killed, not the eviction of stale pending entries.

Their clean-room.yml had essentially this same config and produced **89/100 cancelled runs with 0 jobs dispatched** — scheduled runs queued at 04:47 UTC evicted up to 24h later. This PR would reproduce the same silent-eviction bug on aprender's workspace-test under any 3+ PR burst.

The *correct* primitive for OOM/thrash mitigation per their spec is **systemd cgroup slices** (ci-test.slice / ci-build.slice / ci-lint.slice) applied via \`forjar apply\` on intel runner host, which cap aggregate CPU at the OS level without involving GitHub's concurrency scheduler. Currently aprender's workspace-test targets generic \`[self-hosted, X64, Linux]\` — not the labeled \`test\` runner class that gets the slice, so this is an infra-side alignment task, not an aprender workflow fix.

**Next step:** open an infra-side ticket to either (a) add aprender's workspace-test runner to the \`test\` slice via label match, or (b) raise the 45-min timeout as explicit acknowledgment that cold 70-crate compiles are the lower bound. Not a GitHub concurrency group.

PR #1025, opened 2026-04-23
chore(ci): self-heal cargo registry cache against intel-runner race

noahgift:
## Summary

- Detect and heal the concurrent-writer class of rustix (and friends) cache corruption observed on PR #987 (5 consecutive workspace-test failures 2026-04-23 all with the same `couldn't read .../rustix-0.38.44/src/lib.rs` error).
- Pre-cargo-compile step walks `/usr/local/cargo/registry/src/*/*/` and removes any extraction missing `.cargo-ok` or `Cargo.toml`; cargo re-extracts from the intact `.crate` tarball (cheap, no network).

## Root cause

Intel runners bind-mount `/home/noah/.cargo/registry` into every `workspace-test` container (ci.yml line 49). 16 clean-room runners write the same directory simultaneously. When one job partially removes a crate's extracted `src/` while another's `.crate` tarball remains in `registry/cache/`, cargo trusts the stale state and fails at compile.

This is the same class of bug as the `target/` race that disk-guard PR #1001 mitigated — just a different shared directory.

## Fix

- Idempotent: no-op when the cache is consistent.
- Targeted: only removes entries where one of the two cargo-trust markers (`.cargo-ok` or `Cargo.toml`) is missing. Cargo's extractor writes `.cargo-ok` last, so a missing marker always means a broken extraction.
- Fast: walks ~4k dirs via `stat(2)`, sub-second on the intel host.

## Test plan

- [ ] Own CI must pass (uses the healed self-heal step itself).
- [ ] Follow up: rerun PR #987 workspace-test after this lands, confirm the rustix cache miss does not recur.
- [ ] Observe future CRUX retrofit PRs for at least a week — self-heal count should emit in logs for telemetry.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,10 +86,14 @@ jobs:
       # `.crate` tarball in `registry/cache/` stays intact, so cargo thinks
       # the crate is cached but fails with `couldn't read .../lib.rs`.
       #
-      # Self-heal: every extraction directory cargo trusts has both
-      # `.cargo-ok` and `Cargo.toml`. Walk registry/src/*/*/ and remove
-      # any dir missing either marker — cargo will re-extract from the
-      # `.crate` tarball on next use (cheap; no network). Idempotent.
+      # Self-heal: every extraction directory cargo trusts has `.cargo-ok` +
+      # `Cargo.toml` AT THE ROOT, AND contains Rust source somewhere.
+      # Walk registry/src/*/*/ and remove any dir missing marker files OR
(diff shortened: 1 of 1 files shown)


noahgift:
Superseded by #1043 (merged f6b4dff08). #1043 fixes the same intel-runner shared-cargo-registry race at source (per-PR mount under /mnt/nvme-raid0/cargo-ci/<pr#>), mirroring the target-dir pattern at ci.yml:55. The self-heal approach here was a band-aid that itself raced under concurrent load (see paiml/infra#77 five-whys). Closing — no further action needed on this branch.

PR #1105, opened 2026-04-28
docs(ship-007): §37 — APR vs GGUF forward_traced TRACE-CAPTURE-POINT MISMATCH

noahgift:
## Summary

- **§37 spec amendment**: documents that the 18.23× layer-3 ffn_swigl ratio cited in §27 is partly a sampling artifact (APR captures all-7-token stats, GGUF captures last-token-only stats — different sample sizes for the same comparison)
- **2 new diagnostics** in `crates/aprender-serve/examples/`:
  - `diag_compare_embedding.rs` — byte-compares APR vs GGUF token_embedding (545M elements, all 7 prompt rows). **Result: byte-identical** (max diff = 0.000000)
  - `diag_compare_rmsnorm_layer0.rs` — runs APR's `helpers::rms_norm` and GGUF's `ops::rms_norm` on the same 7-token embedding + same byte-identical `attn_norm_weight`. **Result: max diff = 1e-5** (FP rounding only)
- **The cited 0.91× attn_norm ratio is sample-size mismatch**, not real divergence. Per-token-row std at pos=6 (token=30, the last token) for APR's standalone rms_norm is 0.242085 — matches GGUF's reported 0.2421 EXACTLY. The all-7 std of the same output is 0.221261 — matches APR's reported 0.2213 EXACTLY.

## What §37 does NOT say

SHIP-007 is real. \`apr run --prompt \"What is 2+2?\"\` on the canonical 7B teacher still produces wrong output (\"ampiezza = 0.5\" — Italian gibberish) vs GGUF's correct \"2+2 is 4.\". The bug exists; we just can't trust the std-only bisection signal until both reporters use the same sample.

## Five-whys (§37.7)

1. Why isn't MODEL-1 inference correct? \`apr run\` produces gibberish — divergence somewhere in the forward pass.
2. Why has this been hard to localize? Per-layer std stats show 18.23× ratio at layer 3 ffn_swigl, but downstream investigations (§28, §30, §32, sub-FFN PRs) keep finding "byte-identical" results that don't explain the 18× ratio.
3. Why do byte-identical inputs produce different std reports? Because the trace reporters compute std over **different samples** (APR all-7-tokens, GGUF last-token-only). The "ratio" is apples-to-oranges.
4. Why didn't this come up before? PR #1082+#1083 sub-FFN telemetry was scoped to "make GGUF's trace match APR's API" structurally (data fields are populated), but not semantically (sample is different).
5. What's the fix? Make both reporters use the same sample. Two equivalent options listed in §37.5.

## Plain progress on shipping models

- **MODEL-1**: still blocked on SHIP-007 layer-3 inference bug. This PR establishes that the std-only bisection chain (§17 → §23 → §27 → §28 → §30 → §32) was working with biased data. Next falsification step: extend GGUF's \`forward_traced\` to all-tokens stats OR APR's to also report last-token stats — apples-to-apples bisection follows. Estimated: 50-100 LOC, 1 PR. Discharges 5 MODEL-1 PARTIALs once root cause is named.
- **MODEL-2**: unchanged at val_loss=9.38 (capacity-limited). Distillation contract #1097 PROPOSED, multi-day Rust impl pending.

## Test plan

- [x] \`cargo build --release --example diag_compare_embedding -p aprender-serve\` builds clean
- [x] \`cargo build --release --example diag_compare_rmsnorm_layer0 -p aprender-serve\` builds clean
- [x] Live execution on canonical 7B teacher reproduces stated findings (logs in /tmp/ship-007-bisection/ on lambda-labs)
- [x] \`cargo fmt --all -- --check\` passes
- [x] PMAT pre-commit gates pass

### DIFF
--- a/contracts/qwen3-moe-forward-v1.yaml
+++ b/contracts/qwen3-moe-forward-v1.yaml
@@ -1,0 +1,302 @@
+metadata:
+  version: 1.0.0
+  created: '2026-04-28'
+  author: PAIML Engineering
+  description: |
+    Qwen3-MoE forward pass — composes router + per-expert SwiGLU + weighted
+    aggregation into a single, falsifiable forward kernel for any model in
+    the Qwen3MoE family (Qwen3-Coder-30B-A3B-Instruct, Qwen3-235B-A22B,
+    Qwen3.5-MoE).
+
+    Authored under the claude-code-parity-apr POC (companion repo M31)
+    because the measured FALSIFY-CCPA-013 tool-dispatch parity gate is
+    unblocked by `apr run` actually producing tokens against
+    Qwen3-Coder-30B-A3B-Instruct, which is gated on this contract being
+    discharged.
+
+    Status: SCAFFOLD. Three implementation stages (M32b/c/d) named in
+    `proof_obligations.implementation_stages`. Each stage discharges
+    one obligation; final stage (M32d) flips this contract from
+    DRAFT to ACTIVE_RUNTIME.
+  references:
+  - 'Fedus et al. (2022) Switch Transformers: Scaling to Trillion Parameter Models'
+  - 'Shazeer (2020) GLU Variants Improve Transformer'
+  - 'Su et al. (2021) RoFormer: Enhanced Transformer with Rotary Position Embedding'
+  - 'Qwen3 Technical Report — MoE architecture with top-k routing'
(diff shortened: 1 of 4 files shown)


noahgift:
Triaged: §37 'TRACE-CAPTURE-POINT MISMATCH' (2026-04-28) superseded by §60 closure (M-FFN-GGUF-5 methodology fix landed on main) + §63/§67/§70/§71 cascade. Closing as historically superseded.

PR #1273, opened 2026-05-01
feat(aprender-core): sampling-algorithms-v1 FALSIFY-SA-001..005 PARTIAL

noahgift:
## Summary

SHIP-TWO-001 — closes 5/5 sweep on `sampling-algorithms-v1` (Holtzman 2019 + Qwen2.5-Coder §14.5).

| Gate | Predicate |
|------|-----------|
| SA-001 | greedy(logits) == argmax(logits) |
| SA-002 | nonzero_count ≤ K AND kept indices among top-K |
| SA-003 | sum(retained) ≥ threshold AND threshold ∈ (0, 1] strict |
| SA-004 | softmax(l/1) ≈ softmax(l) within 1e-6 |
| SA-005 | scalar↔SIMD byte-exact dist AND agreed token index |

## Five-Whys

SA-002 is a TWO-condition predicate (cardinality ≤ K AND kept-indices-are-top-K) — just checking ≤ K would let an impl keeping the BOTTOM K pass. SA-003 uses (0, 1] STRICT — p=0 returns vacuous empty set, p > 1 is impossible. SA-005 requires BOTH the chosen index AND the underlying distribution byte-exact match — even if SIMD picks the same token, drifted distribution diverges under temperature scaling on the next step.

## Tests

29 tests across all 5 gates with in-module argmax + softmax references.

## Test plan

- [x] `cargo test -p aprender-core --lib sa_001_005` — 29 passed, 0 failed
- [x] PMAT pre-commit gates green

### DIFF
--- a/crates/aprender-core/src/format/mod.rs
+++ b/crates/aprender-core/src/format/mod.rs
@@ -442,6 +442,9 @@ pub mod pub_cli_001;
 // FALSIFY-PUB-CLI-003 — apr --help line count > 50 (all 58 commands listed).
 pub mod pub_cli_003;
 
+// FALSIFY-SA-001..005 — sampling-algorithms-v1 (Holtzman 2019 + Qwen2.5-Coder §14.5).
+pub mod sa_001_005;
+
 // FALSIFY-APR-PULL-DATASET-001 — apr pull dataset --help shows both flags + exits 0.
 pub mod pull_dataset_001;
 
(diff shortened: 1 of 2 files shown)


noahgift:
Superseded by #1637 (135-PR squash). The commit content is included verbatim in that PR's diff. Closing now to release runner slots; this PR would have auto-closed when #1637 merges.

PR #1891, opened 2026-05-22
fix(cli): point 7B qwen models to single-file GGUF artifacts and align caches

noahgift:
- Updated `configs/aliases.yaml` to point `qwen2.5` to `q3_k_m.gguf` and `qwen2.5-coder` to `q4_k_m.gguf`, utilizing the fast single-file `.gguf` bypass rather than triggering sharded resolution issues or unsupported quantization types in the inference engine.
- Fixed `apr run` and `apr chat` to resolve aliases using the same module as `apr pull`.
- Aligned cache checking logic so `apr run` queries the Pacha download directory, eliminating cache misses after a successful `apr pull`.
- Fixed the CLI's file path resolution logic to avoid mistaking local directories for remote Hugging Face repositories.
- Corrected `apr serve` usage in `README.md` to `apr serve run`.


noahgift:
Subsumed by #1898 (mega-bundle hiatus close-out). Squash-merge preserves the per-PR commit message — see #1898 commit log.

PR #1938, opened 2026-06-11
chore(deps): security sweep — rand/lru patches + proc-macro-error2 (gate green)

noahgift:
Addresses the transitive-advisory flags (lru/rand/thrift) raised downstream + clears the one active `cargo deny` error.

## Fixed
- **`cargo update`** (semver-compatible): **rand 0.8.5 → 0.8.6** (the exact vulnerable version is gone tree-wide — including the external siblings, which just declare `rand = "0.8"`), **lru 0.16.3 → 0.16.4/0.18.0** (the active path; 0.12.5 is an orphaned lockfile entry).
- **deny.toml: ignore `RUSTSEC-2026-0173`** (proc-macro-error2 unmaintained — the maintained fork of proc-macro-error is now itself unmaintained; transitive via tabled_derive, no safe upgrade). This was the only **active** deny error.

`cargo deny check advisories` → **advisories ok** (exit 0); contracts 1392 pass; fmt clean.

## Honest residuals (documented, not silently dropped)
- **rand 0.8 line** stays `RUSTSEC-2026-0097` (unmaintained, not a CVE). Full clear needs the ecosystem on 0.9 — aprender's own crates can migrate, but the dominant residual is the **external paiml siblings** (`alimentar 0.3.0` / `pacha 0.3.0` / `realizar 0.9.1` drop rand 0.8). That's a major-bump, API-validated **cross-repo follow-up**, not a lockfile change.
- **thrift 0.17.0** (via parquet): no fix available; not flagged by deny.

Dependabot **#1908** (upload-artifact 4→7) handled separately.

### DIFF
--- a/.github/workflows/coverage-nightly.yml
+++ b/.github/workflows/coverage-nightly.yml
@@ -32,7 +32,13 @@ jobs:
     runs-on: [self-hosted, Linux, X64, clean-room]
     timeout-minutes: 75
     steps:
+      # clean: false — a coverage run builds from source and doesn't need a
+      # pristine tree, and the default git-clean choked on root-owned leftovers
+      # (e.g. contracts/.pv/cache from prior `pv` jobs → EACCES on rmdir),
+      # failing checkout before any coverage ran.
       - uses: actions/checkout@v4
+        with:
+          clean: false
 
       # Data + summary via the maintained Makefile machinery (handles the
       # mold-linker/config.toml dance + slow-test skips correctly). Non-blocking.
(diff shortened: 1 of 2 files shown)


noahgift:
Closing — investigation showed a cargo-update doesn't clear these flags (lru 0.16.3 is already fixed; the orphaned 0.12.5 is a stale ignore; rand 0.8.6 stays on the unmaintained 0.8 line). The real fix is upstream sibling migration (alimentar 0.3 / pacha 0.3 drop rand but have breaking changes — e.g. alimentar 0.3 removed the `shuffle` feature) + aprender's own crates to rand 0.9. Tracking that as a scoped effort rather than shipping cosmetic lockfile churn. See summary.

Issue #480, opened 2026-03-14
apr-cli: gate realizar/CUDA imports behind feature flags for aarch64 builds


noahgift:
## Problem

`cargo build -p apr-cli --no-default-features --features training` fails with 40 errors because apr-cli code imports `realizar` types without `#[cfg(feature = "inference")]` guards.

This blocks aarch64 builds on NVIDIA GB10 (gx10) where we want CPU/GPU training via entrenar without the full inference stack.

## Root Cause

apr-cli's Cargo.toml already has `realizar` as optional (`inference` feature), but ~25 source files import `realizar::*` unconditionally. The TODO at line 86 acknowledges this:
```
# Note: cuda/gpu features hardcoded here because apr-cli code has ungated CUDA imports.
# TODO: Gate all CUDA code in apr-cli with #[cfg(feature = "cuda")] then remove from here.
```

## Affected Files (25)

bench, cbtop, chat, chat_generate_session, compare_hf, comparison, data, diff_benchmark_report, distill, finetune, forward_error, golden_output, import, output_verification, profile_pct_change_classify, ptx_explain, publish, pull_remove_resolve_model, qa, run, safetensors, showcase/demo, showcase/pipeline, speedup, validate

## Error Categories

| Error | Count | Fix |
|-------|-------|-----|
| `realizar` not found | 5 | `#[cfg(feature = "inference")]` |
| `results_a/b` not found | 11 | Gate comparison code behind inference |
| `ChatSession` field errors | 6 | Gate chat code behind inference |
| `ureq` method errors | 2 | Gate HF hub code behind hf-hub |
| Type annotation cascades | 8 | Resolved by above fixes |
| `hf_hub` import | 1 | `#[cfg(feature = "hf-hub")]` |
| safetensors functions | 5 | Gate behind safetensors-compare |

## Expected Outcome

`cargo build -p apr-cli --no-default-features --features training` compiles on aarch64 and x86_64, producing an `apr` binary with `apr train apply` support.

## Verification

```bash
# On gx10 (aarch64):
cargo build --release -p apr-cli --no-default-features --features training
./target/release/apr train apply --task pretrain --config configs/train/pretrain-350m-cuda-test.yaml
```

PR #2045, opened 2026-06-15
feat(cuda-oxide): opt-in sm_121 Q4K GPU backend (pure-Rust→PTX), parity-proven (PMAT-734)

noahgift:
## Opt-in cuda-oxide sm_121 Q4K backend — a pure-Rust→PTX FOUNDATION (not yet a live decode win)

Wires the verified pure-Rust cuda-oxide Q4K kernel into `aprender-serve` as an **opt-in (`APR_Q4K_OXIDE=1`), sm_121-gated, default-OFF** backend. The kernel is parity-bit-exact vs `TiledQ4KGemv` on GB10 (max|Δ|=5.96e-8) and ~2× faster than it per-kernel at K≥6656.

### ⚠️ HONEST SCOPE (corrected after the real-decode capstone)
This is a **kernel-authoring + integration foundation**, **NOT yet a live decode speedup**:
- Real `apr run <gguf> --gpu` decode with `APR_Q4K_OXIDE=1` vs default = **byte-identical, 1.00× (no change)** on 1.5B + 7B qwen. The oxide gate sits in `q4k_gemv_cached_async`, but GGUF decode uses the **workspace path** (`q4k_gemv_into → Q4kVariant::HwDp4a`), which doesn't call it — so the opt-in is currently inert (hence harmless: default OFF + unreachable).
- The proven ~2× was vs `TiledQ4KGemv`; production decode uses the optimized **`HwDp4a`** kernel, which oxide has **not** been benchmarked against. Whether oxide can beat the real production kernel is the open question (measuring next).
- For common qwen models the K≥6656 GEMVs are **Q6_K** (ffn_down), so a Q4K-only oxide kernel won't apply there regardless.

### Why merge anyway
Correct, opt-in, default-OFF, parity-tested, and the **source-of-record + integration pattern** for pure-Rust→PTX kernels (the north-star: escape hand-PTX Blackwell JIT pain — the kernel runs without the GH-480 workaround). It changes nothing in a normal build.

### To deliver a real decode win (follow-up)
1. Benchmark oxide vs **HwDp4a** (decisive — in progress). 2. If oxide wins, wire it into the workspace-path dispatch. 3. Add a Q6_K oxide kernel for ffn_down. Do not claim a P4 decode lift until step 1 proves a win.

### Contents
PTX asset (CB-510-safe) + dispatch gate (default OFF, sm_120+) + `compute_capability()` + 3-way parity test (cc≥120-gated, skips on sm_89) + contract. fmt/clippy clean; non-cuda build unaffected.


noahgift:
Closing — the decisive A/B (oxide vs the REAL production kernel **HwDp4a**, not TiledQ4KGemv) on GB10 shows oxide is **~4× SLOWER** at every FFN shape (oxide/HwDp4a = 3.4–4.2×). HwDp4a (Q8_1 + half-warp DP4A) is what production decode auto-selects on Blackwell; the ~2× this PR cited was vs TiledQ4KGemv, which production never uses. So this backend is a 4×-slower, inert (wrong dispatch path), opt-in kernel — shipping it into aprender-serve is muda. cuda-oxide remains a proven **capability** (pure-Rust→PTX, parity-correct on sm_121), captured in the merged source-of-record (#2042, experiments/cuda-oxide/) + the draft doc. Revisit only if a DP4A-class oxide kernel can beat HwDp4a. (K-shape-gate/throughput-gate branches are moot — they gated vs the wrong incumbent.)

PR #2068, opened 2026-06-15
fix(serve): make legacy-quant GPU→CPU fallback LOUD + harden the gate (PMAT-781)

noahgift:
## `apr run --gpu` silently degraded to CPU on legacy-quant GGUF — now it's loud

Characterizing apr decode on GB10 showed `apr run --gpu <q4_k_m.gguf>` crawling at ~3 tok/s on CPU with zero indication the GPU was skipped, while `apr bench --fast` reported ~184 tok/s "GPU". **The CPU fallback is correct; the bug was its silence** — and `bench`'s number is measuring a broken path.

## What's actually going on (empirically verified)
- `apr run --gpu` → `run_gguf_generate` has a quant gate + a cpu↔gpu **parity gate**. `apr bench --fast` → `generate_gpu_resident` directly with **no correctness gate** (throughput only).
- The gate blocks legacy GGML quant `{Q4_0, Q4_1, Q5_0, Q5_1}`. qwen2.5-coder q4_k_m stores most attn/FFN tensors as **Q5_0** → gated → CPU.
- **The legacy-quant GPU GEMV kernels are genuinely broken** (separate, pre-existing): forcing the GPU path (`SKIP_PARITY_GATE=1`) produces **gibberish** on GB10 *and* the RTX 4090 (same wrong token, rel_gap≈0.54); `apr qa` → "GPU output failed (CPU passed)"; reproduces on the old `apr 0.4.11` → not introduced. So `bench`'s 184 tok/s is a garbage path; the gate was right to block it.

## Fix
- `inference_result.rs`: `log_cpu_backend` now emits the fallback reason **unconditionally** on stderr (was `if !verbose return`): `Backend: CPU (model uses legacy Q4_0/Q4_1/Q5_0/Q5_1 weights — GPU kernels diverge from CPU; falling back to CPU for correct output)`. Renamed `is_legacy_gguf_quant` → `is_gpu_unsupported_quant`; hardened `model_has_gpu_unsupported_quant` to inspect ALL projection tensors (QKV fused+separate, attn_output, ffn gate/up/down, lm_head), not a subset.
- `gguf_gpu_generate.rs`: the APR-CUDA loader's silent `return None` on legacy quant now prints a clear stderr line.

## Verified (both GPUs)
- GB10: before — silently CPU ~3 tok/s; after — same coherent CPU output but now LOUDLY explains why.
- RTX 4090: Q4_K/Q6_K-pure models (`qwen2.5-coder-1.5b q4_k_m`) → `Backend: GPU (RTX 4090)`, parity accepts, coherent — never gated, stay GPU-resident. Any remaining fallback is now loud, never silent.
- 714 infer-module tests pass; fmt + clippy clean.

Follow-up (separate, larger): the legacy Q4_0/Q4_1/Q5_0/Q5_1 GPU GEMV kernels diverging from CPU on both sm_89 + sm_121 is a real GPU correctness bug to fix so those models can use the GPU.

### DIFF
--- a/contracts/beat-ollama-decode-throughput-speed-v1.yaml
+++ b/contracts/beat-ollama-decode-throughput-speed-v1.yaml
@@ -1,0 +1,114 @@
+contract: beat-ollama-decode-throughput-speed
+metadata:
+  kind: beat-benchmark
+  version: "1.0.0"
+  description: >
+    Pillar-4 SPEED comparison (PMAT-755 / audit gap #4): apr GPU decode tok/s vs
+    ollama warm-daemon decode (eval) tok/s for the SAME Q4_K_M GGUF, SAME
+    host/GPU (RTX 4090 sm_89), SAME prompt. Audit gap #4 asked to convert the
+    "apr ~1.23x faster than ollama" observation (PMAT-742, 2026-06-13) from
+    TRACKING into a falsifiable GATED beat.
+
+    VERDICT (2026-06-15, measured first — see step 4 of the task): STAYS TRACKING.
+    A real ~1.3x clean-decode advantage EXISTS, but apr's `apr run` decode
+    throughput is too VARIABLE on this box to form a non-flaky gate, so shipping a
+    hard 1.10x gate would produce false FAILs. Documented honestly rather than
+    asserted as a green beat.
+
+    SCOPE: STEADY-STATE GPU DECODE only — marginal token-generation rate with
+    model load + FP8-weight-cache build + CUDA-graph capture + prefill amortized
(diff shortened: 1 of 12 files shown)


noahgift:
Closing: branch mixes an abandoned reach-GPU attempt (large under-described batch.rs change) with the loud-fallback, plus base-skew contamination. The genuine findings — (1) silent CPU fallback on legacy-quant GGUF, (2) legacy Q4_0/Q4_1/Q5_0/Q5_1 GPU GEMV kernels produce garbage on both sm_89+sm_121 — will be addressed cleanly. The loud-fallback core (make log_cpu_backend unconditional) is a 2-line change to re-do separately.

PR #2171, opened 2026-06-21
feat(cuda-oxide): FFN gate+up SwiGLU Q4K port — measured NO-GO (DP4A-gated, documented) (PMAT-881)

noahgift:
Honest NO-GO: ported the FFN fusion Q4K kernel to pure-Rust #[kernel], bit-exact on GB10 but 1.58x SLOWER than hand-PTX (Q8_1 DP4A int8 hardware instruction has no oxide intrinsic path). Records the rule: port FMA/softmax kernels (attention=GO PMAT-882), keep DP4A-bound Q4K GEMV/FFN on hand-PTX. Experiment + PMAT-881-STATUS.md.

Generated with Claude Code

### DIFF
--- a/experiments/cuda-oxide/PMAT-881-STATUS.md
+++ b/experiments/cuda-oxide/PMAT-881-STATUS.md
@@ -1,0 +1,98 @@
+# PMAT-881 — cuda-oxide port of the FusedGateUpSwiglu Q4K kernel
+
+**Status: COMPLETE (kernel works, parity PASS, perf measured) — VERDICT: NO-GO for production decode.**
+
+A pure-Rust `#[kernel]` port of the hand-PTX `FusedGateUpSwigluHwDp4aQ4KGemvKernel`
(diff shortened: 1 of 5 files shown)


noahgift:
Parking this PR: it conflicts with the merged PMAT-882 on the shared cuda-oxide files, and it's NO-GO scaffolding for the oxide-attention go-live — which PMAT-884 proved is blocked on CUDA-graph-capture (production decode is graph-captured; a cuLaunchKernel oxide kernel can't be recorded into the graph) AND the e2e decode is host-overhead-bound at ~10 tok/s (attention <0.3% of the per-token budget). The findings + the rule (port FMA/softmax kernels, keep DP4A-bound on hand-PTX) are recorded in memory + the roadmap; the branch persists for the graph-capture follow-up. The real lever is the ~10 tok/s bottleneck — pivoting there (PMAT-885). 🤖

PR #2258, opened 2026-07-02
fix(finetune): rank-aware learning rate — default `apr finetune -m qlora` no longer diverges

noahgift:
## The bug

`LoraOptimizer::find_optimal_rank` binary-searches the **largest** LoRA rank that fits the VRAM budget (up to 256), but `apr finetune` used a fixed CLI default learning rate of **2e-4**. That pairing **diverges** out of the box.

Measured live on an **RTX 4090** — 1.5B QLoRA, lr 2e-4 / rank 256 / seq 2048 on `apr_code_sft_balanced`:

| step | 10 | 20 | 30 | 40 | 50 | … | 160 | epoch avg |
|---|---|---|---|---|---|---|---|---|
| loss | 4.31 | **1.44** | 7.08 | 15.06 | 16.02 | … | 13.83 | **11.18** |

It starts learning (4.31 → 1.44) then blows up past the untrained baseline. 2e-4 is only known-good for the small ranks (≤ ~32) LoRA papers use.

## Root cause

`OptimalConfig` recommended `rank` and `alpha` but **not** a learning rate — so the two knobs the optimizer controls (rank ↑, lr fixed) combined into a divergent config.

## The fix

`OptimalConfig` gains a rank-aware `learning_rate` via `recommended_learning_rate(method, rank)`:

| rank | ≤32 | 64 | 128 | 256 | Full (0) |
|---|---|---|---|---|---|
| lr | 2e-4 | 1e-4 | 5e-5 | 2.5e-5 | 1e-5 |

Anchored at the classic 2e-4 for small ranks (no regression), scaled inversely above, clamped to `[1e-5, 2e-4]`. The rank-256 value (~2.5e-5) matches the empirically-stable ~2e-5.

`apr finetune` makes `--learning-rate` **optional**: omitted → the rank-aware recommendation (recomputed if `--rank` is overridden); an explicit value still wins. Early sub-modes (merge/classify/multi-adapter) keep the classic 2e-4 — the divergence was measured on the instruct LoRA/QLoRA path, so the auto-lowering is scoped there.

## Falsifiers — `FALSIFY-QLORA-RANK-AWARE-LR-001..003`

Pure **unit tests** (CI-enforced, no GPU needed): rank-256 lr ≤ 5e-5 (convergent band; **RED** on the old fixed 2e-4), bounds + monotonicity, and `optimize()` populates it end-to-end. Contract `qlora-rank-aware-lr-v1.yaml` (`pv validate`/`pv lint` clean).

## Testing

- `aprender-train-lora`: 58 pass (incl. 3 new falsifiers)
- `apr-cli` finetune: 69 pass; `--learning-rate` → `Option<f64>` threaded through `run()` + test callers
- fmt clean, clippy clean (touched files)

## Related

Complements #2257 (QLoRA per-epoch val_loss now reflects trained adapters) — honest val_loss makes exactly this kind of divergence visible during training instead of hidden behind a constant.

### DIFF
--- a/contracts/qlora-rank-aware-lr-v1.yaml
+++ b/contracts/qlora-rank-aware-lr-v1.yaml
@@ -1,0 +1,138 @@
+metadata:
+  id: C-QLORA-RANK-AWARE-LR
+  version: 1.0.0
+  created: '2026-07-02'
+  author: PAIML Engineering
+  kind: pattern
+  status: ACTIVE_ALGORITHM_LEVEL
+  description: |
+    Pins the auto-selected learning rate for `apr finetune -m lora/qlora` to the
+    convergent regime at the high LoRA ranks the planner picks to fill VRAM.
+
+    BACKGROUND. `LoraOptimizer::find_optimal_rank` binary-searches the LARGEST
+    rank that fits the VRAM budget (up to 256), then `apr finetune` used a fixed
+    CLI default learning rate of 2e-4. That pairing DIVERGES: measured live on an
+    RTX 4090, a 1.5B QLoRA run at lr 2e-4 / rank 256 / seq 2048 on
+    apr_code_sft_balanced went 4.31 -> 1.44 (learning) then blew up to 11-16,
+    epoch avg 11.18 — worse than the untrained model. The classic 2e-4 is a
+    known-good default only for the small ranks (<= ~32) LoRA papers use; at the
+    VRAM-filling ranks this optimizer auto-selects it is far too hot.
+
+    ROOT CAUSE. `OptimalConfig` recommended `rank` and `alpha` but NOT a learning
+    rate, and the CLI default (2e-4) was decoupled from the auto-selected rank.
+    So the two knobs the optimizer controls (rank up, lr fixed) combined into a
+    divergent configuration out of the box.
+
+    FIX. `OptimalConfig` gains a rank-aware `learning_rate`, computed by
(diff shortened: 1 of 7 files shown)


noahgift:
## E2E validation on the RTX 4090 (fresh binary, identical workload)

Beyond the unit falsifiers — ran the full 1-epoch e2e (1.5B QLoRA, rank 256, seq 2048, `apr_code_sft_balanced`, 160 samples / 18,568 tokens) at both learning rates:

| step | lr **2e-4** (old default) | lr **2.5e-5** (this PR's recommendation) |
|---|---|---|
| 10 | 4.31 | 4.31 |
| 20 | 1.45 | 1.41 |
| 30 | **7.08 ← diverging** | 1.11 |
| 50 | 16.02 | 0.28 |
| 90 | 12.43 | 1.33 |
| 120 | 11.40 | 1.25 |
| 160 | 13.83 | 3.52 |
| **epoch avg** | **11.1770** (worse than untrained) | **1.5800** (converged) |

Same seed path, same samples, 0 NaN / 0 CUDA errors in both runs. The recommendation (`2e-4 × 32/256 = 2.5e-5`) holds the convergent band on the exact workload where the old default blows up — the divergence is entirely the lr, and the rank-aware value fixes it end-to-end, not just in the unit falsifiers.

Logs: `run-FRESH-seq2048.log` (2e-4) / `run-lr2.5e-5-seq2048.log` (2.5e-5), lambda-vector.

PR #2303, opened 2026-07-05
fix(msrv): reconcile rust-version 1.89→1.91 + add MSRV gate (PMAT-MSRV-GATE-001)

noahgift:
## Fable rank-14 — the declared MSRV was a lie

`Cargo.toml` claimed `rust-version = "1.89"` in 8 places, but **nothing verified it** — `rust-toolchain.toml` pins 1.93, so 1.89 was never exercised. It doesn't build:

```
$ cargo +1.89 check --workspace
error: rustc 1.89.0 is not supported by pmcp@2.9.0 (requires 1.91.0)
       wasmtime-internal-*@43.0.2 requires rustc 1.91.0
```

`cargo +1.91 check --workspace` **Finishes clean** → 1.91 is the true floor. This bumps all 8 declarations to 1.91.

## The gate (`scripts/check_msrv.sh`)
Reads the declared `rust-version` and compiles the whole workspace on exactly that toolchain, failing on drift.

| rust-version | `bash scripts/check_msrv.sh` |
|---|---|
| **1.91** (this PR) | ✓ `workspace builds on its declared MSRV 1.91` |
| 1.89 (mutation) | ✗ fails — `pmcp@2.9.0 requires 1.91.0` |

Both directions verified locally.

**Follow-up** (deferred until the in-flight ci.yml PRs land + the merge queue is healthy): wire `check_msrv.sh` into `ci.yml` as a per-PR job.


noahgift:
Status update — **the root MSRV half of this PR has already landed** via the v0.61.0 release cut (#2313).

`main` now has:
```
Cargo.toml:123  rust-version = "1.91"
Cargo.toml:479  rust-version = "1.91"
```

I folded it into the release commit because it was not merely unverified — it was **provably false**: `cargo +1.89 check -p aprender` fails at resolve time, since the committed lockfile pins `pmcp` 2.9.0 and `wasmtime` 43.0.2, both requiring 1.91.0. `cargo install aprender` could not succeed on the toolchain we advertised, and publishes are immutable, so fixing it after the cut would have cost a release.

That leaves this PR with the two parts that are still genuinely open:

1. **The per-crate `rust-version` overrides** this PR touches (`aprender-train` 1.87, `aprender-graph` and 8 others 1.75, `aprender-verify-ml` 1.70) — untouched by the release fix.
2. **Wiring `scripts/check_msrv.sh` into CI.** It is currently referenced by zero workflows and zero Makefile targets, so it enforces nothing. Without that wiring this is a one-time correction rather than a gate, and the drift returns.

Needs a rebase onto `main` before it can move — the root `Cargo.toml` hunks will now conflict, since they are already applied.

Note for whoever wires the gate: `check_runner_labels.sh`’s DISCRIM accepts `cuda|ada|rtx4090`, so an MSRV job can pin `lambda-4090` if a `rustup toolchain install` inside the sovereign-ci container proves awkward.

Issue #1864, opened 2026-05-22
Qwen2.5-7B Q4_K GPU inference produces gibberish — 'ampiezza' (wgpu) / '<|im_start|>' (cuBLAS) — regression vs #374 / #559


noahgift:
## Summary

`apr run` and `apr qa` on `qwen2.5-coder-7b-instruct-q4_k_m.gguf` produce gibberish output via every GPU path, while CPU produces the correct answer. This is the same defect class as the previously-closed [#374](https://github.com/paiml/aprender/issues/374) (Q4K CUDA kernel, hidden_dim=3584) and [#559](https://github.com/paiml/aprender/issues/559) (GPU parity cosine=-0.005 on Qwen2.5-7B GQA hidden=3584, heads=28, kv=4), surfaced during v0.35.0 release dogfooding.

## Reproducer 1: `apr run` (wgpu/Vulkan, RTX 4090)

```
$ apr run /home/noah/models/qwen2.5-coder-7b-instruct-q4_k_m.gguf "What is 2+2?" --max-tokens 16
Backend: wgpu (Vulkan)
[PMAT-333] Dequantizing 28 layers (hidden=3584, heads=28/4, intermediate=18944)
[PMAT-333] Dequantized 337 weights, 28282.5 MB F32
[wgpu] Skipping weight 'lm_head' (2180.0 MB > 2147.5 MB limit) — CPU fallback

Output:
ampiezza

ampiezza

Completed in 40.49s (cached)
exit=0
```

## Reproducer 2: `apr qa` (cuBLAS / FP8 path)

```
$ apr qa /home/noah/models/qwen2.5-coder-7b-instruct-q4_k_m.gguf
✓ PASS Capability Match
✓ PASS Tensor Contract  (339 tensors, all PMAT-235 gates)
✓ PASS Metadata Plausibility  (arch=qwen2, rope_theta=1000000, max_pos=131072)
[GH-129] Early kernel preload: 49 modules compiled
[PMAT-082] cuBLASLt FP8 JIT warmed (3584×16×3584)
[PMAT-053] FP8 weight cache: 197 matrices cached (6742.8 MB) in 212.7ms
✗ FAIL Golden Output  GPU output failed (CPU passed): golden_output_gpu: gibberish (fragment "<|im_start|>" repeats 3+ times)
```

## Reproducer 3: 1.5B Q4K APR works (safety net firing)

```
[...]

PR #2337, opened 2026-07-29
feat(qa): format_parity compared ONE token — now it compares 64 decode steps

noahgift:
`PMAT-QA-FMTPARITY-DECODE-001` — Fable-review **EV rank 4**.

## What it was

`apr qa`'s `format_parity` gate — the tool the doctrine says to reach for *first* — ran one prefill forward per format and compared a single final-position argmax:

```rust
model.forward(&prompt_tokens)            // GGUF,        one call
run_safetensors_forward(&st, &prompt)    // SafeTensors, one call
compare_argmax_results(argmax(a), argmax(b))
```

Zero decode steps. The divergence class that actually ships — cached-decode corruption (PMAT-810 *"CertainlyCertainly"*, GH-1864) — was **structurally invisible** to it. It could only ever have caught a bug that was already wrong at token 0.

## What it is now

A ≥64-step lockstep greedy decode. Two design points decide whether it proves anything:

**Teacher forcing.** Both sides are fed the *same* next token (the GGUF argmax) rather than each following its own. If both free-ran, the first disagreement would put them on different sequences and every later step would compare unrelated distributions — the comparison would turn to noise exactly when it matters most. Teacher forcing keeps all 64 steps on one shared sequence, so a divergence at step *k* is attributable to step *k*.

**The cache path, not a re-prefill.** Both sides use the single-token-plus-cache entry point production uses (`forward_single_with_cache` / `forward_with_cache`), so a KV-cache indexing or write defect is *inside* the system under test. Re-prefilling each step would have been simpler and would have tested the wrong thing. The GGUF cache uses `OwnedQuantizedKVCache::from_config`, not `::new(.., hidden_dim, ..)` — Qwen2.5 is GQA (12 query heads, 2 KV heads), so `hidden_dim` would mis-stride the cache.

A top-1 disagreement whose logit vectors are cosine ≥ 0.98 is exempted as a floating-point near-tie. That exemption is load-bearing, not defensive — the real run below exercises it 4 times out of 64.

## Measured

On the RTX 4090 box that runs `qwen-story-daily`, 1.5B Q4_K_M GGUF vs the matching SafeTensors:

```
format_parity  PASS  value=64.0  threshold=64.0  19.8s
"Cross-format parity VERIFIED over 64 greedy decode steps through the cache path
 (60 exact top-1 matches, 4 near-tie exemptions at cosine >= 0.98)"
```

The old gate took 11.1s for its single forward — the full 64-step comparison costs ~8.7s more.

## RED-turning mutation (the roadmap's acceptance criterion, executed verbatim)

Dropped the SafeTensors-side KV write at decode step 32, rebuilt, re-ran:

```
apr qa exit code = 5
format_parity  FAIL  value=63.0  threshold=64.0
"Cross-format decode DIVERGED at step 33/64: GGUF argmax=151644 != SafeTensors
 argmax=198 (cosine=0.8185 < 0.98). 63 of 64 steps agreed (4 near-tie)."
```

It names **step 33** — the step after the dropped write — which is the physically correct answer. **The gate this replaces would have passed that mutation unchanged**, because step 0 still agreed.

## Making it actually run

The gate had a daily vehicle all along (`qwen-story-daily` → `qwen-story.sh`) but has **never once executed**: beat 2 passes the `.apr`, `run_format_parity_gate` peeks the magic bytes and SKIPs any non-GGUF primary — and `GateResult::skipped` sets `passed: true`, so `apr qa` still printed `ALL GATES PASSED`.

Adds a GGUF leg to beat 2 asserting on the **JSON gate object** (`.skipped == false && .passed == true`), not on that banner, which by construction cannot tell *"ran and passed"* from *"skipped"*. The leg is deliberately independent of `apr qa`'s exit code, since unrelated gates (ollama_parity, perf regression) legitimately fail on some hosts.

A missing GGUF/reference now **FAILs** the leg rather than SKIPping it. A silent skip on absent models is precisely what kept this gate dark.

### Unresolved finding, surfaced rather than worked around

The reference is pinned with `--safetensors-path` because on this runner **auto-discovery resolves to the HF cache snapshot, whose `layers.0.ffn_down_weight` reads back 100% zeros** and trips `F-DATA-QUALITY-001` during conversion — so the gate would FAIL on a bad reference instead of comparing anything. The copy under `~/models` converts fine. Whether that is a bad download or a converter defect (BF16 handling is the obvious suspect) is **not established here** and wants its own ticket. Either way a gate should not depend on which of several copies a search heuristic finds first.

## Verification

| check | result |
|---|---|
| new unit tests | 7, all executed |
| mutation-verify | ignoring `first_divergence` turns `decode_parity_late_divergence_fails_the_gate` RED and nothing else |
| full crate | 6613 passed, 0 failed |
| fmt / clippy | clean |
| `bashrs lint qwen-story.sh` | 0 errors |
| `pv validate apr-cli-qa-v1.yaml` | 0 errors, 0 warnings |

The floor test is the anti-theater one: it asserts ≥64 steps **and** that `--max-tokens` cannot shrink the comparison back to the single-forward shape.

Contract: new equation `format_parity_compares_decode_not_one_prefill` (6 invariants), a new proof obligation, and `FALSIFY-QA-013/014`. Both falsifier commands were executed verbatim — the first draft used two positional filters, which `cargo test` rejects, so they now use `--`. A falsifier that cannot run is the same defect one layer up.

### DIFF
--- a/contracts/apr-cli-qa-v1.yaml
+++ b/contracts/apr-cli-qa-v1.yaml
@@ -54,6 +54,25 @@ equations:
     - "An EXPLICIT --safetensors-path that does not exist still FAILs (user requested a specific reference)"
     - "A reference that IS present but whose outputs diverge still FAILs — the SKIP never swallows a real bug"
 
+  format_parity_compares_decode_not_one_prefill:
+    formula: |
+      apr qa GGUF_MODEL with a usable SafeTensors reference:
+        steps_compared >= 64
+        AND every step is produced by the PRODUCTION cache entry point
+            (forward_single_with_cache / forward_with_cache)
+        AND both sides are teacher-forced with the SAME token each step
+        AND top-1 must agree at every step, except where
+            cosine(logits_gguf, logits_st) >= 0.98 (near-tie exemption)
+    domain: "apr qa format_parity gate, GGUF primary with a convertible SafeTensors reference"
+    codomain: "GateResult { value = steps agreed, threshold = steps compared }"
+    invariants:
+    - "At least 64 decode steps are compared - the gate this replaced ran ONE prefill forward and compared a single final-position argmax, so a divergence appearing at decode step 32 was structurally invisible to it"
+    - "--max-tokens cannot shrink the comparison below the 64-step floor; a smaller request is raised to the floor, a larger one is honored"
+    - "Both formats are driven through the SAME cache path production uses, so a KV-cache indexing or write defect is inside the system under test rather than bypassed by re-prefilling"
+    - "Teacher forcing keeps both sides on one shared sequence: without it the first disagreement would put the two formats on different sequences and every later step would compare unrelated distributions"
+    - "A top-1 disagreement whose logit vectors are cosine >= 0.98 is a floating-point near-tie, not a structural divergence, and is exempted - without this the gate flakes on tied logits while proving nothing about the cache"
+    - "A structural divergence FAILs the gate and names the step, so the failure is actionable"
+
   json_validity:
     formula: |
       forall cmd in JSON_COMMANDS:
@@ -199,6 +218,18 @@ falsification_tests:
   test: "cargo test -p apr-cli --lib resolve_safetensors_explicit_path_is_honored_not_skipped"
   if_fails: "the absent-reference SKIP swallowed a real format divergence"
 
+- id: FALSIFY-QA-013
+  rule: "format_parity compares at least 64 DECODE steps, not one prefill argmax (PMAT-QA-FMTPARITY-DECODE-001)"
+  prediction: "the decode floor is >= 64 and --max-tokens cannot lower it"
+  test: "cargo test -p apr-cli --lib -- format_parity_decode_floor_is_at_least_64_steps max_tokens_cannot_shrink_decode_below_the_floor"
+  if_fails: "the gate has reverted to the single-forward shape - a cached-decode divergence at step 32 becomes invisible again while the gate still reports itself as cross-format parity"
+
+- id: FALSIFY-QA-014
+  rule: "a divergence appearing only after many decode steps FAILs the gate (PMAT-QA-FMTPARITY-DECODE-001)"
+  prediction: "a report whose first divergence is at step 32 yields passed=false, skipped=false, and a message naming the step"
+  test: "cargo test -p apr-cli --lib -- decode_parity_late_divergence_fails_the_gate decode_parity_all_steps_agreed_passes decode_parity_near_ties_do_not_fail_the_gate"
+  if_fails: "late divergence is swallowed - exactly the class (PMAT-810 batched-prefill decode corruption, GH-1864) the gate exists to catch"
+
 proof_obligations:
 - type: invariant
   property: "all 58 commands exit 0 on --help"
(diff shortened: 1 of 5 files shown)


noahgift:
## Follow-up: the open finding is now closed, and the gate was right

The PR body says the 100%-zeros SafeTensors reference was *"not established — bad download or converter defect (BF16 handling is the obvious suspect)."* Established now, and it is none of the suspects I named.

**The local HuggingFace cache blob is corrupt.** HF names LFS blobs by the upstream file's SHA256, which makes the check trivial:

```
blob name : c1b9b30e907950516ba3c646bdf570d8084c25a6410a0cdca80cf04b11bc13a8
sha256sum : 45e779fcd44290d24d84a9e52a98c5494a7ba5351e784dd383e0b26d8b4c50c1
```

Supporting evidence, in the order it ruled things out:

| check | result |
|---|---|
| dtype of the failing copy | **F32**, 339 tensors, 7.1 GB — so *not* a BF16 bug. The copy that converts fine is the BF16 one (338 tensors, 3.1 GB) |
| bytes on disk | `layers.0.mlp.down_proj.weight` really is all zeros; siblings `up_proj`/`q_proj` are normal (absmean ≈ 0.015 / 0.026) |
| extent | 27 of 339 tensors read back zero — 13 `mlp.down_proj`, 13 `self_attn.v_proj`, 1 `lm_head`. By tensor *type*, not a contiguous byte range |
| truncation | ruled out — file size == last data offset |
| sparse file / unwritten holes | ruled out — 7,108,395,008 bytes allocated vs 7,108,390,387 apparent. The zeros were actually written |

So `F-DATA-QUALITY-001` did exactly its job: it refused a genuinely corrupt model. **Nothing to fix in the converter** — the correct read of this is that the density gate caught a real defect, which is a point in its favour.

That leaves auto-discovery as the sole issue, and this PR already addresses it by pinning `--safetensors-path`. Two things worth separating out rather than folding in here:

1. The corrupt blob is only in this box's cache. Deleting it lets HF refetch, but that is 7 GB on someone else's bandwidth and nothing in CI needs it now that the reference is pinned — leaving that call to the operator.
2. **Nothing verifies HF cache blobs against their own names.** The blob filename *is* the expected hash, so an integrity check is essentially free and would have turned this from a confusing "conversion failed" into "your cached blob is corrupt, re-download it." That looks like a real `apr` feature (`apr pull --verify` / a cache-integrity gate) and deserves its own ticket rather than a drive-by.

noahgift:
### Refinement, and a trap worth recording

Everything in the comment above holds. Two additions from sweeping the rest of the cache.

**1. The dead tensors are filled with `-0.0`, not zero bytes.** The first eight bytes of `layers.0.mlp.down_proj.weight` are `00000000 00000080` — `+0.0` followed by `-0.0`. Only two distinct 4-byte words appear in the first 64 KB.

That matters because it defeats the obvious integrity check. I wrote one and it fooled me:

```python
raw = f.read(65536)
if not any(raw):        # byte-level "all zeros"
    flag_as_dead(k)     # <-- MISSES this tensor entirely
```

`any(raw)` is `True` because the sign bit is set, so a byte-level scan reports the blob **clean** while the tensor is 99.98% numerically zero. I briefly talked myself into believing the original finding was wrong on the strength of that scan before catching it. Any cache-integrity feature must compare *values*, not bytes:

```python
a = np.frombuffer(raw, dtype=np.float32)
if np.count_nonzero(a) == 0:   # correct
```

`F-DATA-QUALITY-001` already gets this right — it measures density on decoded values, which is why it caught the model my own scan missed.

**2. Not isolated, but not widespread either.** Re-swept all 81 SafeTensors blobs >100 MB in the cache with the value-level test:

| repo | dead / total float tensors |
|---|---|
| `Qwen--Qwen2.5-Coder-1.5B-Instruct` | 27 / 339 |
| `facebook--galactica-125m` | 12 / 196 |
| `sshleifer--distilbart-cnn-6-6` | 1 / 260 |

3 of 81. Only the Qwen one is load-bearing here, since it is what auto-discovery selects for this gate — which this PR already sidesteps by pinning `--safetensors-path`. The other two aren't referenced by any workflow.

PR #2348, opened 2026-07-31
fix(beat): withdraw the unreproducible ollama decode 1.371x claim — apr is at PARITY, not 1.37x

noahgift:
The Pillar-4 apr-vs-ollama GPU decode beat failed in `cuda-nightly` on 2026-07-31. Its panic message offered two diagnoses — *"apr's clean-decode advantage regressed, OR the #2049 FP8 stall fix is missing"* — and **both were wrong**. This PR corrects the claim rather than the threshold's victim.

## Four independent measurements, one host (lambda RTX 4090 sm_89)

| date | apr median | ollama median | ratio | source |
|---|---|---|---|---|
| 2026-06-15 | 412.3 | 300.7 | **1.371×** | promotion claim (#2067) |
| 2026-07-29 | 332.7 | 299.9 | 1.109× | cuda-nightly, **PASSED** |
| 2026-07-31 | 342.4 | 328.6 | 1.042× | cuda-nightly, FAILED |
| 2026-07-31 | 318.2 | 313.5 | **1.015×** | idle box, this harness |

**Read the ollama column first**: 300.7 / 299.9 / 328.6 / 313.5 across six weeks. The incumbent reproduces, so this is not the measuring rig drifting. apr's own column moved 412 → ~303–342.

The 2026-07-29 **"PASS" at 1.109× was already this regression**, clearing a 1.10 gate by 0.8%. Nobody looked, because it was green.

## What changed

#2323 (2026-07-27) made `auto_q4k` return `Mwv` on every device. sm_89 previously defaulted to `HwDp4a`, whose INT8 Q8_1 activation quant fails the F2 first-token cosine floor (0.9186 < 0.95). The 412.3 figure predates that change.

**Do not read that as "#2323 cost us 23%" — that is not what was measured.** Re-running today with the surviving opt-in, `HW_DP4A_Q4K=1`, measures **20.3 tok/s**, not 412: HwDp4a is F2-rejected, wgpu then fails its own 0.99 gate, and the run ends on CPU SIMD. That is exactly the ~20 tok/s collapse #2323 was written to fix.

So the honest statement is narrow: the 1.371× number was taken under a kernel default that **no longer exists** and cannot be reproduced here. It is **withdrawn as unreproducible, not reattributed**. A correct 318 tok/s is worth far more than a fast kernel the F2 gate refuses to let serve a token.

## Why 0.90 and not 1.10

Keeping 1.10 asserts a win apr does not currently have; it fails about every other night for a reason the gate cannot fix — which is how gates get ignored. Lowering it to ~1.00 would quietly redefine "beat" as "tie".

Instead the assertion is **renamed to what it can still prove** — apr must not *collapse* — and pinned at 0.90: 12% under the worst observed median (1.015) so it does not flake, while still catching the class that actually costs users, since a CPU-fallback measures ratio ~0.065 and violates this floor by **14×**.

The panic message no longer names a cause it cannot know. It reports the measurement and lists the three checks that discriminate (F2 rejection, missing `--features cuda`, `HW_DP4A_Q4K` set).

## Not fixed here

Restoring a real ≥ 1.10× win. **Until apr measures that on this host again, Pillar-4 must not claim a GPU decode win over ollama on sm_89.** Tracked separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -408,6 +408,15 @@ jobs:
       # green. Text-only check, no build.
       - name: Every CI-surface `apr` invocation must be pinned
         run: bash scripts/check_apr_bin_pinned.sh
+      # Poka-yoke: `set` in a SOURCED file mutates the caller's shell. apr_bin.sh
+      # opened with `set -euo pipefail`; qwen-story.sh sources it and had chosen
+      # `set -uo pipefail` deliberately (it must run every beat and tally the
+      # failures). The leak turned errexit on underneath it and the nightly
+      # story died after six lines inside an ADVISORY pmat hunt. Both files read
+      # correctly on their own - only the combination is wrong, which is why
+      # this is mechanical rather than a review note. Text-only, no build.
+      - name: Sourced libraries must not mutate the caller's shell options
+        run: bash scripts/check_sourced_libs_option_neutral.sh --self-test
 
   # Top-level gate: satisfies org ruleset "Green Main" which requires check named "gate".
   # The reusable workflow produces "ci / gate" but rulesets need exact match on "gate".
--- a/.github/workflows/qwen-story-daily.yml
+++ b/.github/workflows/qwen-story-daily.yml
@@ -115,10 +115,20 @@ jobs:
 
       - name: Extract pmat bug-hunt manifest
         id: manifest
+        shell: bash
         run: |
           # Pull the pmat hunt lines out of the log into a structured artifact.
+          #
+          # The pattern used to be '^(    gap|    churn|    fault)' - FOUR
+          # leading spaces - while qwen-story.sh emits these rows with EIGHT
+          # (see the pmat_rows filters). So it matched nothing, on every run:
+          # pmat-manifest.txt was 0 lines even on nights when the story log
+          # plainly contained `        churn ...` rows, which in turn pinned
+          # `growth` at 0 forever and made the "manifest grew by >5" alert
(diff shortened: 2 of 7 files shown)


noahgift:
**Follow-up observation from verifying this PR — the harness has no timeout on the ollama subprocess.**

While re-running the gate locally, `ollama run <model> <prompt> --verbose` hung. `ollama_eval_tps_once` calls `Command::new("ollama").…output()` with no timeout, so the harness waited indefinitely; the run produced no verdict at all and was eventually reaped by an outer `timeout 1200`.

The apr side is protected — `apr_decode_tps_once` discards a trial that early-EOSes, and `APR_ATTEMPTS > APR_MEDIAN_N` gives it slack. The ollama side has neither: `OLLAMA_TRIALS` are collected in a bare loop, and `.expect("failed to parse ollama eval rate …")` only fires if the process *returns*.

In `cuda-nightly` that means a hung incumbent burns the job's entire timeout and reports no measurement, rather than failing fast with "ollama did not respond". Same shape as the defects in #2347 — the failure is real but the surface says nothing useful about it.

Not fixed here to keep this PR to the claim withdrawal. Worth a small follow-up: wrap the ollama call in a timeout and treat a non-responsive daemon as an explicit provisioning failure.

**Verification status for this PR, stated precisely:** the completed harness run measured `apr_median7=318.2 ollama_median=313.5 ratio_median=1.015` and FAILED against the old 1.10 threshold. `1.015 >= 0.90` satisfies the floor this PR sets. I did *not* obtain a second clean green run — that attempt is the one that hung.

Issue #2408, opened 2026-08-10
`apr rm` can never remove anything and `apr list` marks every model orphan — list and rm do not share a namespace


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**7 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P1 | `rm` | `apr rm` can never remove anything — every name `apr list` prints is rejected as "Model not found in cache" | open |
| 2 | P1 | `list` | `apr list` marks every model "(orphan)" and reports "0 tracked" — including a model `apr pull` wrote seconds earlier; the registry manifest is never written | open |
| 3 | P1 | `encrypt` | `apr encrypt --key-file` accepts a ZERO-BYTE key file and produces a file anyone can decrypt; `ALBOR_ENCRYPT_KEY=""` bypasses the empty-passphrase guard the same way | open |
| 4 | P1 | `pull --verify` | `apr pull --verify` looks in a different cache tree than `apr pull` writes, and records no checksums at all for non-sharded models — contradicting its own --help | open |
| 5 | P2 | `publish` | `apr publish --dry-run` lists 1 of 4 files under "Files to upload:" — the companion files it will actually upload are hidden behind -v | open |
| 6 | P2 | `registry aliases` | `apr registry aliases` help says the map comes "from configs/aliases.yaml", but that file is never read — the alias map is compiled in and cannot be extended | open |
| 7 | P2 | `publish` | Auto-generated model card ships a non-compiling usage snippet that also names the wrong file (`model.apr` for a safetensors publish) | open |

## Evidence

<details><summary><b>1. [P1] `apr rm` can never remove anything — every name `apr list` prints is rejected as "Model not found in cache"</b></summary>

**Target:** `rm`

**Repro**

```bash
APR=/tmp/claude-1000/-home-noah-src-aprender/3e7b3507-59c1-4c5a-873b-c4003bfeca64/scratchpad/dogfood/cratesio/bin/apr
S=/tmp/claude-1000/-home-noah-src-aprender/3e7b3507-59c1-4c5a-873b-c4003bfeca64/scratchpad/dogfood/agent-registry-pkg
[...]

Issue #2373, opened 2026-08-10
Epic: dogfood audit of apr 0.63.0 from crates.io — 201 defects, 24 P0


noahgift:
Installed the published crate into a clean root — `cargo install aprender --root <scratch>`, default features, exactly what a user gets — and drove that binary and nothing else. No `cargo run`, no locally built artifact, no bare `apr` from `PATH`. 104 CLI subcommands, 9 MCP tools, ~45 HTTP routes, 700+ invocations against real models.

**201 distinct defects: 24 P0, 93 P1, 84 P2.**

| surface | P0 | P1 | P2 | total | probed |
|---|---|---|---|---|---|
| cli | 18 | 74 | 64 | 156 | 104 subcommands |
| http | 3 | 13 | 12 | 28 | ~45 routes |
| mcp | 3 | 6 | 8 | 17 | 9 tools |

`main` was two commits past the `v0.63.0` tag when this ran, both release-script changes, so every finding is live on `main`.

Full report with evidence: https://claude.ai/code/artifact/2ab8934a-3db7-49fe-adbf-eb0905c2de8e

## Five mechanisms produce most of them

Three of the five are failure modes this repo already documents having been burned by — reproduced *inside the product* rather than in CI.

1. **Gates that cannot fail.** The largest cluster. `apr qa --assert-tps 100000` passed at 4.1 tok/s because the assertion was divided by ten for GGUF and discarded for APR/SafeTensors. `apr rosetta compare-inference` reports `INFERENCE MATCH (100%)` for any pair of models. `apr explain` printed its error to stdout and exited 0.
2. **`apr` PATH-resolves a bare `apr`, in shipped code.** Proved with a shim: `apr mcp` launched from the 0.63.0 artifact executed 0.60.0 for all 8 subprocess-backed tools, while `apr.version` still reported 0.63.0. `apr code` does the same. This is the anti-pattern `CLAUDE.md` opens with, and the CI-side half of it is already tracked in #2358.
3. **Controls that fail open.** `--offline` downloads on `pull` / `chat` / `showcase`; `serve --no-cors` and `--no-metrics` are read only to suppress a banner line.
4. **Output fabricated rather than measured.** `/v1/explain` returns SHAP values independent of the input; `train apply --task pretrain` trains on built-in demo data and exits 0 reporting success; `cbtop --iterations 0` prints a full green 7/7 report having run nothing.
[...]

Issue #601, opened 2026-04-05
Multiple subcommands exit 0 on failure (lint, parity, rm)


noahgift:
## Description

Several subcommands report failures in their output but return exit code 0, making them useless for CI/scripting.

### 1. `apr lint` — exits 0 when lint fails

```bash
$ apr lint model.gguf; echo $?
✗ Lint failed 5 issue(s): 0 error(s), 3 warning(s), 2 info(s)
0
```

Expected: non-zero exit code when lint reports failures.

### 2. `apr parity` — exits 0 on CUDA init failure

```bash
$ apr parity model.gguf; echo $?
error: Validation failed: CUDA init failed: ... CUDA not available
0
```

The output says `error: Validation failed` but exit code is 0.

### 3. `apr rm` — exits 0 when model not found

```bash
$ apr rm nonexistent_model; echo $?
⚠ Model not found in cache
0
```

Expected: non-zero exit code when the model doesn't exist.

## Impact

CI pipelines and scripts that check exit codes will silently pass on real failures.

## Version

`apr 0.4.12`

noahgift:
## Additional case: apr qa --min-executed

```bash
$ apr qa --min-executed 5 --skip-throughput --skip-ollama --skip-golden \
    --skip-gpu-speedup --skip-gpu-state --skip-ptx-parity \
    --skip-format-parity --skip-capability model.gguf; echo $?
✗ GATES FAILED
error: Validation failed: Insufficient gate execution: 3 < 5 minimum
0
```

Another subcommand that prints "error: Validation failed" but exits 0. Running tally of commands affected:
- `lint` (#601a)
- `parity` (#601b / #615)
- `eval` (#616)
- `train plan` (#632)
- `validate` on INVALID (#647)
- `qa --min-executed` (this comment)

The exit-code-lies pattern is systemic — likely a single bug in the error-to-exit-code mapping layer.

noahgift:
## More cases: truncated file validation

All three validation commands exit 0 on a truncated/corrupt model file:

```bash
# 10MB truncated from 1GB GGUF
$ apr validate truncated.gguf; echo $?
VALID: 0 tensors checked   # 0 tensors = corrupt, but "VALID"
0

$ apr inspect truncated.gguf; echo $?
error: Tensor data exceeds file size
0

$ apr check truncated.gguf; echo $?  
error: Validation failed: Data range exceeds file size
0
```

[...]

PR #2410, opened 2026-08-10
fix(rm): apr rm could not remove anything apr list printed

noahgift:
Every name `apr list` printed was rejected by `apr rm`, including a model `apr pull` had just downloaded. The `hf://` ref and the absolute path failed identically, so a user's only way to reclaim cache space was `rm -rf` on `~/.cache/pacha/models`.

Found by dogfooding `cargo install aprender` 0.63.0 from crates.io.

## Before (apr 0.63.0 from crates.io, throwaway `XDG_CACHE_HOME`)

```
$ apr pull hf://hf-internal-testing/tiny-random-gpt2/model.safetensors
✓ Downloaded successfully
  Path: .../pacha/models/f36eaf6d2b971606.safetensors

$ apr list
f36eaf6d2b971606   443.2 KB  SafeTensors  .../f36eaf6d2b971606.safetensors (orphan)

$ apr rm hf://hf-internal-testing/tiny-random-gpt2/model.safetensors
⚠ Model not found in cache
error: File not found: hf://hf-internal-testing/tiny-random-gpt2/model.safetensors
rc=3
```

Varied the reference form — `apr rm f36eaf6d2b971606`, the file name, and the absolute path all failed the same way, and a second fixture (`064a3693fa1ea02c.safetensors`) reproduced identically. The file was still on disk after every attempt.

## Root cause

`apr list` and `apr rm` did not share a namespace.

`remove()` (`crates/apr-cli/src/commands/pull_remove_resolve_model.rs:12`) delegated solely to `ModelFetcher::remove()`, which looks up a URI-derived key in the in-memory `CacheManager` loaded from `<cache>/manifest.json` (`crates/aprender-registry/src/fetcher.rs:398`).

Nothing on the pull path writes that manifest. `run_single_file_streaming` (`crates/apr-cli/src/commands/pull.rs:122`) streams the file straight into the cache directory as `blake3(uri)[..16].<ext>` and never touches the fetcher. `apr list` already had to reconcile against the directory to see those files at all — which is exactly why it labelled every one of them `(orphan)`.

## Why the directory, not the manifest

The brief offered two shapes: teach `pull` to record what it downloaded, or make `rm` resolve against what `list` enumerates. This takes the second.

The manifest can never be complete. Files also land in the cache directory from `apr convert` and from plain copies, neither of which goes through the fetcher. Teaching `pull` to write the manifest fixes one writer and leaves `rm` still unable to remove anything the other two produced — the same class of bug, just narrower. The directory is complete by construction.

So `apr list`'s enumeration is now `scan_cache_dir()`, and `apr rm` resolves against that same function. The two commands read from one source and cannot drift apart again. `list`'s behaviour is unchanged: `scan_cache_dir_for_orphans` is now a filter over `scan_cache_dir`, so contract `apr-list-disk-reconciliation-v1` F-LIST-DISK-001 still holds.

`rm` accepts every form a user can plausibly have in hand: the identifier `list` prints, the file name, a path to the file, and the `hf://` ref that was pulled. An ambiguous stem (one model cached in two formats) is refused and deletes neither file — `rm` never guesses. GH-601's non-zero exit on a genuine miss is unchanged and still covered by a test.

## After (release binary built from this branch)

```
$ apr rm hf://hf-internal-testing/tiny-random-gpt2/model.safetensors
✓ Model removed from cache
  Path: .../pacha/models/f36eaf6d2b971606.safetensors
rc=0

$ apr list
No cached models found.

$ apr rm definitely-not-a-model
⚠ Model not found in cache
rc=3
```

Full `pull → list → rm → list` round trip verified end to end against a temporary `XDG_CACHE_HOME`, never the real cache. The stem form and the absolute-path form were verified the same way.

## Mutation check

Six falsification tests (FT-RMNS-001..006) drive `remove_from_cache_dir` against a throwaway cache directory and assert on the file being present or gone, not on a return shape.

Reverting only the fix — making `resolve_cached_model_files` return an empty `Vec`, i.e. restoring the manifest-only namespace — while keeping the tests turns four of the six RED, and leaves green exactly the two that encode preserved behaviour:

```
test removes_by_filename_and_by_absolute_path ... FAILED
test ambiguous_stem_deletes_nothing ... FAILED
test removes_by_the_hf_reference_that_was_pulled ... FAILED
test removes_by_the_name_list_prints ... FAILED
test non_model_files_are_not_removable ... ok
test unknown_reference_removes_nothing ... ok

FT-RMNS-002: file name form must delete
FT-RMNS-005: ambiguous ref must be an error: None
FT-RMNS-003: rm of the pulled hf:// ref must delete
  /tmp/apr-rm-ns-hfref-540543-ThreadId(5)/d47927c5638f80ab.gguf
assertion `left == right` failed
  left: None
 right: Some("/tmp/apr-rm-ns-stem-540543-ThreadId(6)/064a3693fa1ea02c.safetensors")

test result: FAILED. 2 passed; 4 failed
```

With the fix restored, 6 passed.

## Gates

- `cargo clippy -p apr-cli --lib -- -D warnings` — clean
- `cargo test -p apr-cli --lib` — 6633 passed, 0 failed, 12 ignored

## Adjacent, not fixed here

`apr rm` deletes the model file but leaves its companion sidecars (`<stem>.config.json`, `<stem>.tokenizer.json`, `<stem>.tokenizer_config.json`) in the cache directory. They are small and `apr list` does not show them, so this is cosmetic rather than the reported defect, and cleaning them up is a separate change with its own blast radius.

---

Refs #2408 (partial — see the issue for the findings that remain)

Audit epic: #2373

### DIFF
--- a/crates/apr-cli/src/commands/pull_list.rs
+++ b/crates/apr-cli/src/commands/pull_list.rs
@@ -1,23 +1,21 @@
 
-/// Scan the pacha cache directory for model files that are on disk but may not be
-/// tracked in the manifest (e.g., downloaded before pacha GH-162 added manifest
-/// persistence, or via direct writes outside the fetcher).
+/// Enumerate every model file sitting in the pacha cache directory.
 ///
-/// Contract: apr-list-disk-reconciliation-v1 F-LIST-DISK-001 (paiml/aprender#602).
-fn scan_cache_dir_for_orphans(
-    cache_dir: &Path,
-    known_paths: &HashSet<std::path::PathBuf>,
-) -> Vec<DiskModelEntry> {
+/// This is the single source of truth for "what is in the cache": `apr list`
+/// prints it and `apr rm` resolves against it (RM-NS-001). The pacha
+/// manifest is only an annotation on top of it — it can never be complete,
+/// because files land in this directory from the streaming pull path, from
+/// `apr convert`, and from plain file copies, none of which go through the
+/// fetcher. The directory always is complete, so it is the namespace both
+/// commands share.
+fn scan_cache_dir(cache_dir: &Path) -> Vec<DiskModelEntry> {
     let Ok(read_dir) = std::fs::read_dir(cache_dir) else {
         return Vec::new();
     };
-    let mut orphans = Vec::new();
+    let mut found = Vec::new();
     for entry in read_dir.flatten() {
         let path = entry.path();
         if !path.is_file() {
-            continue;
-        }
-        if known_paths.contains(&path) {
             continue;
         }
         let Some(ext) = path.extension().and_then(|s| s.to_str()) else {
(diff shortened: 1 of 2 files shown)


noahgift:

Issue #2390, opened 2026-08-10
`--json` emits human-formatted text on 10 commands, so no consumer can parse it


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**11 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P1 | `pull / rm / publish` | `--json` emits human-formatted text (unparseable) on `apr pull`, `apr rm`, and `apr publish` | open |
| 2 | P1 | `import` | apr import --json emits the human-readable table on stdout and no JSON at all; --quiet is likewise ignored | open |
| 3 | P1 | `profile` | `apr profile --json` is ignored entirely, and `--format json` prefixes stdout with human progress lines so the output does not parse as JSON | open |
| 4 | P1 | `code` | `apr code --json` is advertised in `apr code --help` but emits plain text; only the separate `--output-format json` works | open |
| 5 | P1 | `train apply --json` | `apr train apply --json` emits human-formatted text on stdout, not JSON | open |
| 6 | P1 | `chat` | `apr chat --json` emits the human banner instead of JSON and exits 0 | open |
| 7 | P1 | `canary` | `apr canary create` and `apr canary check` advertise `--json` but always emit human-readable text — no JSON in any flag position | open |
| 8 | P1 | `showcase` | `apr showcase --json` emits zero JSON — the advertised "Output results as JSON" produces only the ASCII banner and human log | open |
| 9 | P1 | `check` | `apr check --json` exits 0 when stages FAIL; the same command without --json exits 5 | open |
| 10 | P1 | `profile` | `apr profile --json` is listed in its own --help but is silently ignored (emits human output), unlike every other command in this slice | open |
| 11 | P1 | `hex` | apr hex --json is ignored in 7 of 8 display modes and entirely on SafeTensors — emits human-formatted tables on stdout instead of JSON | open |

## Evidence

<details><summary><b>1. [P1] `--json` emits human-formatted text (unparseable) on `apr pull`, `apr rm`, and `apr publish`</b></summary>

**Target:** `pull / rm / publish`

**Repro**

```bash
[...]

PR #2411, opened 2026-08-10
fix(cli): --json printed human text on pull, rm, publish and import

noahgift:
Four commands advertise `--json  Output as JSON` in their own `--help` and
then wrote human-formatted text to stdout, so anything piping them to a
parser failed at the first byte. Reproduced against apr 0.63.0 installed
from crates.io, reading stdout only (stderr discarded — the library chatter
already goes to stderr, so this is exactly the stream a consumer parses):

    $ apr pull qwen2.5-coder --dry-run --json     # rc=0
    === APR Pull ===

    Model:    qwen2.5-coder
    Resolved: hf://Qwen/Qwen2.5-Coder-7B-Instruct-GGUF/...
    $ ... | python3 -c 'import json,sys; json.load(sys.stdin)'
    json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

`apr rm`, `apr publish --dry-run` and `apr import` failed identically.
`apr publish` was the worst of them: it dumped the entire generated model
card, YAML front-matter and all, onto stdout under `--json`. The flag was
being presented as the machine-readable contract — `apr pull`'s own
not-found error tells users to run `apr registry aliases --json`.

Root cause is that the flag never reached the commands. `--json` is a
global on `Cli`, and the dispatcher simply did not forward it:

  crates/apr-cli/src/dispatch.rs:851  Commands::Rm { model_ref } => pull::remove(model_ref)
  crates/apr-cli/src/dispatch.rs:850  Commands::List => pull::list(cli.json, cli.quiet)

The line above `Rm` forwards it, which is why `apr list --json` was correct
all along. `Commands::Pull`, `Commands::Import` and the publish arm in
dispatch_analysis.rs:1549 had the same omission, so no amount of work
inside those commands could have seen the flag.

The fix forwards `cli.json` and, in each command, separates resolving the
facts from rendering them. `DryRunReport`, `DryRunPlan`, `remove_stdout`,
`import_json_stdout` and `q4k_import_json_stdout` each return the exact
string that goes to stdout, so a unit test asserts the bytes a consumer
parses rather than an internal shape. Errors follow the convention the
already-correct commands use (`apr validate --json`, `apr stamp --json`):
the diagnostic goes to stderr and the exit code carries the outcome, so
stdout holds a whole document or nothing — never a half-written one.

After, from a release binary built at this commit (apr 0.63.0 (d37f2fefc)):

    $ apr pull qwen2.5-coder --dry-run --json
    {
      "model": "qwen2.5-coder",
      "resolved": "hf://Qwen/Qwen2.5-Coder-7B-Instruct-GGUF/qwen2.5-coder-7b-instruct-q4_k_m.gguf",
      "revision": "main",
      "revision_kind": "RefName",
      "offline": false,
      "mode": "dry-run"
    }

All four now parse. `apr rm missing --json` writes 0 bytes to stdout, keeps
its rc=3, and puts `error: File not found: ...` on stderr. The default
human output of all four is byte-identical to 0.63.0 (verified by diffing
the two binaries' stdout) and the exit codes are unchanged.

Mutation check: reverting the five renderers to their 0.63.0 behaviour
while keeping the tests turns 9 of the 12 new tests RED, and the failure
messages reprint the original defect:

    test result: FAILED. 4 passed; 9 failed; 0 ignored

    ---- commands::pull::tests::pull_dry_run_json_stdout_carries_no_human_decoration ----
    human decoration "=== APR Pull ===" leaked into `--json` stdout:
    === APR Pull ===

    Model:    qwen2.5-coder
    ...
    ---- commands::publish::tests::test_publish_dry_run_json_stdout_parses_as_json ----
    `apr publish --dry-run --json` must write parseable JSON to stdout, but a
    consumer got expected value at line 1 column 1. Actual stdout was:
    === DRY RUN: Would publish to paiml/test-model ===
    ...

The 3 that stay green under the mutation are the ones asserting the human
mode is unchanged, which is what they are there for.

`apr pull`'s streaming download path is deliberately not covered: it has 66
further print sites including progress bars, and half-suppressing its
banner would only make unparseable output look intentional, so the banner
now lives inside the dry-run renderer instead. `apr chat` (an interactive
REPL, which needs a streaming protocol rather than a document) and
`apr train apply` (already forwards `cli.json`; its stdout is only reachable
through a real training run) are left for separate work.

6634 apr-cli lib tests pass; fmt and clippy -D warnings clean.

---

Refs #2390 (partial — see the issue for the findings that remain)

Audit epic: #2373


noahgift:

Issue #2383, opened 2026-08-10
`apr rerank` panics on ordinary flag values; `--normalize false` / `--with-pooler false` are documented but unreachable


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**4 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P0 | `rerank` | `apr rerank --input-ids` panics (slice index out of range) on any token id >= vocab_size instead of validating | open |
| 2 | P0 | `rerank` | `apr rerank --hidden-dim` / `--num-heads` panic on an incompatible value instead of erroring | open |
| 3 | P1 | `embed --normalize / rerank --with-pooler` | `--normalize false` and `--with-pooler false` are documented in --help but rejected by the parser, and the 'off' state is unreachable | open |
| 4 | P2 | `rerank` | `apr rerank --passage` + `--passages` is documented as mutually exclusive but is silently accepted, and one of the two inputs is discarded | open |

## Evidence

<details><summary><b>1. [P0] `apr rerank --input-ids` panics (slice index out of range) on any token id >= vocab_size instead of validating</b></summary>

**Target:** `rerank`

**Repro**

```bash
APR=/tmp/claude-1000/-home-noah-src-aprender/3e7b3507-59c1-4c5a-873b-c4003bfeca64/scratchpad/dogfood/cratesio/bin/apr
# build the cross-encoder fixture (ms-marco-MiniLM-L-6-v2, already in the pacha cache)
mkdir -p cesrc && cp /home/noah/.cache/pacha/models/57e6e922118ea840.safetensors cesrc/model.safetensors && cp /home/noah/.cache/pacha/models/57e6e922118ea840.config.json cesrc/config.json && cp /home/noah/.cache/pacha/models/57e6e922118ea840.tokenizer.json cesrc/tokenizer.json
timeout 300 "$APR" import cesrc/model.safetensors -o ce.apr --arch bert > /dev/null 2>&1
timeout 180 "$APR" rerank ce.apr --input-ids 101,999999,102 --token-type-ids 0,0,0 > out.txt 2>&1; echo "rc=$?"; cat out.txt
```

**Observed**

```
rc=101

[...]

PR #2412, opened 2026-08-10
fix(apr-cli): --normalize false and --with-pooler false were documented but unparseable, so both flags were permanently ON

noahgift:
Found by dogfooding `cargo install aprender` 0.63.0.

`apr embed --help` tells the user "Pass `--normalize false` to keep raw magnitudes", and `apr rerank --help` tells them "Cross-encoders that skip the pooler should pass `--with-pooler false`". Neither spelling parses. Against the 0.63.0 binary installed from crates.io:

```
$ apr embed model.apr --vocab tok.json --text hi --normalize false
error: unexpected argument 'false' found
rc=2

$ apr embed model.apr --vocab tok.json --text hi --normalize=false
error: unexpected value 'false' for '--normalize' found; no more were expected
rc=2

$ apr rerank model.apr --vocab tok.json --query q --passages p --with-pooler false
error: unexpected argument 'false' found
rc=2
```

The field is a bare `bool`, so clap derives a `SetTrue` switch and `default_value_t = true` pins it on: true when omitted, true when passed bare, rejected when given a value. The "off" state had no reachable spelling at all — un-normalised embeddings could not be produced, and a cross-encoder without a pooler dense layer could not be loaded.

**Root cause:** `crates/apr-cli/src/extended_commands.rs:1238` (`normalize`) and `:1193` (`with_pooler`) — `#[arg(long, default_value_t = true)]` on a bare `bool`. Both handlers already consume the value correctly (`embed.rs:283` gates `l2_normalize`, `rerank.rs:254` passes it to `CrossEncoder::new`); only the parser was wrong.

**Fix:** `num_args(0..=1)` + `default_missing_value = "true"` + `ArgAction::Set`, so the syntax `--help` already documents parses. Omitted and bare both still mean true, so no existing invocation changes meaning. There is no negatable-flag convention already in this CLI to follow — I grepped for `no-` longs, `default_missing_value` and `ArgAction::Set` and found no prior instances — and this spelling is the one the help text promises.

## Measured effect

Real `cross-encoder/ms-marco-MiniLM` (90 MB safetensors imported to APR v2, 105 tensors), text `"hello world"`, mean pooling:

| binary | invocation | `normalize` | ‖v‖ |
|---|---|---|---|
| crates.io 0.63.0 | flag omitted | true | 1.000000 |
| crates.io 0.63.0 | bare `--normalize` | true | 1.000000 |
| this build | `--normalize false` | false | **9.499828** |
| this build | `--normalize=false` | false | **9.499828** |
| this build | flag omitted / bare | true | 1.000000 |

Not a single-input result: with normalisation off, `"the quick brown fox"` gives 9.492353 and `"a completely different sentence about databases"` gives 8.943374; both are 1.000000 with it on. On the rerank side the same passage scores logit 8.796466 with the pooler and 0.108038 without, so the flag reaches the model.

## Mutation check

Twelve falsifiers assert the parsed value for each documented spelling, in the existing `pretrain.rs` CLI-parse style (16 MiB worker thread — the flattened command enum overflows the default test stack). Reverting only the two `#[arg(...)]` attributes and keeping the tests turns 6 of the 12 RED, quoting the exact strings a user sees:

```
---- commands::embed::tests::embed_normalize_space_separated_false_turns_it_off stdout ----
assertion `left == right` failed: `--normalize false` is documented in --help and MUST reach the handler as false
  left: Err("error: unexpected argument 'false' found\n\nUsage: apr embed [OPTIONS] --vocab <FILE> <MODEL>\n\n...")
 right: Ok(false)

---- commands::rerank::tests::rerank_with_pooler_space_separated_false_turns_it_off stdout ----
assertion `left == right` failed: `--with-pooler false` is documented in --help and MUST reach the handler as false
  left: Err("error: unexpected argument 'false' found\n\nUsage: apr rerank [OPTIONS] <MODEL>\n\n...")
 right: Ok(false)

test result: FAILED. 6 passed; 6 failed
```

Restoring the attributes returns 12/12 green. Full `cargo test -p apr-cli --lib`: 6634 passed, 0 failed. `cargo fmt --all -- --check` and `cargo clippy -p apr-cli --lib -- -D warnings` both exit 0 (status captured without a pipe).

## Trade-off, asserted rather than latent

A bare `--normalize` placed immediately before the MODEL positional now swallows the path as its value. That is covered by a test asserting the error is legible and names the offending token; the documented order `apr embed MODEL --normalize` is unaffected. `require_equals` would have avoided it but would break the `--normalize false` spelling `--help` documents.

## Adjacent, not fixed here

Three more bare `bool` fields carry `default_value_t = true` and have the same unreachable-off shape:

- `crates/apr-cli/src/commands/mono.rs:23` — `apr mono publish --dry-run`, so that command can never actually publish
- `crates/aprender-train/src/config/cli/extended.rs:125` — `--model-card`
- `crates/aprender-train/examples/finetune_test_gen.rs:352` — `--publish`

---

Refs #2383 (partial — see the issue for the findings that remain)

Audit epic: #2373

### DIFF
--- a/crates/apr-cli/src/commands/embed.rs
+++ b/crates/apr-cli/src/commands/embed.rs
@@ -449,4 +449,141 @@ mod tests {
             _ => panic!("expected ValidationFailed"),
         }
     }
+
+    // ── `apr embed --normalize` CLI surface ──────────────────────────
+    //
+    // `--help` documents "Pass `--normalize false` to keep raw
+    // magnitudes". With a bare `bool` field clap derives a SetTrue
+    // switch, so `default_value_t = true` pins the value ON and the
+    // documented `false` is rejected by the parser (exit 2). These
+    // tests assert the PARSED VALUE for each documented spelling, not
+    // merely that parsing succeeds.
+
+    /// Parse an exact `apr embed …` argv and return the `normalize`
+    /// field. `Err` carries clap's rendered error (what the user sees).
+    ///
+    /// Parsed on a 16 MiB thread: the flattened `Commands` enum is large
+    /// enough to overflow the default test stack (same reason as the
+    /// `apr pretrain` CLI tests).
+    fn parse_embed_normalize_argv(argv: &[&str]) -> std::result::Result<bool, String> {
+        let argv: Vec<String> = argv.iter().map(|s| (*s).to_string()).collect();
+        std::thread::Builder::new()
+            .stack_size(16 * 1024 * 1024)
+            .spawn(move || {
+                use clap::Parser;
+                match crate::Cli::try_parse_from(&argv) {
+                    Ok(cli) => match *cli.command {
+                        crate::Commands::Extended(crate::ExtendedCommands::Embed {
+                            normalize,
+                            ..
+                        }) => Ok(normalize),
+                        other => panic!("expected ExtendedCommands::Embed, got {other:?}"),
+                    },
+                    Err(e) => Err(e.to_string()),
+                }
+            })
+            .expect("spawn parse thread")
+            .join()
+            .expect("parse thread must not panic")
+    }
+
+    /// `apr embed <MODEL> --vocab … --text … <extra>` — the documented
+    /// argument order, with `extra` appended.
+    fn parse_embed_normalize(extra: &[&str]) -> std::result::Result<bool, String> {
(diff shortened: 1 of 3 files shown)


noahgift:

Issue #2378, opened 2026-08-10
`apr run`: 7 defects incl. `--stream` emitting empty text and `--trace-output` always writing an empty event list


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**10 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P1 | `run` | `apr run --stream` emits one event per token but every event's `text` field is empty, so nothing streams | open |
| 2 | P1 | `run` | `apr run --trace-output FILE` always writes `"events": []`, and `--trace-level chrome` ignores `--trace-output` and writes trace-<epoch>.json into the CWD | open |
| 3 | P1 | `run` | `apr run --trace --trace-level layer` prints per-step timings that are hardcoded fractions of wall time, presented as measurements, and contradicts the real profile in the same output | open |
| 4 | P1 | `run` | `apr run` silently accepts a GGUF whose final-norm tensor is half the declared hidden size — exit 0, no warning, altered output | open |
| 5 | P1 | `run` | `--backend`, `--trace-level` and `-f/--format` accept any string silently; `--backend <typo>` prints "Backend override: <typo>" and quietly runs the default backend | open |
| 6 | P2 | `run` | Every `apr run --trace` emits a false `[CONTRACT WARN] gpu-decode-profiling-v1 TOKEN_ACCOUNTING` — LmHead.count is deterministically tokens_processed-1 | open |
| 7 | P2 | `run / chat` | Two misleading user-facing error strings: an empty lm_head on a dense Qwen2 is blamed on "a MoE per-expert tensor", and a chat error prints the literal template placeholder `{stem}` | open |
| 8 | P0 | `run (default install, wgpu backend)` | Default `apr run <gguf>` peaks at 13.9 GB RSS and 11.1 s because it dequantizes the whole model to F32 for a wgpu path its own parity gate then rejects | open |
| 9 | P1 | `run (wgpu Q4_K numerics)` | The wgpu Q4_K backend is numerically wrong on RTX 4090 (cosine 0.884 vs CPU) and is still tried first by default | open |
| 10 | P1 | `run (--features cuda build)` | `--features cuda` build dumps developer debug traces — raw pointers and raw tensor values — to stderr on a normal `apr run` | open |

[...]

PR #2427, opened 2026-08-10
fix(run): --stream sent nothing, --trace-output recorded nothing, and a typo'd --backend ran the default backend

noahgift:
Six of the ten `apr run` findings in #2378, each reproduced against the installed
0.63.0 binary first and verified against a binary built from this branch.

## What a user saw

`apr run model.gguf "hi" -n 8 --no-gpu --stream` on 0.63.0:

```
{"event":"token","index":0,"token_id":40,"text":""}
{"event":"token","index":1,"token_id":2776,"text":""}
...
{"...,"text":"I'm here to help! How can","tokens":[40,2776,...],"event":"final"}
```

Eight events, every one of them empty. The ids were right — the final event
decodes them into the whole reply — so a consumer rendering `text` as events
arrive saw nothing until the run had already finished. Same branch, same command:

```
{"event":"token","index":0,"token_id":40,"text":"I"}
{"event":"token","index":1,"token_id":2776,"text":"'m"}
{"event":"token","index":2,"token_id":1588,"text":" here"}
{"event":"token","index":3,"token_id":311,"text":" to"}
{"event":"token","index":4,"token_id":1492,"text":" help"}
{"event":"token","index":5,"token_id":0,"text":"!"}
{"event":"token","index":6,"token_id":2585,"text":" How"}
{"event":"token","index":7,"token_id":646,"text":" can"}
```

`--trace-output FILE` wrote `"events": []` at every documented `--trace-level`,
and `--trace-level chrome` wrote its chrome JSON to `trace-<epoch>.json` in the
CWD while leaving the requested path holding the empty stub. Before / after,
same five levels:

```
before                        after
none:    events_len= 0        none:    events = 2 ['model_load', 'generate']
basic:   events_len= 0        basic:   events = 2 ['model_load', 'generate']
layer:   events_len= 0        layer:   events = 2 ['model_load', 'generate']
payload: events_len= 0        payload: events = 2 ['model_load', 'generate']
chrome:  events_len= 0        chrome:  chrome traceEvents = 12
trace-1786290823.json         stray trace-*.json in CWD: (none)
```

`--backend banana` printed `Backend override: banana` and ran the default
backend at exit 0 — the precise outcome the `--backend cuda` guard exists to
prevent, because it makes any throughput measured through that run meaningless.
Now:

```
run --backend banana   -> rc=2  error: invalid value 'banana' for '--backend <BACKEND>'
                                  [possible values: cuda, cpu, wgpu]
run -f banana          -> rc=2  error: invalid value 'banana' for '--format <FORMAT>'
                                  [possible values: text, json, srt, vtt]
run --trace-level banana -> rc=2 error: invalid value 'banana' for '--trace-level <LEVEL>'
                                  [possible values: none, basic, layer, payload, chrome]
chat --backend banana  -> rc=2  (same)
run --backend cpu      -> rc=0  (unchanged)
run --backend cuda     -> rc=5  (the build-capability guard still fires, unchanged)
```

## Root causes

| Finding | Cause |
|---|---|
| 1 `--stream` empty `text` | `run_entry.rs` wrote a literal `"text": ""` per event; nothing ever decoded the ids one at a time. Now filled from the model's own tokenizer (single-token decode, the same call the SSE streaming handler makes), resolved once per streamed run and only under `--stream`. |
| 2 `"events": []` + chrome path | `inference_result.rs:337` ended a `format!` string with that literal. Rebuilt with `serde_json` and filled from measured timings (`model_load` = `load_ms`, `generate` = `inference_ms`). The same `format!` also interpolated the model path raw into a JSON string, so a path containing `"` produced non-JSON — fixed by construction. `print_chrome_trace` now takes the requested path. |
| 3 fabricated layer timings | `gguf_generate_result.rs:373-380` computed `per_token_ms * {0.85, 0.08, 0.02, 0.017}` and printed it under a column headed `Time`. The values stay (they are the only estimate available without the brick profiler) but the table now says `ESTIMATED`, marks each value `~`, prints the share it assumed, and labels `TOTAL` as wall clock incl. model load and `RATE` as end-to-end — the 19x contradiction with the `[BRICK-PROFILE]` block six lines above was two unlabelled numbers. |
| 5 unvalidated flag values | `--backend`, `--trace-level`, `-f` were free-form `String`s. clap `value_parser`s on `run` and `chat`. |
| 6 false `TOKEN_ACCOUNTING` warning | `profiler_contracts.rs:69` compared `LmHead.count != tokens_processed`. `generate_with_cache` samples the final token and breaks at `tokens.len() >= max_seq_len` *before* feeding it back, so LmHead never fires for it while `tokens_processed` counts it — the gap was deterministically 1 on every healthy run (10/11, 11/12, 12/13, 16/17, 37/38). Invariant is now "one LmHead per forward pass": `tokens_processed` or `tokens_processed - 1`; real miscounting still warns. |
| 7 two misleading strings | `matmul_fused.rs:501` blamed "a MoE per-expert tensor" for an empty `lm_head.weight` on a dense tied-embedding Qwen2.5 with no experts at all; it now names both known causes and says to run `apr tensors` to find the 0-byte tensor. `chat.rs:301` printed the literal `{stem}.tokenizer.json` — an unsubstituted format placeholder inside a plain string — instead of the filename it looked for. |

## Two existing tests were holding defects in place

- `run_tests_stream_output.rs` asserted `v["text"].is_string()`. That is true of
  `""`, so the test passed for the entire life of the empty-text bug. It now
  asserts the content, and that concatenating the streamed pieces reproduces the
  final text.
- `run_tests_chrome_trace.rs` asserted against a hand-maintained **copy** of
  `print_chrome_trace`'s body kept in the test file — it could only prove the
  copy agreed with itself. The builder is now extracted (`build_chrome_trace`)
  and the tests call it.

## Mutation check

Each fix reverted with the tests kept, RED verbatim:

```
trace_output_events_are_populated_from_measured_timings
  trace-output must carry events, got an empty list: {"events":[], ...}

token_accounting_silent_when_last_token_is_never_fed_back
  left: Some("... TOKEN_ACCOUNTING: LmHead.count=10 outside [10, 11] for tokens_processed=11 ...")
  right: None

validate_empty_data_does_not_blame_moe_alone
  a dense tied-embedding model is the other known cause and must be named; got:
  ... likely a MoE per-expert tensor was registered with len-0 data — see aprender#1789

stream_token_events_carry_their_own_decoded_text
  token event 0 must carry decoded text, got: {"event":"token","index":0,"token_id":40,"text":""}

chrome_trace_honours_requested_output_path
  chrome trace must be written to /tmp/apr-chrome-trace-.../requested.json: No such file or directory

layer_trace_marks_derived_timings_as_estimates
  the table must say the per-step values are estimated; got: ... Step  Time  Share ...

test_find_qwen_tokenizer_error_message_content_when_no_cache
no_qwen_tokenizer_message_handles_bare_filename
  1. Pacha cache ({stem}.tokenizer.json alongside model)
```

Restored: all 8 GREEN. `cargo test -p apr-cli --lib` 6635 passed,
`cargo test -p aprender-serve --lib` 15486 passed, `cargo fmt --all -- --check`
clean, `cargo clippy -p apr-cli --lib` and `-p aprender-serve --lib` clean with
`-D warnings`.

## Not fixed here

- **4** `apr run` accepts a GGUF whose `output_norm.weight` is half the declared
  `embedding_length`. Needs shape validation at model load against the tensor
  contract, not a CLI change.
- **8, 9** the default wgpu path dequantizes the whole model to F32 (13.9 GB RSS
  for a 1.0 GB Q4_K) before a parity gate that then rejects it at cosine 0.884 on
  this GPU. Backend-ordering and wgpu-kernel work.
- **10** the `--features cuda` build's unconditional debug spew (238 `eprintln!`
  under `crates/aprender-serve/src/cuda/`) — needs a CUDA build to verify.

Also left alone deliberately: `apr serve --backend` / `--trace-level` still take
free-form strings. Same fix applies, different cluster's file.

Refs #2378 (partial) — findings 1, 2, 3, 5, 6, 7 fixed; 4, 8, 9, 10 remain.
Audit epic: #2373

### DIFF
--- a/crates/apr-cli/src/commands/chat.rs
+++ b/crates/apr-cli/src/commands/chat.rs
@@ -296,16 +296,40 @@ fn find_qwen_tokenizer(model_path: &Path) -> Result<Option<Qwen2BpeTokenizer>, C
         }
     }
 
-    Err(CliError::InvalidFormat(
-        "No Qwen tokenizer found. Searched:\n\
-         1. Pacha cache ({stem}.tokenizer.json alongside model)\n\
-         2. Model directory (tokenizer.json)\n\
-         3. HuggingFace cache (~/.cache/huggingface/hub/models--Qwen--*/snapshots/*/tokenizer.json)\n\
-         4. APR cache (~/.apr/tokenizers/qwen2/tokenizer.json)\n\n\
+    Err(CliError::InvalidFormat(no_qwen_tokenizer_message(
+        model_path,
+    )))
+}
+
+/// The "no tokenizer" message for `model_path`, with every search location
+/// spelled out as a concrete path.
+///
+/// This message used to print the literal text `{stem}.tokenizer.json` — the
+/// format placeholder was written inside a plain string, so it was never
+/// substituted and the user was shown a template instead of the filename `apr`
+/// actually looked for. Naming the real paths is the difference between "go
+/// find out what apr means by stem" and "create this file".
+fn no_qwen_tokenizer_message(model_path: &Path) -> String {
+    let dir = model_path
+        .parent()
+        .filter(|p| !p.as_os_str().is_empty())
+        .map_or_else(|| ".".to_string(), |p| p.display().to_string());
+    let stem = model_path
+        .file_stem()
+        .and_then(|s| s.to_str())
+        .unwrap_or("<model>");
+    let home = dirs::home_dir().map_or_else(|| "~".to_string(), |h| h.display().to_string());
+
+    format!(
+        "No Qwen tokenizer found for {model}. Searched:\n\
+         1. Pacha cache:       {dir}/{stem}.tokenizer.json\n\
+         2. Model directory:   {dir}/tokenizer.json\n\
+         3. HuggingFace cache: {home}/.cache/huggingface/hub/models--Qwen--*/snapshots/*/tokenizer.json\n\
+         4. APR cache:         {home}/.apr/tokenizers/qwen2/tokenizer.json\n\n\
          To fix: Download a Qwen model with tokenizer:\n\
-           apr pull hf://Qwen/Qwen2.5-0.5B-Instruct-GGUF"
-            .to_string(),
-    ))
+           apr pull hf://Qwen/Qwen2.5-0.5B-Instruct-GGUF",
+        model = model_path.display(),
+    )
 }
 
 /// Normalize repeated punctuation (max 3 repeats of `!`, `?`, `.`).
--- a/crates/apr-cli/src/commands/chat_command_find_qwen.rs
+++ b/crates/apr-cli/src/commands/chat_command_find_qwen.rs
@@ -59,27 +59,45 @@
         assert!(result.is_ok() || result.is_err());
     }
 
+    /// The message must name the file the user has to produce.
+    ///
+    /// This test used to build its OWN copy of the error string and assert on
+    /// that, so it passed while the shipped message printed the literal
+    /// `{stem}.tokenizer.json` — an unsubstituted format placeholder — to every
+    /// user who ran `apr chat` on a model with no tokenizer. It now calls the
+    /// real message builder.
     #[test]
     fn test_find_qwen_tokenizer_error_message_content_when_no_cache() {
-        // When find_qwen_tokenizer fails, it should return an InvalidFormat error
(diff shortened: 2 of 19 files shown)


noahgift:

PR #2385, opened 2026-08-10
fix(serve): --no-cors and --no-metrics were inert — an operator who hardened a deployment had not

noahgift:
Dogfooding `cargo install aprender` 0.63.0: `apr serve run --no-metrics --no-cors`
produced byte-for-byte the same HTTP surface as `apr serve run` with no flags at
all. GET /metrics still returned 200 with the full Prometheus counter set, and
every response still carried `access-control-allow-origin: *`. The flags parsed,
the help text advertised them, and nothing downstream ever read them — so an
operator who believed they had hardened a deployment had not.

Repro against the published binary:

```
APR=.../cratesio/bin/apr   # apr 0.63.0 (v0.63.0+no-git)
"$APR" serve run /home/noah/.cache/pacha/models/d47927c5638f80ab.gguf \
    --port 18501 --no-metrics --no-cors &
curl -s -i http://127.0.0.1:18501/metrics | head -8
curl -s -D - -o /dev/null http://127.0.0.1:18501/health | grep -i access-control
```

Before:

```
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
access-control-allow-origin: *
access-control-expose-headers: *
# HELP realizar_requests_total Total number of requests
realizar_requests_total 0

access-control-allow-origin: *
access-control-expose-headers: *
```

After, same model, same port, same flags, binary built from this branch:

```
HTTP/1.1 404 Not Found
{"error":"not_found","message":"Route not found. See /health for available endpoints."}

(no access-control headers)
```

## Root cause

Neither flag reached the router. `--no-metrics` set `ServerConfig.metrics`, which
was consulted at exactly one place — `crates/apr-cli/src/commands/serve/mod.rs:89`
— to decide whether to print the `GET /metrics` banner line; the route itself was
registered unconditionally at `crates/aprender-serve/src/api/router.rs:63`.
`--no-cors` set `ServerConfig.cors`, a field carrying `#[allow(dead_code)]` and
the comment "accepted but not yet implemented"
(`crates/apr-cli/src/commands/serve/types.rs:29`), while `router.rs:136` applied
`CorsLayer::permissive()` to every router unconditionally. The startup banner was
the only thing either flag could move.

## Change

`RouterConfig` now carries `cors` and `metrics` alongside `openai_api`, both
defaulting to `true`, so unflagged behaviour is unchanged.
`create_router_with_config` registers `/metrics`, `/metrics/dispatch` and
`/metrics/dispatch/reset` only when metrics are enabled, and applies the
[...]

PR #2434, opened 2026-08-10
fix(mcp): apr mcp dropped every tool result on stdin EOF, and one bad byte killed the server

### LINKED ISSUE #2393: MCP protocol layer: 7 defects incl. responses dropped on stdin EOF and one bad UTF-8 byte killing the server

**7 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

|---|---|---|---|---|
| 1 | P1 | `apr mcp / initialize` | initialize hard-errors (-32602) on any protocolVersion other than the exact string 2024-11-05, instead of replying with the version the server supports | open |
| 2 | P1 | `apr mcp (stdio read loop)` | A single invalid UTF-8 byte on stdin terminates the whole MCP server (exit 1); all subsequent requests on the session are dropped | open |
| 3 | P2 | `apr mcp / ping` | The base-protocol `ping` method returns -32601 Method not found; a client using ping keepalive will treat the server as dead | open |
| 4 | P2 | `apr mcp / JSON-RPC error codes` | Well-formed JSON that is missing `jsonrpc` or `method` is reported as -32700 Parse error instead of -32600 Invalid Request — inconsistent with the server's own handling of a WRONG jsonrpc value | open |
| 5 | P2 | `apr mcp / JSON-RPC batch array` | A JSON-RPC batch array is rejected with a serde-internal message ("invalid type: map, expected a string") that tells the client nothing about batching | open |
| 6 | P2 | `apr mcp / tools/call argument validation` | A required argument supplied with the wrong JSON type is reported as "Missing required argument", contradicting the request the client actually sent | open |
| 7 | P0 | `apr mcp (stdio transport, all tools/call)` | tools/call responses are silently DROPPED when stdin reaches EOF — server exits 0 having answered initialize and tools/list but not the tool call | open |

## Evidence

<details><summary><b>1. [P1] initialize hard-errors (-32602) on any protocolVersion other than the exact string 2024-11-05, instead of replying with </b></summary>

**Target:** `apr mcp / initialize`

**Repro**

```bash
for v in 2025-06-18 2025-03-26 2024-11-05 2024-10-07; do
  printf '%s\n' "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"initialize\",\"params\":{\"protocolVersion\":\"$v\",\"capabilities\":{},\"clientInfo\":{\"name\":\"claude-code\",\"version\":\"1\"}}}" \
  | timeout 30 "$APR" mcp | cut -c1-160 | sed "s/^/[$v] /"
done
```

**Observed**

```
[2025-06-18] {"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unsupported protocolVersion: client requested \"2025-06-18\", server speaks \"2024-11-05\""}}
[2025-03-26] {"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unsupported protocolVersion: client requested \"2025-03-26\", server speaks \"2024-11-05\""}}
[2024-11-05] {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2024-11-05","capabilities":{"tools":{"listChanged":false}},"serverInfo":{"name":"aprender-mcp","version":"0.63.0"}}}
[2024-10-07] {"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"Unsupported protocolVersion: client requested \"2024-10-07\", server speaks \"2024-11-05\""}}
```

**Expected / root cause**

The MCP lifecycle contract is: if the server supports the requested version it responds with that version; OTHERWISE it responds with a version it does support and lets the client decide whether to proceed or disconnect. Returning a JSON-RPC error aborts the handshake, so a client that negotiates any version newer than 2024-11-05 can never connect. `apr mcp --help` and crates/aprender-mcp/README.md:25 both advertise this server for "Claude Code / Cursor / Cline" via .mcp.json, and those clients negotiate 2025-03-26/2025-06-18. Fix site: crates/aprender-mcp/src/server.rs:114-134 (handle_initialize) — replace the -32602 early return with a success response carrying crate::PROTOCOL_VERSION.

**Variations tried:** Five protocolVersion values (2025-06-18, 2025-03-26, 2024-11-05, 2024-10-07, "latest") — only the exact literal "2024-11-05" is accepted; every other value including an OLDER dated version is rejected, so this is not a floor check. Also confirmed the field is genuinely optional (omitting protocolVersion, and omitting `params` entirely, both succeed), and that a second initialize is accepted — so the rejection is specifically string inequality, not handshake state.

</details>

<details><summary><b>2. [P1] A single invalid UTF-8 byte on stdin terminates the whole MCP server (exit 1); all subsequent requests on the session ar</b></summary>

**Target:** `apr mcp (stdio read loop)`

**Repro**

```bash
cd /tmp && printf '{"jsonrpc":"2.0","id":1,"method":"tools/list"}\n\xff\n{"jsonrpc":"2.0","id":2,"method":"tools/list"}\n' > r3.bin
timeout 30 "$APR" mcp < r3.bin > r3.out 2> r3.err; echo "rc=$?"; echo "responses=$(wc -l < r3.out)"; cat r3.err
```

**Observed**

```
rc=1
responses=1
error: Aprender error: mcp server: stream did not contain valid UTF-8

(only the id=1 response was written; the id=2 request after the bad byte is never answered)
```

**Expected / root cause**

A malformed input line must produce a -32700 Parse error and the server must keep serving — which it already does correctly for malformed/truncated JSON (`{not json` -> -32700, next line still answered). Non-UTF-8 bytes should be handled the same way (read bytes and lossily decode, or emit -32700 and continue) rather than propagating the io::Error out of the read loop and killing the process. Fix site: crates/aprender-mcp/src/server.rs:291 — `let line = line?;` inside run_stdio's `for line in stdin.lock().lines()`.

**Variations tried:** Three distinct malformed-byte shapes, all 3 kill the server with the identical message and rc=1: (a) a lone 0xFF line in the middle of a valid stream, (b) a leading 0x80 byte before any valid request, (c) an invalid 2-byte sequence 0xC3 0x28 embedded inside a JSON string value of an otherwise well-formed tools/call. Contrasted against the well-formed-UTF-8 malformed-JSON cases (`{not json`, truncated `{"jsonrpc":"2.0","id":1,"method":"tools/li`), which correctly return -32700 with rc=0 and leave the server serving subsequent lines — proving the difference is the UTF-8 decode path, not JSON parsing.

</details>

<details><summary><b>3. [P2] The base-protocol `ping` method returns -32601 Method not found; a client using ping keepalive will treat the server as </b></summary>

**Target:** `apr mcp / ping`

**Repro**

```bash
printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"ping"}' | timeout 30 "$APR" mcp
```

**Observed**

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32601,"message":"Method not found: ping"}}
```

**Expected / root cause**

`ping` is part of the MCP base protocol (not a capability that has to be advertised): the receiver should respond promptly with an empty result `{"jsonrpc":"2.0","id":1,"result":{}}`. Clients that poll with ping for liveness interpret an error/absent pong as a dead connection and may terminate or restart the server. Fix site: crates/aprender-mcp/src/server.rs:102-111 (the method match in handle_request) — add a `"ping" => JsonRpcResponse::success(id, json!({}))` arm.

**Variations tried:** Compared against the methods that are genuinely NOT advertised: resources/list and prompts/list also return -32601, but that IS correct because the initialize result advertises only `capabilities.tools` — so those are working, not broken, and are not reported. Also ran ping both before and after a successful initialize + notifications/initialized handshake: same -32601 either way, so it is not a handshake-state artifact.

</details>

<details><summary><b>4. [P2] Well-formed JSON that is missing `jsonrpc` or `method` is reported as -32700 Parse error instead of -32600 Invalid Reque</b></summary>

**Target:** `apr mcp / JSON-RPC error codes`

**Repro**

```bash
printf '%s\n' '{"id":1,"method":"tools/list"}'      | timeout 30 "$APR" mcp
printf '%s\n' '{"jsonrpc":"2.0","id":1}'            | timeout 30 "$APR" mcp
printf '%s\n' '{"jsonrpc":"1.0","id":1,"method":"tools/list"}' | timeout 30 "$APR" mcp
```

**Observed**

```
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error: missing field `jsonrpc` at line 1 column 30"}}
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error: missing field `method` at line 1 column 24"}}
{"jsonrpc":"2.0","id":1,"error":{"code":-32600,"message":"Invalid Request: jsonrpc must be \"2.0\", got \"1.0\""}}
```

**Expected / root cause**

Per JSON-RPC 2.0, -32700 means the payload was not valid JSON; a syntactically valid JSON object that is not a valid Request object is -32600 Invalid Request, and the response should echo the request's id (1) rather than null. The server already gets this right when `jsonrpc` has the wrong VALUE (-32600, id echoed) but falls back to serde's error for a MISSING field, so the two cases disagree. This also loses the id, so a client cannot correlate the failure. Fix site: crates/aprender-mcp/src/server.rs:296-303 — deserialize to a permissive value first, then validate shape, instead of letting serde_json's missing-field error map to -32700.

**Variations tried:** Varied which required field is absent (jsonrpc, method) — both -32700; varied wrong value vs missing (jsonrpc:"1.0" -> -32600 with id echoed) to establish the inconsistency; confirmed genuinely-malformed JSON (`{not json`, truncated line) correctly yields -32700, and that 400-deep nested params yields a clean -32700 "recursion limit exceeded" with no panic, so the parse path itself is sound.

</details>

<details><summary><b>5. [P2] A JSON-RPC batch array is rejected with a serde-internal message ("invalid type: map, expected a string") that tells the</b></summary>

**Target:** `apr mcp / JSON-RPC batch array`

**Repro**

```bash
printf '%s\n' '[{"jsonrpc":"2.0","id":1,"method":"tools/list"},{"jsonrpc":"2.0","id":2,"method":"tools/list"}]' | timeout 30 "$APR" mcp
```

**Observed**

```
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error: invalid type: map, expected a string at line 1 column 1"}}
```

**Expected / root cause**

Batching is optional for a server that speaks 2024-11-05, so refusing it is defensible — but the error must say so. "invalid type: map, expected a string at line 1 column 1" is serde's attempt to parse the first array element as the `jsonrpc` string field; it names neither batching nor arrays, and column 1 points at the '[' which is valid JSON. A client author debugging this gets no signal. Expected either per-element batch handling, or -32600 with a message like "Invalid Request: JSON-RPC batch arrays are not supported; send one request per line". Fix site: crates/aprender-mcp/src/server.rs:296-303.

**Variations tried:** Tried a 2-element batch and a 1-element batch — identical message both times. Confirmed the same two requests sent as separate newline-delimited lines are both answered correctly (ids 1 and 2), so the content is valid and only the array framing is at fault. Also confirmed the server stays alive after the batch line and answers the next line normally.

</details>

<details><summary><b>6. [P2] A required argument supplied with the wrong JSON type is reported as "Missing required argument", contradicting the requ</b></summary>

**Target:** `apr mcp / tools/call argument validation`

**Repro**

```bash
# stdin must be held open, else the response is dropped (see mcp-toolscall-response-dropped-on-eof)
{ printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"apr.validate","arguments":{"model_path":123}}}'; sleep 3; } | timeout 30 "$APR" mcp
```

**Observed**

```
{"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"Missing required argument: model_path"}],"isError":true}}
```

**Expected / root cause**

`model_path` is present — it is a number where the declared inputSchema says {"type":"string"}. The message should say the type is wrong (e.g. "Argument model_path must be a string, got number"), because an LLM or client told "missing" will retry by adding a key it already sent. The isError:true envelope shape itself is correct MCP.

**Variations tried:** Compared the three neighbouring cases to isolate the wrong message: arguments omitted entirely, arguments:{} (empty object), and arguments:{"model_path":123} all return the byte-identical "Missing required argument: model_path", so the wrong-type case is genuinely indistinguishable from truly-absent. Also checked arguments as a non-object ("notanobject") on a no-arg tool (apr.version) — that is silently accepted and succeeds, and name:"" correctly yields "Unknown tool: ".

</details>

<details><summary><b>7. [P0] tools/call responses are silently DROPPED when stdin reaches EOF — server exits 0 having answered initialize and tools/l</b></summary>

**Target:** `apr mcp (stdio transport, all tools/call)`

**Repro**

```bash
D=/tmp/claude-1000/-home-noah-src-aprender/3e7b3507-59c1-4c5a-873b-c4003bfeca64/scratchpad/dogfood/agent-mcp-tools
mkdir -p $D/pin && ln -sf "$APR" $D/pin/apr
export PATH="$D/pin:$PATH"
for i in 1 2 3; do
  printf '%s\n' \
   '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"t","version":"1"}}}' \
   '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"apr.version","arguments":{}}}' \
  | timeout 60 "$APR" mcp > $D/eof$i.out 2>$D/eof$i.err
  echo "run$i rc=${PIPESTATUS[1]} lines=$(wc -l <$D/eof$i.out)"; cat $D/eof$i.out
done

# also with a 3s tool and a trailing tools/list:
printf '%s\n' '<initialize>' \
 '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"apr.validate","arguments":{"model_path":"/home/noah/models/Qwen3.5-0.8B-Q4_K_M.gguf"}}}' \
 '{"jsonrpc":"2.0","id":3,"method":"tools/list","params":{}}' | timeout 120 "$APR" mcp > $D/eof4.out
```

**Observed**

```
run1 rc=0 lines=1
{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2024-11-05","capabilities":{"tools":{"listChanged":false}},"serverInfo":{"name":"aprender-mcp","version":"0.63.0"}}}
run2 rc=0 lines=1   (identical)
run3 rc=0 lines=1   (identical)
-> the id=2 response is never emitted, 3/3.

eof4.out: rc=0 lines=2
  id=1 initialize   present
  id=3 tools/list   present
  id=2 tools/call   ABSENT
stderr empty in all runs.

Mechanism: server.rs:342 routes "tools/call" to spawn_tools_call_worker (a worker thread that writes its own response, server.rs:406), while initialize/tools/list are answered inline on the read loop. The read loop returns on EOF (server.rs:298) without joining in-flight workers, so the worker's write never happens.
```

**Expected / root cause**

A JSON-RPC server must answer every request that carries an id, or emit a JSON-RPC error. On EOF it must drain/join in-flight tools/call workers before exiting (or exit non-zero to signal the loss). As shipped, the canonical `printf ... | apr mcp` invocation — the form documented in the tool's own help text for scripted clients — silently loses every tool result while exiting 0, which is indistinguishable from a tool producing no output. Keeping stdin open until the response arrives is the only working client pattern and is undocumented.

**Variations tried:** 4/4 deterministic across two shapes: 3 identical runs with the instant apr.version tool, plus one run with the ~3s apr.validate tool followed by a tools/list — proving it is not a fast-tool race and that the transport keeps serving inline methods while the worker's reply is lost. Independently corroborates the trap flagged by the protocol slice.

**Exit code:** `0 (server), response missing`

</details>

noahgift:
Driving the shipped 0.63.0 `apr mcp` over stdio, the canonical scripted invocation loses its answer:

```
printf '<initialize>\n<tools/call apr.version>\n' | apr mcp
-> exit 0, ONE response line (id=1), the id=2 tool result never written
```

3/3 deterministic, stderr empty. `tools/call` runs on a worker thread that writes its own response, while `initialize`/`tools/list` answer inline on the read loop. The loop returned the instant stdin closed without joining those workers, so the process exited while a worker still owed the client a reply. Exit 0 with no output is indistinguishable, to a client, from a tool that produced nothing — the failure is silent. Holding stdin open was the only working client pattern and was documented nowhere.

Same loop, second defect: one invalid UTF-8 byte terminated the whole server.

```
rc=1, "error: Aprender error: mcp server: stream did not contain valid UTF-8"
```

Every request after the bad byte was lost. `BufRead::lines()` surfaces a bad byte as an `io::Error` and `let line = line?` propagated it out. The server already handled well-formed-UTF-8 malformed JSON correctly (-32700, keep serving), so it disagreed with itself about what a malformed message costs.

Both now: lines are read as BYTES and decoded per line, so a bad line is a malformed *message* answered with -32700; and EOF joins every in-flight worker before returning.

## The other five

| Finding | Before | After |
|---|---|---|
| `initialize` negotiation | -32602 on anything but the literal `2024-11-05` (including *older* versions — string inequality, not a floor check) | success carrying `2024-11-05`, per the MCP lifecycle |
| `ping` | -32601 Method not found | `{"result":{}}` |
| Missing `jsonrpc`/`method` | -32700 with `id:null` (while a *wrong* `jsonrpc` value was already correctly -32600 with id echoed) | -32600 with id echoed; malformed JSON stays -32700 |
| Batch array | serde's `invalid type: map, expected a string at line 1 column 1` | -32600 naming batching |
| Wrong-type argument | `Missing required argument: model_path` for an argument that WAS sent | `Argument model_path must be a string, got number` |

The negotiation defect barred every client this server advertises itself to: Claude Code and Cursor propose 2025-03-26 / 2025-06-18, and `README.md:25` names them.

## Root causes

All in `crates/aprender-mcp/src/server.rs` at 0.63.0: the read loop (`for line in stdin.lock().lines()` / `let line = line?`, no worker join on EOF), `handle_initialize`'s -32602 early return, the absent `ping` arm, and `serde_json::from_str::<JsonRpcRequest>` mapping a missing field to a parse error.

## Why the loop was made generic

`run_stdio` is now a two-line binding over a new `serve_stream<R, W>`. Not cosmetic: FALSIFY-MCP-010 and -011 live in the read loop, not in request handling, so no `handle_request` test can see either — and invalid UTF-8 cannot even be expressed as a `&str` input. CI's `workspace-test` runs `--lib` plus an explicit allowlist that does not include this crate's `tests/` directory, so an integration-only guard for a P0 would never run. Making the loop generic puts six falsifiers for it in the `--lib` target CI actually executes.

## Mutation check

Reverted each fix, kept the tests, rebuilt the binary so the harness could not reuse a fixed one:

```
serve_stream_answers_tools_call_before_returning_on_eof
  tools/call response (id=2) was DROPPED at EOF; got 1 response(s): [...]
serve_stream_answers_every_pipelined_tools_call
  id=2 unanswered; got 1 of 6
serve_stream_survives_invalid_utf8_line
  serve_stream must not propagate an error out of the session:
  stream did not contain valid UTF-8: invalid utf-8 sequence of 1 bytes
falsify_mcp_007_protocol_version_mismatch_negotiates_down
  proposing 1999-01-01 must not abort the handshake, got:
  Some(JsonRpcError { code: -32602, ... })
wrong_type_names_the_type_and_never_says_missing
  left:  "Missing required argument: model_path"
  right: "Argument model_path must be a string, got number"
```

At the binary level the reverted build reproduced the issue verbatim, including `tools/call response (id=2) was DROPPED on stdin EOF` and rc=1 `stream did not contain valid UTF-8`.

## Two tests did not survive that check

- `falsify_m1`'s FALSIFY-MCP-007 **asserted the defect** — it required -32602 on any mismatch, so the gate was actively holding the handshake broken. Rewritten together with its contract entry.
- A `tools/call` "ordering variant" in the new integration file **PASSED against the unfixed binary**: serializing the large `tools/list` response gives the worker time to finish before EOF, so it could never observe the drop it claimed to guard. Removed — a test that stays green on the defect is theater.

`validate.rs` / `finetune.rs` wrong-type tests asserted only `is_error == Some(true)`, which is shape; they passed while the message contradicted the request. Strengthened to assert the text.

## Contract

`contracts/apr-mcp-server-v1.yaml` gains FALSIFY-MCP-010 through -014; the loader test's hardcoded nine-gate count becomes a named constant. `pv validate` clean.

## End-to-end verification

Against a binary built from this tree, using the issue's own repros: all four proposed protocolVersions negotiate to 2024-11-05, ping pongs, both missing-field cases are -32600 with id echoed, the batch array is named, the wrong-type argument names the type, the tools/call result arrives 3/3 on EOF, and the bad-byte stream answers ids 1 and 2 plus a -32700 with rc=0.

Note: the crates.io 0.63.0 reference binary was never present at the path the task specified, so the "before" side is the issue's recorded evidence plus a locally rebuilt reverted binary, which reproduced every finding.

Fixes #2393
Audit epic: #2373

### DIFF
--- a/contracts/apr-mcp-server-v1.yaml
+++ b/contracts/apr-mcp-server-v1.yaml
@@ -150,11 +150,16 @@ falsification_conditions:
 
   - id: FALSIFY-MCP-007
     description: >
-      An `initialize` whose `params.protocolVersion` does not match the
-      server's supported version is rejected with JSON-RPC error code
-      -32602 (Invalid Params); the negotiation must not advance.
+      An `initialize` whose `params.protocolVersion` differs from the
+      server's is answered with a SUCCESS result carrying the version the
+      server does support, never a JSON-RPC error. The MCP lifecycle makes
+      version negotiation a proposal the client may accept or abandon by
+      disconnecting; answering -32602 aborts the handshake, which barred
+      every client proposing a newer dated version (Claude Code / Cursor
+      propose 2025-03-26 / 2025-06-18) from connecting at all. Revised
+      2026-08-10 from the inverted gate found by the 0.63.0 dogfood (#2393).
     test_file: crates/aprender-mcp/tests/falsify_m1.rs
-    test_name: falsify_mcp_007_protocol_version_mismatch_is_minus_32602
+    test_name: falsify_mcp_007_protocol_version_mismatch_negotiates_down
     status: ENFORCED
 
   - id: FALSIFY-MCP-008
@@ -183,14 +188,81 @@ falsification_conditions:
     test_name: falsify_mcp_009_no_reply_to_notification
     status: ENFORCED
 
+  - id: FALSIFY-MCP-010
+    description: >
+      Every request carrying an `id` is answered BEFORE the read loop
+      returns, including a `tools/call` still in flight when stdin reaches
+      EOF. `tools/call` runs on a worker thread that writes its own
+      response; the 0.63.0 loop returned the instant stdin closed without
+      joining those workers, so the canonical `printf ... | apr mcp`
+      invocation silently dropped every tool result while exiting 0 —
+      indistinguishable, to the client, from a tool that produced no output.
+      Added 2026-08-10 from the 0.63.0 dogfood (#2393, P0).
+    test_file: crates/aprender-mcp/src/server.rs
+    test_name: serve_stream_answers_tools_call_before_returning_on_eof
+    status: ENFORCED
+
+  - id: FALSIFY-MCP-011
+    description: >
+      A line that is not valid UTF-8 is a malformed MESSAGE: it is answered
+      with -32700 and the session keeps serving, exactly as an already-
+      correct malformed-JSON line is. The 0.63.0 loop read lines via
+      `BufRead::lines()`, which surfaced a bad byte as an io::Error that
+      propagated out and killed the process (exit 1), losing every request
+      after it — while well-formed-UTF-8 malformed JSON was handled
+      correctly, so the server disagreed with itself.
+      Added 2026-08-10 from the 0.63.0 dogfood (#2393).
+    test_file: crates/aprender-mcp/src/server.rs
+    test_name: serve_stream_survives_invalid_utf8_line
+    status: ENFORCED
+
+  - id: FALSIFY-MCP-012
+    description: >
+      Valid JSON that is not a valid Request object is -32600 Invalid
+      Request with the client's `id` echoed, NOT -32700 Parse error with
+      `id: null`; genuinely malformed JSON stays -32700; and a JSON-RPC
+      batch ARRAY is declined with a message that names batching rather
+      than leaking serde's "invalid type: map, expected a string". The
+      server already classified a WRONG `jsonrpc` VALUE correctly, so the
+      missing-field path disagreed with its own neighbour and lost the id
+      the client needs to correlate the failure.
+      Added 2026-08-10 from the 0.63.0 dogfood (#2393).
+    test_file: crates/aprender-mcp/src/server.rs
+    test_name: missing_required_field_is_invalid_request_with_id_echoed
+    status: ENFORCED
+
+  - id: FALSIFY-MCP-013
+    description: >
+      `ping` is MCP base protocol, not an advertised capability: it is
+      answered with an empty result at any point in the session. 0.63.0
+      returned -32601 Method not found, which a keepalive client reads as a
+      dead server and acts on by restarting it.
+      Added 2026-08-10 from the 0.63.0 dogfood (#2393).
+    test_file: crates/aprender-mcp/src/server.rs
+    test_name: ping_returns_empty_result
+    status: ENFORCED
+
+  - id: FALSIFY-MCP-014
+    description: >
+      A required `tools/call` argument supplied with the WRONG JSON TYPE is
+      reported as a type error naming the received type, never as
+      "Missing required argument". 0.63.0 emitted the byte-identical
+      "missing" message for absent, empty and wrong-type arguments, so a
+      client (or an LLM) told the argument was missing retries by adding a
+      key it already sent, and loops.
+      Added 2026-08-10 from the 0.63.0 dogfood (#2393).
+    test_file: crates/aprender-mcp/src/tools/args.rs
+    test_name: wrong_type_names_the_type_and_never_says_missing
+    status: ENFORCED
+
 # ── Acceptance gate ──
-# Promoting this contract out of ACTIVE requires all 9 conditions above to
+# Promoting this contract out of ACTIVE requires all 14 conditions above to
 # hold simultaneously on a clean checkout with `cargo test -p aprender-mcp`.
 
 qa_gate:
   id: F-MCP-SERVER-GATE
   name: "apr mcp end-to-end contract"
-  pass_criteria: "All 9 FALSIFY-MCP-* tests listed in falsification_conditions pass"
+  pass_criteria: "All 14 FALSIFY-MCP-* tests listed in falsification_conditions pass"
   run: "cargo test -p aprender-mcp --no-fail-fast"
   falsification: >
     Rename or delete any referenced test_file / test_name and the
--- a/crates/aprender-contracts/tests/apr_mcp_server_contract.rs
+++ b/crates/aprender-contracts/tests/apr_mcp_server_contract.rs
@@ -5,8 +5,10 @@
 //!
 //! 1. The YAML file exists and parses as valid YAML.
 //! 2. Top-level `status: ACTIVE`.
-//! 3. Exactly 9 entries in `falsification_conditions`, with ids
-//!    FALSIFY-MCP-001 through FALSIFY-MCP-009 (no gaps, no duplicates).
+//! 3. Exactly 14 entries in `falsification_conditions`, with ids
+//!    FALSIFY-MCP-001 through FALSIFY-MCP-014 (no gaps, no duplicates).
+//!    010-014 were added 2026-08-10 for the transport and protocol defects
+//!    found by the 0.63.0 crates.io dogfood (#2393).
 //! 4. Every entry has a non-empty `test_file` that exists on disk relative
 //!    to the workspace root, plus a non-empty `test_name` and
 //!    `status: ENFORCED`.
@@ -75,29 +77,39 @@ fn apr_mcp_server_contract_is_active() {
     );
 }
 
+/// Number of FALSIFY-MCP gates the spec defines. Raised from 9 to 14 on
+/// 2026-08-10 when the 0.63.0 crates.io dogfood (#2393) surfaced five gaps
+/// this contract had no gate for: responses dropped on stdin EOF (-010), a
+/// bad UTF-8 byte killing the session (-011), Invalid-Request vs Parse-error
+/// classification (-012), `ping` (-013), and wrong-type argument diagnostics
+/// (-014).
+const EXPECTED_GATES: usize = 14;
+
 #[test]
-fn apr_mcp_server_contract_has_exactly_nine_conditions() {
+fn apr_mcp_server_contract_has_exactly_expected_conditions() {
     let contract = load_contract();
     assert_eq!(
         contract.falsification_conditions.len(),
-        9,
-        "spec defines 9 FALSIFY-MCP gates; contract has {}",
+        EXPECTED_GATES,
+        "spec defines {EXPECTED_GATES} FALSIFY-MCP gates; contract has {}",
         contract.falsification_conditions.len()
     );
 }
 
 #[test]
-fn apr_mcp_server_contract_ids_are_falsify_mcp_001_through_009() {
+fn apr_mcp_server_contract_ids_are_contiguous_from_001() {
     let contract = load_contract();
     let actual: Vec<String> = contract
         .falsification_conditions
         .iter()
         .map(|c| c.id.clone())
         .collect();
-    let expected: Vec<String> = (1..=9).map(|n| format!("FALSIFY-MCP-{n:03}")).collect();
+    let expected: Vec<String> = (1..=EXPECTED_GATES)
+        .map(|n| format!("FALSIFY-MCP-{n:03}"))
+        .collect();
     assert_eq!(
         actual, expected,
-        "ids must be exactly FALSIFY-MCP-001..009 in order (no gaps, no duplicates)"
+        "ids must be exactly FALSIFY-MCP-001..{EXPECTED_GATES:03} in order (no gaps, no duplicates)"
     );
 }
 
--- a/crates/aprender-mcp/src/server.rs
+++ b/crates/aprender-mcp/src/server.rs
@@ -61,6 +61,13 @@ type InFlight = Arc<Mutex<HashMap<serde_json::Value, CancelHandle>>>;
 #[derive(Debug, Default)]
 pub struct AprMcpServer {
     in_flight: InFlight,
+    /// Join handles for `tools/call` workers spawned by
+    /// [`Self::spawn_tools_call_worker`]. The read loop MUST join these
+    /// before returning on EOF, otherwise the process exits while a worker
+    /// still owes the client a response and the answer is lost — see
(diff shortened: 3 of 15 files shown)


noahgift:

noahgift:

Issue #2396, opened 2026-08-10
Ollama-compat surface: `stream:true` discarded, model-discovery endpoints absent


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**3 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

| status | |
|---|---|
| addressed by | #2371 |
| remaining | 2 of 3 findings still open |

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P1 | `POST /api/chat, POST /api/generate` | Ollama created_at is a raw unix epoch ("1786289584.000000000Z"), not RFC3339 — the real Ollama Go client fails to unmarshal the whole response | fixed by #2371 |
| 2 | P2 | `GET /api/tags, GET /api/show, POST /api/embeddings` | Ollama model-discovery endpoints (/api/tags, /api/show, /api/embeddings) are absent although the server banner advertises "Ollama-Parity Endpoints" | open |
| 3 | P1 | `POST /api/chat, POST /api/generate (stream:true)` | Ollama /api/chat and /api/generate accept and then silently discard "stream":true — always one buffered JSON object, never NDJSON | open |

## Evidence

<details><summary><b>1. [P1] Ollama created_at is a raw unix epoch ("1786289584.000000000Z"), not RFC3339 — the real Ollama Go client fails to unmars</b></summary>

**Target:** `POST /api/chat, POST /api/generate`

**Repro**

```bash
APR=/tmp/claude-1000/-home-noah-src-aprender/3e7b3507-59c1-4c5a-873b-c4003bfeca64/scratchpad/dogfood/cratesio/bin/apr
"$APR" serve run /home/noah/.cache/pacha/models/d47927c5638f80ab.gguf --port 18320 &
sleep 25
curl -s -m 120 -X POST http://127.0.0.1:18320/api/chat -H 'Content-Type: application/json' \
  -d '{"model":"default","messages":[{"role":"user","content":"What is 2+2?"}],"stream":false,"options":{"num_predict":8}}'
# prove it breaks the real client type (ollama api.ChatResponse uses time.Time)
mkdir -p /tmp/gochk && cd /tmp/gochk && go mod init t >/dev/null 2>&1
cat > main.go <<'EOF'
package main
import ("encoding/json";"fmt";"time")
[...]

PR #2435, opened 2026-08-10
fix(serve): /realize/model invented its own provenance and the GGUF serve path was not Ollama-compatible at all

### LINKED ISSUE #2402: `/realize/model` returns fabricated provenance; `/realize/reload` points at a flag `apr serve run` does not have

**2 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

|---|---|---|---|---|
| 1 | P1 | `GET /realize/model` | /realize/model returns fabricated provenance and model metadata: content_hash is the literal string "blake3:0" repeated 16 times, size_bytes is hardcoded 0, and context_length/quantization/format are hardcoded constants that contradict the loaded model | open |
| 2 | P1 | `POST /realize/reload` | /realize/reload's 501 tells the user to "Start server with --registry flag" — `apr serve run` has no --registry flag, so the endpoint is unreachable by any documented invocation | open |

## Evidence

<details><summary><b>1. [P1] /realize/model returns fabricated provenance and model metadata: content_hash is the literal string "blake3:0" repeated </b></summary>

**Target:** `GET /realize/model`

**Repro**

```bash
M=/home/noah/.cache/pacha/models/d47927c5638f80ab.gguf
"$APR" inspect "$M" | grep -iE 'File Size|context_length|Quantization'
"$APR" serve run "$M" --port 18310 --host 127.0.0.1 --context-length 128 > /tmp/serve2.log 2>&1 &
until curl -sf --max-time 2 http://127.0.0.1:18310/health >/dev/null; do sleep 2; done
curl -sS -w '\nHTTP_STATUS:%{http_code}\n' --max-time 10 http://127.0.0.1:18310/realize/model
```

**Observed**

```
# ground truth from `apr inspect` on the SAME file:
│    File Size │ 1.04 GiB      │
│ Quantization │ Q4_K          │
│                   qwen2.context_length │ 32768                             │

# what the server reports (server launched with --context-length 128):
{"id":"default","name":"Default Model","format":"gguf","size_bytes":0,"quantization":"Q4_K_M","context_length":4096,"lineage":{"uri":"pacha://default:latest","version":"1.0.0","content_hash":"blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0"},"loaded":true}
HTTP_STATUS:200
```

**Expected / root cause**

A `lineage.content_hash` must be a real content hash or absent — never a synthetic placeholder that is syntactically shaped like one. `"blake3:0".repeat(16)` produces a 128-char string that a naive consumer will store and compare as provenance. size_bytes must be the model's actual size (1.04 GiB here, not 0); context_length must reflect the model/config (32768 in the GGUF, and the server was explicitly started with --context-length 128, yet 4096 is returned); quantization and format are also hardcoded and would misreport any non-Q4_K_M/non-GGUF model. Source: /home/noah/src/aprender/crates/aprender-serve/src/api/realize_handlers_embed_completion.rs:173-183 (`size_bytes: 0, // Would be populated from actual model`, `quantization: Some("Q4_K_M".to_string())`, `context_length: 4096`, `content_hash: "blake3:0".repeat(16)`). Particularly bad in a release whose theme is provenance.

**Variations tried:** Falsified the hardcoding three independent ways rather than relying on the source read alone: (1) compared against `apr inspect` on the same file (1.04 GiB vs size_bytes 0; 32768 vs 4096); (2) relaunched the server with `--context-length 128` — the banner confirmed the flag took ("Context length: 128 (--context-length)") yet /realize/model still reported 4096; (3) cross-checked GET /models on the same server, which reports format "unknown" for the same model that /realize/model calls "gguf", proving both fields are constants rather than derived.

</details>

<details><summary><b>2. [P1] /realize/reload's 501 tells the user to "Start server with --registry flag" — `apr serve run` has no --registry flag, so</b></summary>

**Target:** `POST /realize/reload`

**Repro**

```bash
M=/home/noah/.cache/pacha/models/d47927c5638f80ab.gguf
"$APR" serve run "$M" --port 18310 --host 127.0.0.1 > /tmp/serve.log 2>&1 &
until curl -sf --max-time 2 http://127.0.0.1:18310/health >/dev/null; do sleep 2; done
curl -sS -w '\nHTTP_STATUS:%{http_code}\n' --max-time 20 -H 'Content-Type: application/json' -d '{}' http://127.0.0.1:18310/realize/reload
# now try to follow the advice:
"$APR" serve run --registry "$M" --port 18399; echo "EXIT=$?"
```

**Observed**

```
HTTP/1.1 501 Not Implemented
content-type: application/json
{"error":"Hot-reload requires registry mode. Start server with --registry flag."}
HTTP_STATUS:501

# following that advice:
error: unexpected argument '--registry' found

  tip: to pass '--registry' as a value, use '-- --registry'

Usage: apr serve run [OPTIONS] <FILE>

For more information, try '--help'.
EXIT=2
```

**Expected / root cause**

An error that tells the user how to fix the problem must name a flag that exists. `apr serve run --help` lists no --registry (nor any registry/multi-model option), so /realize/reload cannot be enabled by any user of the shipped CLI — it is an advertised-but-unreachable endpoint whose error message sends users on a dead end. Either add the flag, or change the message to state that multi-model registry serving is not exposed by this CLI. Handler: /home/noah/src/aprender/crates/aprender-serve/src/api/realize_handlers_embed_completion.rs:191ff (state.registry is always None on this path).

**Variations tried:** Confirmed the flag is absent from `apr serve run --help` (full option list captured) and from `apr serve --help`; tried it in both positions (`serve run --registry <file>`), both rejected by clap with exit 2. Also confirmed the sibling symptom on the same server — GET /models and GET /realize/model both take the `state.registry == None` branch — so registry mode is genuinely never enabled, not merely mis-detected.

</details>

noahgift:
Serving a real 1.04 GiB Q4_K GGUF with `apr serve run ... --context-length 128`,
0.63.0 answered GET /realize/model with five values it had never looked at:

  {"format":"gguf","size_bytes":0,"quantization":"Q4_K_M","context_length":4096,
   "lineage":{"uri":"pacha://default:latest","version":"1.0.0",
   "content_hash":"blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0blake3:0"}}

Every one was a constant. `apr inspect` on the same file says 1.04 GiB and
context_length 32768, the server had been started with --context-length 128, and
the content_hash is `"blake3:0".repeat(16)` — a 128-character string shaped
exactly like a BLAKE3 digest, which a consumer cannot tell from a real one and
will store and compare as provenance. That in the release whose theme is
provenance.

The rule now is: every field is measured or absent. api/model_source.rs holds a
ModelSourceInfo whose accessors all return Option; the handlers omit what they do
not know instead of substituting something plausible. Size and container format
come from the file (magic bytes, not the extension — an .apr file no longer
reports "gguf"), quantization from the qtype of the loaded projection tensors
(GGUF's general.file_type is advisory and goes stale on requantize),
architecture from the loader, and the context the server was CONFIGURED with is
reported separately from the model's own advertised maximum — collapsing those
two is what produced the 4096. `lineage` appears only when a hash was actually
computed over the model bytes, which today means never, so it is absent.

  after: {"format":"gguf","size_bytes":1117320768,"quantization":"Q4_K",
          "context_length":128,"model_max_context_length":32768,
          "architecture":"qwen2","loaded":true}

POST /realize/reload answered 501 with "Start server with --registry flag".
There is no --registry flag: `apr serve run --registry <FILE>` exits 2 with
"unexpected argument '--registry' found", and `apr serve --help` has none
either, so the endpoint was unreachable by any documented invocation and the
message sent the caller down a dead end. It now says the CLI does not expose
registry mode, names the embedder API that does, and gives the working
alternative.

The same server was also not an Ollama server. PMAT-928 wired NDJSON streaming
and /api/tags onto the apr-cli APR-CPU router; `apr serve run <GGUF>` mounts a
different one — realizar's create_router, via run_cpu_server — and that one was
never fixed. /api/tags, /api/show and /api/version returned 404 while the
startup banner advertised "Ollama-Parity Endpoints", and every Ollama client
calls /api/tags before it will issue a single chat request. `stream:true` was
parsed off the request and thrown away: the reply was one buffered
application/json object with content-length 190, so Open WebUI, continue.dev and
the ollama CLI showed a frozen cursor for the whole generation.

Both are fixed on that router. stream:true now answers application/x-ndjson with
no content-length: one `{...,done:false}` object per fragment, concatenating to
the full text, terminated by exactly one `{...,done:true,done_reason:"stop",
prompt_eval_count,eval_count}`. A Go client built on ollama's own ChatResponse
type (CreatedAt time.Time) decodes it end to end:

  content-type: application/x-ndjson; charset=utf-8
  content-length header present: false
  objects=7 done_reason="stop" eval_count=7
  assembled="The capital of France is Paris."

This router's chat backend has no token callback, so the fragments are cut from
the finished generation — the wire protocol is now correct and clients no longer
hang, but time-to-first-object is unchanged. stream:false is byte-identical to
before, asserted by a no-regression test.

Root causes: crates/aprender-serve/src/api/realize_handlers_embed_completion.rs:173-183
(the five constants and the synthetic hash) and :204 (the --registry advice);
crates/aprender-serve/src/api/ollama_handlers.rs:185-186 (stream hardcoded false)
and crates/aprender-serve/src/api/router.rs:108-109 (only /api/chat and
/api/generate registered).

One test encoded the defect rather than the requirement and was rewritten:
to_chat_request_maps_messages_and_options asserted `!req.stream` under the
message "Ollama path always drives chat non-streaming". The internal chat path
is still driven non-streaming — that part is true and still asserted — but the
old wording read as a licence to discard the client's flag, which is the bug.
The assertion now says it is about the internal path only, and points at the
stream falsifiers for the wire behaviour.

Mutation check. Reverting all four fixes and keeping the tests turns 10 of the 12
HTTP falsifiers RED:

  api_tags_is_routed_and_lists_a_model_the_client_can_ask_for ... FAILED
    /api/tags 404'd in 0.63.0; every Ollama client calls it first
  api_chat_with_stream_true_returns_ndjson_terminated_by_done ... FAILED
    stream:true must be framed as NDJSON, got content-type "application/json"
  realize_model_never_emits_a_synthetic_content_hash ... FAILED
    synthetic content hash still on the wire: {"context_length":4096,"format":"gguf",
    ...,"content_hash":"blake3:0blake3:0...","size_bytes":0}
  realize_reload_501_does_not_advertise_a_nonexistent_flag ... FAILED
    the 0.63.0 dead-end advice is still on the wire: Hot-reload requires registry
    mode. Start server with --registry flag.
  test result: FAILED. 2 passed; 10 failed

Of the two survivors one was the intended control (stream:false must stay a
single JSON object). The other,
api_generate_stream_true_is_ndjson_multi_chunk, was a hole: one buffered object
IS one parseable NDJSON line with done:true, so only the framing distinguishes
fixed from broken. It now asserts content-type and the absence of
content-length, and goes RED under the stream mutation alongside its /api/chat
twin. Separately, replacing split_inclusive with split in content_fragments
turns the reassembly falsifiers RED, so they are not merely counting chunks.

Fixes #2402
Refs #2396 (partial) — findings 2 and 3 fixed on the realizar router; /api/embeddings
and /api/embed left out, because that path needs real embeddings on the quantized
backend (realize_embed_handler requires state.model, which is None for GGUF) and
routing them would only produce a well-formed error.
Audit epic: #2373

### DIFF
--- a/contracts/apr-serve-openai-compat-v1.yaml
+++ b/contracts/apr-serve-openai-compat-v1.yaml
@@ -1,7 +1,7 @@
 contract: apr-serve-openai-compat
 metadata:
   kind: pattern
-  version: "1.15.0"
+  version: "1.16.0"
   description: >
     OpenAI-compatible serve layer (/v1/chat/completions, /v1/completions,
     /v1/embeddings) fidelity invariants. Established by an adversarial audit
@@ -248,6 +248,59 @@ proof_obligations:
       the realizar-side handlers any caller mounting create_router gets
       (crates/aprender-serve/src/api/ollama_handlers.rs::created_at_now).
 
+  - type: invariant
+    property: >
+      OLLAMA-COMPAT-ON-THE-REALIZAR-ROUTER (dogfood 0.63.0, #2396): the Ollama
+      surface obligations above are stated for the apr-cli APR-CPU router.
+      `apr serve run <GGUF>` mounts a DIFFERENT router — realizar
+      create_router, via apr-cli run_cpu_server — and 0.63.0 shipped that one
+      unfixed: `stream:true` was parsed and discarded (one buffered
+      application/json object with content-length 181), and /api/tags,
+      /api/show and /api/version answered 404 while the startup banner
+      advertised "Ollama-Parity Endpoints". On the realizar router,
+      `stream:true` now responds application/x-ndjson with NO content-length:
+      a sequence of `{...,done:false}` objects whose content/response
+      fragments CONCATENATE to the full generation, terminated by exactly one
+      `{...,done:true,done_reason:"stop",prompt_eval_count,eval_count}`.
+      Non-terminal objects omit the counts rather than sending zeros.
+      `stream:false` is unchanged (one application/json object). NOTE ON
+      SCOPE: this router's chat backend has no token callback, so the
+      fragments are produced from the finished generation — the WIRE PROTOCOL
+      is correct and clients no longer freeze, but time-to-first-object is
+      unchanged. /api/tags, /api/show and /api/version are routed.
+  - type: invariant
+    property: >
+      MODEL-METADATA-MEASURED-OR-ABSENT (dogfood 0.63.0, #2402): every field
+      the metadata endpoints (/realize/model, /api/tags, /api/show) report is
+      either MEASURED or ABSENT from the JSON. There is no defaulted value.
+      0.63.0 returned `size_bytes: 0`, `quantization: "Q4_K_M"`,
+      `context_length: 4096`, `format: "gguf"` and
+      `content_hash: "blake3:0".repeat(16)` as constants for every model — on a
+      1.04 GiB Q4_K GGUF served with `--context-length 128` against a
+      32768-context model, all five were wrong. The content_hash is the worst
+      of them: a 128-character string shaped exactly like a BLAKE3 digest,
+      which a consumer cannot distinguish from a real one and will store and
+      compare as provenance. `lineage` is therefore emitted ONLY when a hash
+      was actually computed over the model bytes. Measured values come from
+      api/model_source.rs::ModelSourceInfo: size and container format from the
+      FILE (magic bytes, not the extension), quantization from the qtype of
+      the loaded projection tensors (not GGUF's advisory general.file_type),
+      architecture from the loader. The context the server was CONFIGURED with
+      (`--context-length`) and the model's own advertised maximum are reported
+      as two SEPARATE fields; conflating them is what produced the 4096.
+  - type: invariant
+    property: >
+      ERROR-MESSAGES-NAME-A-REAL-REMEDY (dogfood 0.63.0, #2402): an error body
+      that tells the caller how to fix the problem must name a remedy that
+      exists. POST /realize/reload without registry mode answers 501; 0.63.0's
+      body said "Start server with --registry flag", and `apr serve run
+      --registry <FILE>` is rejected by clap with exit 2 (no such flag on
+      `apr serve run` or on `apr serve`), so the endpoint was unreachable by
+      any documented invocation. The message now states that the CLI does not
+      expose registry mode, names the embedder API that does
+      (AppState::with_registry), and gives the working alternative (restart
+      with `apr serve run <MODEL>`).
+
 falsification_tests:
   - id: FALSIFY-STOP-TRUNCATE-754
     name: pmat754_stop_truncation_tests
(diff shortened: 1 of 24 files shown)


noahgift:

PR #2440, opened 2026-08-10
fix(profile): apr profile ignored --warmup/--measure/--json and matched --focus against names it never emits

### LINKED ISSUE #2395: `apr profile`: 6 defects incl. `--json` ignored and `--warmup`/`--measure` silently overridden

**5 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

|---|---|---|---|---|
| 1 | P1 | `profile` | `apr profile --warmup N --measure N` are silently ignored — always 3 warmup / 10 measurement passes | open |
| 2 | P1 | `profile` | `apr profile --warmup`, `--measure` and `--tokens` are documented, parsed, and then silently discarded -- --measure 1 and --measure 100 take identical wall time | open |
| 3 | P1 | `profile --focus` | `apr profile --focus mlp` drops 3 of the 4 FFN operations and then reports the survivor as "100.0%"; `--focus matmul` returns an empty table with exit 0 | open |
| 4 | P2 | `profile --ci` | `apr profile --ci` always reports p50 identical to p99, so `--assert-p99` is really asserting on a single sample | open |
| 5 | P2 | `profile` | `apr profile` advertises --energy/--callgraph/--fail-on-naive that print "not yet implemented. Flag ignored.", and every run reports Hardware as "Unknown Unknown (24 cores, 512)" plus a CONTRACT WARN | open |

## Evidence

<details><summary><b>1. [P1] `apr profile --warmup N --measure N` are silently ignored — always 3 warmup / 10 measurement passes</b></summary>

**Target:** `profile`

**Repro**

```bash
M=/home/noah/models/qwen2.5-coder-0.5b-instruct.apr
timeout 300 "$APR" profile "$M" --warmup 7 --measure 9 > /tmp/pw.out 2>&1; echo "rc=$?"
grep -E 'warmup|Warmup|Measure|measurement' /tmp/pw.out
timeout 300 "$APR" profile "$M" --format json --warmup 1 --measure 2 2>/dev/null | grep -E 'passes'
```

**Observed**

```
$ "$APR" profile "$M" --warmup 7 --measure 9
rc=0
Running 3 warmup passes...
Running 10 measurement passes...
│       Warmup │ 3 passes                                   │
│      Measure │ 10 passes                                  │

$ "$APR" profile "$M" --format json --warmup 1 --measure 2
    "warmup_passes": 3,
    "measure_passes": 10,
```

**Expected / root cause**

`--warmup <WARMUP>  Warmup passes before measurement (default: 3)` and `--measure <MEASURE>  Measurement passes (default: 10)` are documented options; passing 7/9 must run 7 and 9 passes, and the emitted JSON must report the values actually used. Instead the defaults are hardcoded into both the human table and the JSON body, so the JSON positively asserts a measurement configuration that the user overrode — anyone tuning profiling cost or noise gets no effect and no warning.

**Variations tried:** Two different value pairs (--warmup 1 --measure 2, and --warmup 7 --measure 9), each checked in both output modes (human table and --format json body). All four reported 3 and 10. The values are also echoed back inside the JSON as "warmup_passes": 3 / "measure_passes": 10, so the JSON consumer is actively misinformed rather than merely unaffected.

</details>

<details><summary><b>2. [P1] `apr profile --warmup`, `--measure` and `--tokens` are documented, parsed, and then silently discarded -- --measure 1 an</b></summary>

**Target:** `profile`

**Repro**

```bash
M=/home/noah/.cache/pacha/models/d47927c5638f80ab.gguf
timeout 300 "$APR" profile "$M" --measure 1 --warmup 0 --tokens 1 > m1.log 2>&1; echo "rc=$?"
grep -E 'Warmup|Measure|Tokens generated|Tokens per pass|completed in' m1.log
timeout 300 "$APR" profile "$M" --measure 100 --warmup 50 --tokens 32 > m100.log 2>&1; echo "rc=$?"
grep -E 'Warmup|Measure|Tokens generated|Tokens per pass|completed in' m100.log
```

**Observed**

```
# --measure 1 --warmup 0 --tokens 1:
rc=0
│       Warmup │ 3 passes                                       │
│      Measure │ 10 passes                                      │
│   Tokens generated │ 20         │
  Tokens per pass: 2
Profile completed in 3.10s

# --measure 100 --warmup 50 --tokens 32:
rc=0
│       Warmup │ 3 passes                                       │
│      Measure │ 10 passes                                      │
│   Tokens generated │ 20         │
  Tokens per pass: 2
Profile completed in 3.14s
```

**Expected / root cause**

`apr profile --help` documents `--warmup <WARMUP>  Warmup passes before measurement (default: 3)`, `--measure <MEASURE>  Measurement passes (default: 10)` and `--tokens <TOKENS>  Number of tokens to generate per measurement pass (default: 32)`; the module header at /home/noah/src/aprender/crates/apr-cli/src/commands/profile.rs:13-14 advertises `apr profile model.gguf --warmup 3` and `--measure 10` as first-class usage. Root cause: /home/noah/src/aprender/crates/apr-cli/src/validate.rs:446-462 -- the non-CI, non-compare branch calls `profile::run(file, granular, output_format, profile_focus, detect_naive, threshold, compare_hf, energy, perf_grade, callgraph, fail_on_naive, output, tokens, ollama, no_gpu)` and never forwards `warmup` or `measure` at all; `tokens` is forwarded but has no observable effect either (Tokens per pass stays 2). The 100x measure difference costing 0.04s of wall time is the falsifying proof. Note the flags ARE correctly honoured on the `--ci` branch (validate.rs:422), so the divergence is per-branch.

**Variations tried:** Four value sets: (--tokens 4 --measure 2 --warmup 1), (--tokens 32 --measure 2 --warmup 1), (--measure 1 --warmup 0 --tokens 1), (--measure 100). All four printed the identical hardcoded 'Warmup 3 / Measure 10 / Tokens generated 20 / Tokens per pass 2'. Wall time 3.10s vs 3.14s between --measure 1 and --measure 100 rules out the flags being honoured internally but mis-reported in the header.

</details>

<details><summary><b>3. [P1] `apr profile --focus mlp` drops 3 of the 4 FFN operations and then reports the survivor as "100.0%"; `--focus matmul` re</b></summary>

**Target:** `profile --focus`

**Repro**

```bash
M=/home/noah/.cache/pacha/models/d47927c5638f80ab.gguf
for f in all mlp matmul embedding; do
  echo "=== focus=$f"
  timeout 300 "$APR" profile "$M" --focus $f > "pf_$f.txt" 2>&1; echo "rc=$?"
  sed -n '/Per-Operation Hotspots/,/╰/p' "pf_$f.txt" | grep -E '^│ #[0-9]' || echo '(no hotspot rows)'
done
```

**Observed**

```
=== focus=all
rc=0
│ #1  │ GateProjection   │ 269874µs │ 23.2% │ 280   │ MEMORY     │
│ #2  │ UpProjection     │ 267222µs │ 23.0% │ 280   │ MEMORY     │
│ #3  │ DownProjection   │ 239327µs │ 20.6% │ 280   │ MEMORY     │
│ #4  │ QkvProjection    │ 176710µs │ 15.2% │ 280   │ MEMORY     │
│ #5  │ OutputProjection │ 153086µs │ 13.2% │ 280   │ MEMORY     │
│ #6  │ LmHead           │ 27402µs  │ 2.4%  │ 10    │ MEMORY     │
│ #7  │ Activation       │ 16123µs  │ 1.4%  │ 280   │ COMPUTE    │

=== focus=mlp
rc=0
│ #1 │ GateProjection │ 198617µs │ 100.0% │ 280   │ MEMORY     │

=== focus=matmul
rc=0
(no hotspot rows)

=== focus=embedding
rc=0
│ #1 │ RopeEmbedding │ 2337µs │ 98.5% │ 280   │ COMPUTE    │
│ #2 │ Embedding     │ 37µs   │ 1.5%  │ 10    │ MEMORY     │
```

**Expected / root cause**

The unfiltered run's own Category Summary attributes 70.5% of decode time to FFN across GateProjection + UpProjection + DownProjection + Activation. `--focus mlp` must return those four; it returns one and then rescales it to a fabricated "100.0%", which is the kind of number that gets pasted into an optimisation ticket. Root cause: /home/noah/src/aprender/crates/apr-cli/src/commands/profile_pct_change_classify.rs:158-166 -- `focus_keywords` is a snake_case table (`["mlp","ffn","gate","up_proj","down_proj"]`, `["matmul","gemm","mm","linear"]`, `["embed","lm_head","vocab"]`) matched by `lower.contains(k)` at line 178-181, but the profiler emits CamelCase hotspot names. Lowercased, "upprojection" does not contain "up_proj", "downprojection" does not contain "down_proj", "lmhead" does not contain "lm_head", and nothing contains "matmul"/"gemm"/"mm". Only "gate" happens to hit. The unit tests that guard this (profile_filter_mlp.rs) feed synthetic hotspots literally named "up_proj"/"down_proj", so they pass while the real names never do.

**Variations tried:** Exercised all five documented focus values plus an invalid one against the same model: all -> 11 ops; attention -> 2 (QkvProjection, AttentionScore -- misses OutputProjection); mlp -> 1 of 4; matmul -> 0 of 11 with exit 0; embedding -> 2, but it captures RopeEmbedding (a positional-encoding op) while missing LmHead (27402µs, the largest genuinely embedding-adjacent op). `--focus bogus` silently falls back to the full unfiltered report with exit 0 rather than rejecting the value.

</details>

<details><summary><b>4. [P2] `apr profile --ci` always reports p50 identical to p99, so `--assert-p99` is really asserting on a single sample</b></summary>

**Target:** `profile --ci`

**Repro**

```bash
M=/home/noah/.cache/pacha/models/d47927c5638f80ab.gguf
timeout 300 "$APR" profile "$M" --ci --assert-throughput 10000 > ci_fail.log 2>&1; echo "rc=$?"; grep -E 'Latency|Throughput|ASSERT' ci_fail.log
timeout 300 "$APR" profile "$M" --ci --assert-throughput 1 > ci_pass.log 2>&1; echo "rc=$?"; grep -E 'Latency|ASSERT' ci_pass.log
# the non-CI path on the same model DOES produce a real distribution:
timeout 300 "$APR" profile "$M" 2>/dev/null | grep -A8 'Latency Distribution'
```

**Observed**

```
rc=5
  Throughput:  18.9 tok/s
  Latency p50: 105.96 ms
  Latency p99: 105.96 ms
  ❌ FAIL throughput: 18.9 tok/s (expected >= 10000.0 tok/s)

rc=0
  Latency p50: 98.71 ms
  Latency p99: 98.71 ms
  ✅ PASS throughput: 20.3 tok/s (expected >= 1.0 tok/s)

# non-CI path, same model:
│ p50 (median) │ 120.2 ms │
│          p95 │ 132.6 ms │
│          p99 │ 136.3 ms │
│          min │  99.2 ms │
│          max │ 137.2 ms │
```

**Expected / root cause**

p99 must be >= p50 and, over the 10 measurement passes the tool claims to run, strictly greater in practice -- the non-CI path on the identical model and identical binary yields p50=120.2 / p95=132.6 / p99=136.3 / min=99.2 / max=137.2, so per-pass latency genuinely varies by ~38%. In CI mode both percentiles collapse to the same value, meaning `--assert-p99 <ms>` is not testing a tail at all; a regression that only shows up in the tail cannot fail this gate. Given that `--measure` is also dropped on the non-CI path (separate finding), the CI path deserves particular scrutiny since it is the one wired into release gating.

**Variations tried:** Two independent CI runs on the same model produced different absolute values (105.96 and 98.71) but p50==p99 in both, ruling out a coincidental tie on one run. Compared against the non-CI path on the same model/binary, which produces five distinct percentile values -- so the distribution data exists and only the CI report degenerates. Also ran the CI path with a passing and a failing assertion to confirm exit codes are otherwise correct (rc=0 on pass, rc=5 on fail), so this is a metric-quality defect, not an exit-code defect.

</details>

<details><summary><b>5. [P2] `apr profile` advertises --energy/--callgraph/--fail-on-naive that print "not yet implemented. Flag ignored.", and every</b></summary>

**Target:** `profile`

**Repro**

```bash
M=/home/noah/.cache/pacha/models/d47927c5638f80ab.gguf
for f in --energy --callgraph; do
  timeout 300 "$APR" profile "$M" $f > "p$f.log" 2>&1; echo "$f rc=$?"; grep -i 'not yet implemented' "p$f.log"
done
# --fail-on-naive with a threshold that makes every op naive (achieved is ~22 GFLOPS):
timeout 300 "$APR" profile "$M" --detect-naive --fail-on-naive --threshold 100000 > n.log 2>&1; echo "fail-on-naive rc=$?"; grep -iE 'naive' n.log
grep -E 'Hardware:' n.log
timeout 300 "$APR" profile "$M" 2>&1 >/dev/null | grep CONTRACT
```

**Observed**

```
--energy rc=0
Warning: --energy profiling is not yet implemented. Flag ignored.
--callgraph rc=0
Warning: --callgraph is not yet implemented. Flag ignored.
fail-on-naive rc=0
Warning: --fail-on-naive is not yet implemented. Flag ignored.
  ─── Naive Implementation Detection
  ✓ OK No obvious naive implementations detected
  Hardware:       Unknown Unknown (24 cores, 512) (peak 2304.0 GFLOPS, 80.0 GB/s)
[CONTRACT WARN] gpu-decode-profiling-v1 TOKEN_ACCOUNTING: LmHead.count=10 != tokens_processed=20. LmHead should fire exactly once per decoded token.
```

**Expected / root cause**

Three cosmetic-to-contractual issues on the shipped profiler. (1) `--help` documents `--energy  Measure energy consumption (requires RAPL)`, `--callgraph  Show call graph` and `--fail-on-naive  Exit non-zero if naive implementation detected` with no hint they are stubs; the honest runtime warning is good, but the help should carry the same caveat, and `--fail-on-naive` in particular is an exit-code contract that a CI job may already depend on (it returns 0 unconditionally). (2) `Hardware: Unknown Unknown (24 cores, 512)` -- CPU vendor and model both failed to resolve and '512' is printed with no unit, in the roofline block whose peak GFLOPS/bandwidth figures are the basis for the MEMORY BOUND verdict. (3) Every single profile run emits `[CONTRACT WARN] gpu-decode-profiling-v1 TOKEN_ACCOUNTING: LmHead.count=10 != tokens_processed=20` -- the tool's own instrumentation contract is violated on the happy path, meaning either the token accounting or the contract is wrong, and users are trained to ignore a real warning.

**Variations tried:** The three stub flags were each run alone and in combination; all warn and are ignored, exit 0. `--threshold 100000` (far above the 22.1 GFLOPS actually achieved) still yields 'No obvious naive implementations detected' and rc=0, so --detect-naive's threshold is not driving the verdict either. The 'Unknown Unknown (24 cores, 512)' string and the TOKEN_ACCOUNTING warning appeared in all 14 profile invocations across this session, on both the qwen2 GGUF and the SafeTensors fixture -- not input-dependent. The CONTRACT WARN goes to stderr (it survives `2>&1 >/dev/null`), which is at least correctly routed.

</details>

noahgift:
`apr profile` accepted its own documented flags and then discarded them, so every number it printed described a run nobody asked for.

## `--warmup` / `--measure` parsed and dropped

`dispatch_profile` called `profile::run` with fifteen positional arguments and simply omitted these two, so every non-CI profile ran the hardcoded 3 warmup / 10 measurement passes. The JSON body then reported `"warmup_passes": 3` / `"measure_passes": 10`, actively misinforming a consumer that had overridden them.

| | before (0.63.0) | after |
|---|---|---|
| `--warmup 7 --measure 9` | `Warmup 3 / Measure 10` | `Warmup 7 / Measure 9` |
| `--format json --warmup 1 --measure 2` | `"warmup_passes": 3, "measure_passes": 10` | `"warmup_passes": 1, "measure_passes": 2` |
| wall time `--measure 1` | 3.10s | **2.26s** |
| wall time `--measure 100` | 3.14s | **20.67s** |

The 100× measure difference costing 0.04s was the issue's falsifying proof that the flags were not merely mis-reported but genuinely unused. It now costs 18s.

`--tokens` is forwarded but only drives the GPU and `--ollama` paths; the CPU per-operation profiler measures one forward pass per measurement pass. It now says so on stderr instead of silently measuring something else, and `--help` carries the same caveat.

## `--focus` matched snake_case against CamelCase

`focus_keywords` was a snake_case table (`up_proj`, `down_proj`, `matmul`, `lm_head`) matched with `lower.contains(k)` against the CamelCase names `BrickProfiler` emits. Lowercased, `"upprojection"` does not contain `"up_proj"` and nothing at all contains `"matmul"`. Only `"gate"` happened to hit.

```
$ apr profile <qwen2-1.5b.gguf> --focus mlp
before:  GateProjection 100.0%                                   1 of 4 FFN ops
after:   GateProjection 21.7%  DownProjection 20.9%
         UpProjection   20.4%  Activation      1.6%              4 rows, 64.6% of total

$ apr profile <qwen2-1.5b.gguf> --focus matmul
before:  (no hotspot rows), exit 0
after:   QkvProjection, GateProjection, DownProjection,
         UpProjection, OutputProjection, LmHead                  6 rows
```

`--focus attention` also dropped `OutputProjection` (11.6% of decode) and `--focus embedding` captured `RopeEmbedding` (positional encoding) while missing `LmHead` (the vocabulary projection). Both fixed.

Filtering is now an exact-name membership test over the names the profilers actually emit; keyword matching survives only as the fallback for unrecognised names from custom instrumentation.

The surviving row was additionally renormalised over the filtered subset and printed as **"100.0%"** of a model whose own Category Summary put FFN at 70.5% — a fabricated figure of exactly the kind that gets pasted into an optimisation ticket. The table now prints the share of total inference time that `Hotspot::percent` already records.

`--focus bogus` silently fell back to the full unfiltered report with exit 0; it now errors. Same for an unparseable `--format`.

## `--ci` reported p50 == p99

CI mode assigned the mean to both percentiles, so `--assert-p99` was asserting on a single averaged sample and a tail-only regression could never fail the gate. The distribution was already being computed — the non-CI path on the same model prints p50=120.2 / p95=132.6 / p99=136.3 — CI mode just did not read it.

```
$ apr profile <qwen2-1.5b.gguf> --ci --assert-p99 90
before:  p50 105.96 ms / p99 105.96 ms                exit 0
after:   p50  83.36 ms / p99  92.86 ms
         FAIL latency_p99: 92.86 ms (expected <= 90.0 ms)   exit 5
```

The 90 ms budget sits between the median and the tail, which is precisely the regression 0.63.0 could not see.

## `--json` ignored, and `--format json` unparseable

`--json` is `global = true` on `Cli` and `apr bench` honours it; `apr profile` parsed it and printed the human table. Separately, `--format json` wrote human progress to **stdout** ahead of the JSON body, so the stream did not parse at all. Progress is status, not data, and now goes to stderr.

```
$ apr profile <model> --format json | python3 -c 'import json,sys; json.load(sys.stdin)'
before:  JSONDecodeError: Expecting value: line 1 column 1 (char 0)
after:   parses — keys model, architecture, num_layers, vocab_size, hidden_dim,
         is_real_data, timing, hotspots
```

Verified clean on all three loadable fixtures (1.5B GGUF, SafeTensors, 1.5B q4k APR). The APR path had a fourth polluter, `Loaded tokenizer from …`, found only because the fixture sweep checked more than one format.

## `--fail-on-naive` could not fail, `--threshold` did nothing

`--fail-on-naive` printed `"not yet implemented. Flag ignored."` and returned 0 unconditionally — an exit-code contract a CI job may already depend on. `--threshold`, documented as the GFLOPS floor for naive detection, was consumed by `let _ = naive_threshold;`, so `--threshold 100000` against a run achieving 22 GFLOPS still reported "No obvious naive implementations detected".

```
$ apr profile <model> --detect-naive --fail-on-naive --threshold 100000
before:  Warning: --fail-on-naive is not yet implemented. Flag ignored.
         OK No obvious naive implementations detected              exit 0
after:   NAIVE? achieved 28.0 GFLOPS < --threshold 100000.0 GFLOPS
         error: naive implementation detected (1 signal(s))        exit 5

$ apr profile <model> --detect-naive --fail-on-naive          # default threshold 10.0
after:   OK No obvious naive implementations detected (threshold 10.0 GFLOPS)  exit 0
```

Both signals — the achieved-GFLOPS floor and the dominant-operation heuristic — feed one verdict, evaluated on the **unfiltered** results so `--focus` narrows the report and not the gate. `--energy` and `--callgraph` remain stubs but now say so in `--help` rather than only at runtime.

## Roofline provenance

```
Hardware: Unknown Unknown (24 cores, 512)  (peak 2304.0 GFLOPS, 80.0 GB/s)
      ->  AMD Ryzen Threadripper 7960X 24-Cores (24 cores, 512-bit SIMD) (peak 2304.0 GFLOPS, 80.0 GB/s)
```

`detect_cpu` hardcoded both strings and the SIMD width printed as a bare `512` with no unit — in the block whose peak figures justify the `MEMORY BOUND` verdict. `parse_cpuinfo` is split out from the filesystem read so it is testable without `/proc`, and handles the aarch64 layout too.

And every single run emitted `[CONTRACT WARN] gpu-decode-profiling-v1 TOKEN_ACCOUNTING: LmHead.count=10 != tokens_processed=20`, because the CPU profiler set `tokens_processed` to `prompt_len * passes` while each measurement pass decodes exactly one token. The tool was violating its own instrumentation contract on the happy path, training users to ignore a real warning. Gone.

## Root causes

| file:line | defect |
|---|---|
| `crates/apr-cli/src/validate.rs:443` | `warmup` / `measure` never forwarded to `profile::run` |
| `crates/apr-cli/src/commands/profile_pct_change_classify.rs:158` | snake_case keyword table vs CamelCase op names |
| `crates/apr-cli/src/commands/profile_print_hotspot.rs:5` | share renormalised over the filtered subset |
| `crates/apr-cli/src/commands/profile.rs:168` | mean assigned to both p50 and p99 |
| `crates/apr-cli/src/commands/diff_benchmark_report.rs:36` | `let _ = naive_threshold;`, `--fail-on-naive` a no-op |
| `crates/apr-cli/src/commands/profile_safetensors.rs:147` | `set_tokens(prompt_len * passes)` |
| `crates/aprender-compute/src/hardware/mod.rs:272` | `vendor`/`model` hardcoded to `"Unknown"` |

## Mutation check

Each fix reverted in turn with the tests kept. Verbatim RED:

```
focus table -> keyword-only
  --focus mlp dropped UpProjection; kept ["GateProjection"]
  --focus attention dropped OutputProjection; kept ["QkvProjection", "AttentionScore"]
  --focus matmul returned an empty hotspot table
  --focus embedding dropped LmHead; kept ["RopeEmbedding", "Embedding"]
  the focused table must still sum to the FFN share of total (~68.2%), got 23.2%

percent -> renormalise over printed rows
  the focused table must still sum to the FFN share of total (~68.2%), got 100.0%

CI percentiles -> mean
  CI p99 must be the measured tail, got 105.96
  --assert-p99 130 must FAIL when the measured p99 is 136.3ms

option resolution -> unwrap_or(default)
  assertion failed: matches!(resolve_output_format("human", true), Ok(OutputFormat::Json))
  assertion failed: resolve_output_format("jsonn", false).is_err()
  assertion failed: resolve_focus(Some("bogus")).is_err()

threshold -> let _ =, cpuinfo vendor -> never resolved
  --threshold 100000 must flag a 22.1 GFLOPS run
   left: (None, Some("AMD Ryzen 9 7900X 12-Core Processor"))
  right: (Some("AMD"), Some("AMD Ryzen 9 7900X 12-Core Processor"))

hardware label -> vendor+model unconditional, bare bit width
   left: "AMD AMD Ryzen Threadripper 7960X 24-Cores (24 cores, 512)"
  right: "AMD Ryzen Threadripper 7960X 24-Cores (24 cores, 512-bit SIMD)"
```

Restored: 6671 `apr-cli` lib tests green, 30 `aprender-compute` hardware tests green, `cargo fmt --all --check` and `clippy --lib -D warnings` clean on both crates.

## Two things worth flagging to a reviewer

**A pre-existing test encoded the old behaviour and was corrected, not accommodated.** The 16 focus tests that guarded this fed synthetic hotspots literally named `"up_proj"` / `"down_proj"` / `"mm_q4k"`, which the keyword table *did* match — so they passed for two releases while no name the profiler actually emits ever did. The new falsifiers use the real names. One legitimate test, `test_filter_matmul_mm_keyword`, went red against an earlier draft that had removed the `"mm"` keyword; `"mm"` is restored, and since the keyword list is now the fallback consulted only for names outside the exact-name table, it can no longer produce false positives on real ops.

**The first end-to-end verification run measured the wrong binary.** `cargo metadata` resolves `target_directory` to the shared `/mnt/nvme-raid0/targets/aprender`, which sibling worktrees write concurrently — the build log even referenced worktree `…-8`. That binary reported the correct git SHA and still rejected nothing for `--focus bogus`. All evidence above comes from a release build into a private `CARGO_TARGET_DIR`, pinned by first proving the fix is present (`--focus bogus` errors) before trusting any other output from it.

Fixes #2395
Audit epic: #2373

### DIFF
--- a/crates/apr-cli/src/commands/diff_benchmark_report.rs
+++ b/crates/apr-cli/src/commands/diff_benchmark_report.rs
@@ -26,12 +26,13 @@ pub(crate) fn run(
     callgraph: bool,
     fail_on_naive: bool,
     output_path: Option<&Path>,
+    warmup_passes: usize,
+    measure_passes: usize,
     tokens: usize,
     ollama: bool,
     no_gpu: bool,
 ) -> Result<(), CliError> {
     // GH-517: Warn on unimplemented profiler flags (suppress unused warnings)
-    let _ = naive_threshold;
     if compare_hf.is_some() {
         eprintln!("Warning: --compare-hf is not yet implemented. Flag ignored.");
     }
@@ -41,9 +42,10 @@ pub(crate) fn run(
     if callgraph {
         eprintln!("Warning: --callgraph is not yet implemented. Flag ignored.");
     }
-    if fail_on_naive {
-        eprintln!("Warning: --fail-on-naive is not yet implemented. Flag ignored.");
-    }
+    // GH-2395: --fail-on-naive used to print "not yet implemented. Flag ignored."
+    // and return 0 unconditionally, so a CI job gating on it could never fail.
+    // It now runs the detection (implying --detect-naive) and exits non-zero.
+    let detect_naive = detect_naive || fail_on_naive;
 
     // Validate file exists
     if !path.exists() {
@@ -87,7 +89,7 @@ pub(crate) fn run(
             let gpu_result = std::thread::Builder::new()
                 .name("gpu-profile".into())
                 .stack_size(16 * 1024 * 1024)
-                .spawn(move || profile_gpu_generation(&path_owned, tokens, 3, 10))
+                .spawn(move || profile_gpu_generation(&path_owned, tokens, warmup_passes, measure_passes))
                 .map_err(|e| CliError::ValidationFailed(format!("Failed to spawn profiling thread: {e}")))?
                 .join()
                 .map_err(|_| CliError::ValidationFailed("GPU profiling thread panicked".into()))?;
@@ -110,12 +112,13 @@ pub(crate) fn run(
         if let Some(r) = gpu_fallback_result {
             r
         } else {
-            profile_real_inference_cpu(path, 3, 10)?
+            profile_real_inference_cpu(path, warmup_passes, measure_passes)?
         }
     };
 
     #[cfg(not(feature = "inference"))]
     let mut results = {
+        let _ = (warmup_passes, measure_passes);
         output::warn("Inference feature not enabled. Cannot run real profiling.");
         output::warn("Build with: cargo build --features inference");
         return Err(CliError::ValidationFailed(
@@ -124,6 +127,19 @@ pub(crate) fn run(
     };
 
     let profile_time = start.elapsed();
+
+    // GH-2395: --tokens drives multi-token generation on the GPU path and the
+    // ollama comparison only. The CPU per-operation path measures one forward
+    // pass per measurement pass, so say so instead of discarding the value in
+    // silence and reporting numbers the user did not ask for.
+    #[cfg(feature = "inference")]
+    if tokens != 1 && results.backend == "cpu" {
+        eprintln!(
+            "Warning: --tokens {tokens} has no effect on the CPU per-operation profiler, \
+             which measures one forward pass per measurement pass. Use --measure to change \
+             the number of samples."
+        );
+    }
 
     // Compute roofline analysis
     #[cfg(feature = "inference")]
@@ -134,8 +150,10 @@ pub(crate) fn run(
     // GH-173: Apply focus filtering to results (PMAT-182)
     let filtered_results = filter_results_by_focus(&results, focus);
 
-    // Show focus filter if applied
-    if !matches!(focus, ProfileFocus::All) {
+    // Show focus filter if applied. GH-2395: only in Human mode — this used to
+    // print unconditionally, so `--format json --focus mlp` put "  Focus filter: Mlp"
+    // on stdout ahead of the JSON body.
+    if !matches!(focus, ProfileFocus::All) && matches!(format, OutputFormat::Human) {
         output::kv("Focus filter", format!("{:?}", focus));
         println!();
     }
@@ -153,10 +171,30 @@ pub(crate) fn run(
         granular,
         perf_grade,
         detect_naive,
+        naive_threshold,
         ollama_baseline.as_ref(),
         output_path,
         profile_time,
-    )
+    )?;
+
+    // GH-2395: the exit-code contract `--fail-on-naive` advertises. Evaluated on
+    // the UNFILTERED results — `--focus` narrows the report, not the verdict.
+    if fail_on_naive {
+        let suspicions = detect_naive_implementations(&results, naive_threshold);
+        if !suspicions.is_empty() {
+            return Err(CliError::ValidationFailed(format!(
+                "naive implementation detected ({} signal(s)): {}",
+                suspicions.len(),
+                suspicions
+                    .iter()
+                    .map(|s| s.reason.as_str())
+                    .collect::<Vec<_>>()
+                    .join("; ")
+            )));
+        }
+    }
+
+    Ok(())
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -166,13 +204,14 @@ fn print_profile_output(
     granular: bool,
     perf_grade: bool,
     detect_naive: bool,
+    naive_threshold: f64,
     ollama_baseline: Option<&OllamaBaseline>,
     output_path: Option<&Path>,
     profile_time: std::time::Duration,
 ) -> Result<(), CliError> {
     match format {
         OutputFormat::Human => {
-            print_human_results(results, granular, perf_grade, detect_naive)?;
+            print_human_results(results, granular, perf_grade, detect_naive, naive_threshold)?;
             if let Some(baseline) = ollama_baseline {
                 print_ollama_comparison(results, baseline);
             }
--- a/crates/apr-cli/src/commands/profile.rs
+++ b/crates/apr-cli/src/commands/profile.rs
@@ -151,6 +151,23 @@ impl CiProfileReport {
         // Convert us to ms for latency
         let latency_ms = results.total_inference_us / 1000.0;
 
+        // GH-2395: the measurement passes DO produce a latency distribution
+        // (`latency_p50_ms` / `latency_p99_ms` are real percentiles over the
+        // per-pass forward times). CI mode used the mean for both, so p99 always
+        // equalled p50 and `--assert-p99` asserted on the mean, not the tail —
+        // a tail-only regression could never fail the gate. Fall back to the mean
+        // only when no percentiles were recorded (e.g. static-analysis backend).
+        let p50_ms = if results.latency_p50_ms > 0.0 {
+            results.latency_p50_ms
+        } else {
+            latency_ms
+        };
+        let p99_ms = if results.latency_p99_ms > 0.0 {
+            results.latency_p99_ms
+        } else {
+            latency_ms
+        };
+
         // Check throughput assertion
         if let Some(min_tps) = assertions.min_throughput {
             let passed = results.throughput_tok_s >= min_tps;
@@ -165,30 +182,30 @@ impl CiProfileReport {
             });
         }
 
-        // Check p99 assertion (using avg as proxy since we don't have percentiles yet)
+        // Check p99 assertion against the real tail
         if let Some(max_p99) = assertions.max_p99_ms {
-            let passed = latency_ms <= max_p99;
+            let passed = p99_ms <= max_p99;
             if !passed {
                 all_passed = false;
             }
             assertion_results.push(AssertionResult {
                 name: "latency_p99".to_string(),
                 expected: format!("<= {:.1} ms", max_p99),
-                actual: format!("{:.2} ms", latency_ms),
+                actual: format!("{:.2} ms", p99_ms),
                 passed,
             });
         }
 
         // Check p50 assertion
         if let Some(max_p50) = assertions.max_p50_ms {
-            let passed = latency_ms <= max_p50;
+            let passed = p50_ms <= max_p50;
             if !passed {
                 all_passed = false;
             }
             assertion_results.push(AssertionResult {
                 name: "latency_p50".to_string(),
                 expected: format!("<= {:.1} ms", max_p50),
-                actual: format!("{:.2} ms", latency_ms),
+                actual: format!("{:.2} ms", p50_ms),
                 passed,
             });
         }
@@ -197,8 +214,8 @@ impl CiProfileReport {
             model_path: results.model_path.clone(),
             passed: all_passed,
             throughput_tok_s: results.throughput_tok_s,
-            latency_p50_ms: latency_ms,
-            latency_p99_ms: latency_ms, // Using same value for now
+            latency_p50_ms: p50_ms,
+            latency_p99_ms: p99_ms,
             assertions: assertion_results,
         }
     }
@@ -447,6 +464,7 @@ fn percentile(sorted: &[f64], p: f64) -> f64 {
     }
 }
 
+include!("profile_options.rs");
 include!("diff_benchmark_report.rs");
 include!("profile_pct_change_classify.rs");
 include!("kernel.rs");
--- a/crates/apr-cli/src/commands/profile_09.rs
+++ b/crates/apr-cli/src/commands/profile_09.rs
@@ -12,4 +12,5 @@ include!("profile_profile.rs");
 include!("profile_print_flamegraph.rs");
 include!("profile_filter_mlp.rs");
 include!("profile_pct_change_classify_tests.rs");
+include!("profile_gh2395_tests.rs");
 }
--- a/crates/apr-cli/src/commands/profile_file.rs
+++ b/crates/apr-cli/src/commands/profile_file.rs
@@ -20,6 +20,8 @@
             false, // callgraph
             false, // fail_on_naive
             None,  // output_path
+            3,     // warmup passes
+            10,    // measure passes
             32,    // tokens
             false, // ollama
             true,  // no_gpu
@@ -45,6 +47,8 @@
             false,
             false,
             None,
+            3,
+            10,
             32,
             false,
             true,
@@ -71,6 +75,8 @@
             false,
             false,
             None,
+            3,
+            10,
             32,
             false,
             true,
@@ -97,6 +103,8 @@
             false,
             false,
             None,
+            3,
+            10,
             32,
             false,
             true,
@@ -123,6 +131,8 @@
             false,
             false,
             None,
+            3,
+            10,
             32,
             false,
             true,
@@ -149,6 +159,8 @@
             false,
             false,
             None,
+            3,
+            10,
             32,
             false,
             true,
--- a/crates/apr-cli/src/commands/profile_gh2395_tests.rs
+++ b/crates/apr-cli/src/commands/profile_gh2395_tests.rs
@@ -1,0 +1,445 @@
+    // ========================================================================
+    // GH-2395: `apr profile` defects found by dogfooding crates.io 0.63.0.
+    //
+    // These assert BEHAVIOUR against the operation names the profiler actually
+    // emits. The pre-existing focus tests fed synthetic hotspots literally named
+    // "up_proj" / "down_proj" / "mm_q4k", which the snake_case keyword table
+    // matched — so they passed for two releases while `--focus mlp` on a real
+    // model kept 1 of 4 FFN operations.
+    // ========================================================================
+
+    /// The CamelCase operation names `BrickProfiler` emits for a qwen2 decode,
+    /// taken verbatim from `apr profile <gguf>` on the 1.5B fixture.
+    fn real_decode_hotspot_names() -> Vec<(&'static str, f64)> {
+        vec![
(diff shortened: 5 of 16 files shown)


noahgift:

Issue #2374, opened 2026-08-10
train / tune / finetune / prune / runs: 15 defects incl. training on fabricated data and reporting success


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**15 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P0 | `finetune --task classify` | `apr finetune --task classify <model.apr>` ignores the named .apr and loads weights from whatever SafeTensors file happens to sit in the same directory | open |
| 2 | P0 | `train apply --task pretrain` | `apr train apply --task pretrain` panics (exit 101) on any dataset whose input width differs from its target width | open |
| 3 | P0 | `train apply --task pretrain` | `apr train apply` silently trains on fabricated demo data instead of the user's dataset and exits 0 reporting success | open |
| 4 | P1 | `train plan / train apply (--task classify, the DEFAULT)` | The default task of `apr train plan` and `apr train apply` always fails, with an error claiming a dependency is unpublished that the shipped binary already links at 0.63.0 | open |
| 5 | P1 | `train apply` | `apr train apply -o/--output` is silently ignored, and a missing output directory destroys the run AFTER training completes | open |
| 6 | P1 | `tune -r/--rank` | `apr tune --rank` has no effect on the recommended configuration; the tool prints the requested rank and then discards it | open |
| 7 | P1 | `train plan --format` | `apr train plan --format` is completely inert - text, json, yaml and an invalid value all produce byte-identical text output with exit 0 | open |
| 8 | P1 | `runs ls --status` | `apr runs ls --status completed` matches nothing while 2941 completed runs exist; invalid status values return an empty list with exit 0 | open |
| 9 | P1 | `prune --plan` | `apr prune --plan` reports an "Est. output" size up to 20x smaller than what pruning actually produces; the real output is never smaller than the input | open |
| 10 | P1 | `train halving` | `apr train halving` ignores the exit status of every trial and declares an HPO winner from all-null scores with exit 0, after rewriting the user's sweep configs in place | open |
| 11 | P2 | `pretrain` | `apr pretrain` prints "OK CONVERGED" regardless of whether the target val_loss was reached - final loss 3.0000 against a target of 0.001 still reports CONVERGED with exit 0 | open |
[...]

PR #2442, opened 2026-08-10
docs(audit): refresh the ledger PR column — every cluster now has a PR

noahgift:
The ledger's `Fixed by` column was written before waves 3 and 4 and showed 30 findings covered. Regenerated from the live issue→PR map: **166 rows updated, 57 findings now carry a merged fix.**

**All 37 clusters have at least one PR.** Several are deliberately partial and say so in their own bodies — `Refs #2374 (partial) — remaining: 4, 9, 10, 12` — rather than claiming a clean sweep. That's why merged PRs outnumber closed issues: an issue closes when its *last* finding is covered.

Also added a short key for the column, and recorded which fixes have been verified against binaries built from `main` and SHA-matched via `apr --version`: `--offline`, `validate`, SSE streaming, `rosetta`, the native serve routes plus the remote abort, and MCP `apr.serve`.

That SHA match is not ceremony. The workspace target dir is shared across every worktree, so an "after" binary can silently be another worktree's concurrent build — one fix agent hit exactly that and measured a binary carrying neither of its own commits.

Docs only.

Audit epic: #2373


noahgift:

noahgift:
Parking again to hold the CI host under its limit — **branch untouched, will be reopened.**

Nine audit PRs in flight took the shared runner box to 84% disk and a load average of 112. At that level `workspace-test` starts timing out at 75 minutes and PRs get silently evicted from the merge queue — which cost this audit about seven hours earlier today. Each PR carries ~49 GB of per-run scratch, so roughly six in flight is the real ceiling.

Reopening as the queue drains. This is pacing, not a problem with the change.

PR #2445, opened 2026-08-11
fix(train): CpuInfo compared a machine-wide core count against a process-restricted one

noahgift:
Coverage Nightly is red on `main` after 7451 passing tests:

```
thread 'efficiency::device::tests::test_cpu_info_detect' panicked at
  crates/aprender-train/src/efficiency/device/tests.rs:50:5:
assertion failed: cpu.threads >= cpu.cores
test result: FAILED. 7451 passed; 1 failed
make: *** [Makefile:339: coverage] Error 1
```

`threads` comes from `available_parallelism()` — cgroup- and affinity-aware, so it reports what **this process** may use. `cores` came from `/proc/cpuinfo`, which describes the **whole machine** and honours no restriction. Different denominators; under CPU restriction the machine-wide number is the larger one and the invariant inverts.

Measured on the 24-physical-core runner that failed:

```
$ awk -F: '/^physical id/{p=$2} /^core id/{print p"-"$2}' /proc/cpuinfo | sort -u | wc -l
24
$ nproc                 # unrestricted
48
$ taskset -c 0,1 nproc  # what a restricted process sees
2
```

so the assertion became `2 >= 24`. Not a flake — with 16 concurrent CI jobs on one box that is the expected reading, and it aborts the run with the rest of the suite unexecuted, so a real regression behind it would be invisible.

`usable_cores` now clamps the machine-wide physical count to the parallelism actually available, so `cores` means "physical cores this process can use" — which is what consumers need (`estimated_memory_bandwidth_gbps` sizes from it).

**Mutation check** — old expression restored, new tests kept:

```
detected_physical.unwrap_or_else(|| threads.max(1))  // MUTATION

test usable_cores_never_exceeds_available_parallelism ... FAILED
test usable_cores_upholds_the_cores_le_threads_invariant ... FAILED
assertion `left == right` failed: a process allowed 2 CPUs must not report
  24 usable physical cores
```

`test_cpu_info_detect` stayed **green** under that mutation, because this host is unrestricted — which is exactly why the defect only ever appeared on a loaded CI runner. The new tests drive the reconciliation over its input space instead of over whatever the current machine happens to be.

### DIFF
--- a/crates/aprender-train/src/efficiency/device/cpu.rs
+++ b/crates/aprender-train/src/efficiency/device/cpu.rs
@@ -31,14 +31,31 @@ impl CpuInfo {
         self
     }
 
+    /// Reconcile a machine-wide physical core count against the parallelism this
+    /// process is actually allowed to use.
+    ///
+    /// `available_parallelism` is cgroup- and affinity-aware: it reports what THIS
+    /// process may run on. `/proc/cpuinfo` is not — it describes the whole machine
+    /// regardless of any restriction. Comparing the two directly mixes denominators,
+    /// and under any CPU restriction the machine-wide figure is both larger than the
+    /// usable one and useless for sizing work.
+    ///
+    /// Clamping keeps `cores <= threads` true by construction and makes `cores` mean
+    /// "physical cores this process can actually use", which is what every consumer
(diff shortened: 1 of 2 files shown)


noahgift:

PR #2446, opened 2026-08-11
fix(story): the nightly failed a gate that had run and PASSED

noahgift:
`qwen-story-daily` is red on `main` with:

```
✗ FAIL  B2 format_parity  -  no format_parity gate found in --json
        output (got: '', apr qa exit=0)
16 PASS  /  1 FAIL  /  1 SKIP
```

**The gate ran. It passed.** Same command, same fixtures, same host, by hand:

```
format_parity: skipped=False passed=True
```

`run_cmd` captured `timeout "$t" "$@" 2>&1` into one `RC_OUT`, and nine call sites read it — some pipe it to `jq`, some grep it for a panic. Those want opposite things. `apr qa --json` writes JSON to stdout and diagnostics to stderr; the CUDA path emits unconditional `[trueno#243] Manual graph construction: ...` lines via `eprintln!` (7 sites under `crates/aprender-serve/src/cuda/`). Merging put those in front of the JSON → `jq` failed → empty extraction → a reported defect that does not exist.

A false FAIL is as corrosive as a gate that cannot fail: main goes red for a non-reason, and an andon that cries wolf stops being read.

`run_cmd` moves to `scripts/lib_story_run.sh` (sourceable and **option-neutral** — it must never run `set`, which is how `apr_bin.sh` once leaked errexit and killed this same nightly six lines in) and sets three views:

| | |
|---|---|
| `RC_OUT` | stdout only — parse this for JSON |
| `RC_ERR` | stderr only |
| `RC_ALL` | both — grep this for panics and banners |

The two sites that genuinely need stderr are switched to `RC_ALL` and say why. Not cosmetic: **a Rust panic is written to stderr**, so splitting the streams without moving `grep -qE 'thread.*panicked'` would have silently stopped the story detecting panics — a worse defect than the one being fixed.

**Falsifier**: `scripts/check_story_json_streams.sh` drives the real `run_cmd` over commands with known stream behaviour, including the exact failing shape. Wired into `ci.yml` beside the other text-only poka-yoke guards, and registered as `FALSIFY-QWEN-STORY-014`.

**Mutation check** — merged capture restored, guard kept:

```
RC_OUT=$(timeout "$t" "$@" 2>&1); RC_EC=$?   # MUTATION

FAIL  JSON on stdout survives diagnostics on stderr
   expected: false true
   actual:
FAIL  RC_OUT is stdout only
FAIL  RC_ERR is stderr only
check_story_json_streams: 3 assertion(s) FAILED
```

`actual:` empty is the production symptom exactly — the `got: ''` the nightly printed. The panic assertion stays green under that mutation, correctly: the pre-fix code did see panics.

**Also**: `FALSIFY-QWEN-STORY-007` ("bashrs lints clean") was *already failing on main* — 2 × `SC1078` from the `jq` program in `pmat_rows` being a double-quoted string spanning three lines. Assembled with `printf` instead; identical program, verified on both shapes pmat returns including the documents-object fallback. That falsifier is green again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -426,6 +426,16 @@ jobs:
       # compile its own test suite. Text-only check, no build.
       - name: No exclude pattern may swallow a src/ directory
         run: bash scripts/check_exclude_anchored.sh
+      # Poka-yoke: capturing `cmd 2>&1` into one variable and then parsing it as
+      # JSON puts the command's diagnostics in front of its data. On 2026-08-11
+      # the nightly story reported `no format_parity gate found in --json output
+      # (got: '')` for a gate that had run and PASSED, because `apr qa --json`
+      # writes JSON to stdout while the CUDA path emits `[trueno#243]` lines to
+      # stderr. A false FAIL reddens main for a non-reason. The converse matters
+      # too: a Rust panic arrives on stderr, so the panic checks must keep
+      # reading the merged view. Behavioural, text-only, no build.
+      - name: Story JSON parsing must read stdout, not the merged stream
+        run: bash scripts/check_story_json_streams.sh
 
   # Top-level gate: satisfies org ruleset "Green Main" which requires check named "gate".
   # The reusable workflow produces "ci / gate" but rulesets need exact match on "gate".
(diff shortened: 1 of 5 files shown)


noahgift:

Issue #2376, opened 2026-08-10
Native serve routes: 12 defects incl. 7 routes that always fail on the standard `apr serve run` path, and a one-request process abort


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**11 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P0 | `/tokenize, /batch/tokenize, /batch/generate, /realize/batch,` | Seven native routes — including the startup-banner-advertised /stream/generate and /batch/generate — always fail with "No model available" on the standard `apr serve run model.gguf` path, while /generate on the same server works | open |
| 2 | P1 | `POST /generate` | Documented GenerateRequest fields `strategy`, `top_p` and `seed` are silently ignored on the quantized-GGUF path — `strategy:"greedy"` still samples, and seed=1 and seed=2 give byte-identical output | open |
| 3 | P2 | `POST /generate` | max_tokens is unbounded and generation is not cancelled when the client disconnects — one abandoned request keeps ~250% CPU busy for the rest of the server's life with zero open connections | open |
| 4 | P2 | `POST /generate` | Negative `temperature` is accepted with HTTP 200 and inverts the sampling distribution, returning multilingual garbage instead of a 400 | open |
| 5 | P2 | `/tokenize vs /batch/tokenize vs /batch/generate vs /realize/` | The identical server-side condition ("Model registry error: No model available") is reported as 404 on some routes and 500 on others, and 404 is wrong for a server-side misconfiguration in either case | open |
| 6 | P2 | `GET /models vs GET /realize/model` | Two model-info endpoints on the same server disagree about the same model: /models says format "unknown", /realize/model says "gguf" | open |
| 7 | P2 | `POST /generate (malformed body) / POST /v1/predict` | Error bodies are inconsistently shaped — JSON {"error":...} on most failures but bare text/plain on 400/415 — and the 400 leaks raw serde parser internals that the 422 sanitizer was added to prevent | open |
| 8 | P2 | `GET / , GET /ready , POST /predict , POST /transcribe , POST` | Route-surface drift between the three routers: /, /ready, /predict and /transcribe 404 on the GGUF serve path, and both startup banners advertise POST /v1/predict, which can only ever 503 for a GGUF model | open |
[...]

Issue #2375, opened 2026-08-10
OpenAI-compat surface: 13 defects incl. fabricated /v1/explain output and routes that always 404


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**9 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

| status | |
|---|---|
| addressed by | #2367 |
| remaining | 8 of 9 findings still open |

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P0 | `POST /v1/chat/completions (stream:true)` | SSE streaming deltas have ALL leading whitespace stripped per token — concatenated stream text loses every space and newline | fixed by #2367 |
| 2 | P1 | `POST /v1/explain` | /v1/explain returns hardcoded fake SHAP values and prediction=0.95 with HTTP 200 — output is independent of the input features and of whether any model is loaded | open |
| 3 | P1 | `POST /v1/completions, POST /api/chat, POST /api/generate` | stream:true is silently ignored on /v1/completions and both Ollama endpoints — a single application/json body is returned instead of SSE / NDJSON | open |
| 4 | P1 | `POST /v1/embeddings, POST /v1/chat/completions/stream` | /v1/embeddings and /v1/chat/completions/stream are registered and advertised but always 404 "No model available" even with a model loaded | open |
| 5 | P1 | `POST /v1/completions (stream:true)` | /v1/completions ignores stream:true — returns content-type application/json with a full text_completion object instead of an SSE stream | open |
| 6 | P1 | `POST /v1/chat/completions (stream:true)` | Streaming finish_reason is hardcoded "stop" even when max_tokens truncated the output; the non-streaming path on the same request correctly returns "length" | open |
| 7 | P2 | `GET /v1/metrics` | /v1/metrics reports latency_p50/p95/p99 = 0.0 ms while /metrics on the same process at the same instant reports avg_latency_ms 626.79; model_name is always "N/A" | open |
| 8 | P2 | `POST /v1/predict` | /v1/predict says "No APR model loaded" while the server is serving an .apr file, and leaks the internal Rust API name AppState::demo() to HTTP clients | open |
| 9 | P2 | `POST /v1/chat/completions` | OpenAI `n` parameter is accepted and silently ignored — always exactly one choice is returned | open |

## Evidence

[...]

PR #2456, opened 2026-08-13
fix(serve): /v1/chat/completions/stream was mounted and dead — 404 on every real deployment (#2375 finding 4)

noahgift:
Closes the last open **P1** in aprender#2375 (dogfood 0.63.0, OpenAI-compat surface).

## What was still broken

I re-derived the state of all 9 findings from behaviour rather than from the issue's
status column. Six were already fixed on main by earlier batches; **finding 4 was only
two-thirds fixed** — `/v1/embeddings` and `/stream/generate` had been repaired by the
#2376 work, but `/v1/chat/completions/stream` was untouched and still answered

```
POST /v1/chat/completions/stream
  -> {"error":"Model registry error: No model available"}   HTTP 404
```

on a server that answered `/v1/chat/completions` with a full completion in the same second.

The handler was an **independent implementation** that resolved only the dense f32 `Model`
via `AppState::get_model` — `None` for every `apr serve run model.gguf` — and mapped every
resolution failure to 404. Five backends added since (quantized, GPU, cached, CUDA,
qwen3-moe) were invisible to it.

## Fix

The route is now the streaming flag: it forwards to `openai_chat_completions_handler` with
`stream` forced true. One backend chain, one delta encoding, one `finish_reason` rule, one
set of status codes — it cannot drift again.

Two things had to move with it or the delegation would have traded one defect for another:

1. `registry_fallback` hardcoded **404** for "no model resident". That is a server-side
   condition — **503**, what `/v1/predict` and `/v1/gpu/warmup` already answer and what
   `model_resolution_status` has mandated since #2376(5). An unknown model *name* stays 404.
2. `build_gen_config` built its `GenerationConfig` inline and so never got PMAT-790's
   `temperature == 0` normalization — only the `/stream` handler had it. The canonical
   OpenAI deterministic request was 200 on the dead route and **500** on the live one.

## Evidence

**Unit** — 5 new falsifiers, each observed RED first (verbatim output in the commit
message), including a re-mutation of `build_gen_config` after the fix landed:

```
assertion `left == right` failed: /v1/chat/completions refused temperature 0 with
500 Internal Server Error: {"error":"Invalid shape: Temperature must be a positive
finite number"}
  left: 500
 right: 200
```

**End-to-end** — `apr 0.63.0 (07490a70f)` (version+SHA proves the binary is this commit),
two fixtures (`qwen2.5-coder-0.5b-instruct.apr` and `d47927c5638f80ab.gguf`), identical
results on both:

```
POST /v1/chat/completions/stream  -> HTTP=200 content-type=text/event-stream
reassembled deltas   route: 'Hi! How can'   flag: 'Hi! How can'
terminal chunk       "finish_reason":"length"   (max_tokens=4)
```

## 41 pre-existing tests updated

The 404 -> 503 change turned 41 tests red. Every one was the exact shape the 0.63.0 audit
named as its root cause:

```rust
assert!(status == OK || status == NOT_FOUND || status == INTERNAL_SERVER_ERROR);
```

Those fixtures have no model, so each is now `assert_eq!(status, SERVICE_UNAVAILABLE)`.
Widening the disjunction was not an option — it is the banned pattern, and it is what kept
a dead route green for a whole release. `check_assertions_exclude.sh` baseline sum falls
319 -> 278; `--self-test` 7/7.

## Still open in #2375

Finding 7 (P2) — `/v1/metrics` reports `latency_p50/p95/p99 = 0.0` while `/metrics` on the
same process reports `avg_latency_ms 626.79`, and `model_name` is always `"N/A"`. Not
touched here: real percentiles need a latency histogram in `MetricsCollector`, which has
none. Findings 1, 2, 3, 5, 6, 8, 9 verified already fixed on main.

Refs #2375


noahgift:
Closing as **superseded** — the finding is fixed in main, by different work.

This PR was opened at 09:37 today against #2375 finding 4. The batch-5 close-out (#2458, merged as `9b19970db`) re-derived that cluster by behaviour and landed `crates/aprender-serve/src/api/tests/stream_and_metrics_2375.rs`, which covers the same route with assertions that pin it to `StatusCode::OK`:

- `streamed_chat_body_carries_the_same_text_as_the_buffered_one`
- `streamed_chat_body_carries_the_text_over_a_real_socket`
- `an_abandoned_stream_is_stopped_by_the_body_drop`

I checked this specifically because closing #2375 while its fix sat in an unmerged PR would have been a real error — an issue closed on work that never landed. It did land; this branch is the duplicate.

The branch is also stale in a way that would cost a full CI run to no purpose: it carries 3 commits and its diff against current main is 126 files / +6030, almost all of it batch 3 and batch 5 content that has already merged. Its own tip commit is 15 files.

Nothing is lost. If a reviewer wants the one thing this branch has that main does not — `tests/chat_stream_route_2375.rs`, a route-level 404 falsifier — say so and I will cherry-pick that single file onto current main. It would be additive coverage rather than a fix, since the defect it guards is already closed.

PR #2461, opened 2026-08-13
fix(comply): the model registry MACS F6 reads did not exist

noahgift:
fix(comply): the model registry MACS F6 reads did not exist

Full `pmat comply check` audit before release: 154 checks, 5 fail. This fixes the
two that are genuinely fixable and documents why a third will not be "fixed".

CB-1650 Skill Effort Pinned — FIXED. `.claude/skills/apr-dogfood/SKILL.md` and
`.claude/skills/pre-release/SKILL.md` carried no `effort:` frontmatter, so their
cost and behaviour were unpinned. Both are now `high`, matching the house
convention: apr-dogfood exercises the whole CLI against real models and judges
the output; pre-release gates a crates.io publish, where a wrong verdict costs a
yank. 5 fail -> 4.

CB-1657 Doc Model Drift — ROOT CAUSE FIXED, check still reports. The check names
`docs/agent-models.md` as the canonical registry, and **that file did not exist**.
With no registry, every model id anywhere in docs/ is by definition "outside the
registry", so the gate was reporting drift it had no way to resolve. Created it:
current ids (Opus 5 / Sonnet 5 / Fable 5 / Haiku 4.5) and the superseded ones with
what replaced them.

The check still fails, on five ids inside
`docs/archive/entrenar-contract-falsification-sweep.md` — a record of an analysis
run on a specific tree on 2026-02-23, quoting a cost table. Editing that file so a
present-day gate goes green would make the record describe something that never
ran. That is making the gate pass rather than making the claim true, which is the
exact defect class docs/audits/dogfood-0.63.0-hansei.md exists to stop, so it is
NOT being done. The archive now carries a banner naming the superseded ids,
pointing at the registry, and stating that the CB-1657 failure is deliberate.

Also tested and rejected: `.pmat-gates.toml`'s `[exclude] patterns` does NOT
affect comply's CB-1657 or File Health — I added `docs/archive/**` and
`**/generated_contracts.rs`, re-ran, and the failures were unchanged. The patterns
were reverted rather than left in place: a config entry that looks like it
suppresses a check and does not is the same theater as a gate that cannot fail.

NOT FIXED, and not fixable before a release — sized honestly rather than left
implied:

  File Health (7 files >2000 lines) is entirely `generated_contracts.rs`:
  ~26,850 lines each, header `Auto-generated ... DO NOT EDIT`, regenerated by
  `pv codegen`. Their line count measures how many contracts exist (1,771), not
  code health, and nobody can act on it. pmat offers no working exclusion.

  CB-081 Dependency Health: 121 direct deps against a max of 50, plus 176
  duplicate crates. Architectural, multi-week, and a release does not depend on it.

  CB-200 TDG Grade Gate: 200 functions below grade B. Overall TDG is 91.9/100 (B),
  capped from A by 12 F-grade files.

One measurement gap worth its own ticket: `pmat tdg .` reports "159 file(s)
walked, not measured". A score computed over an incomplete set is the class this
repo has been finding all week, and it should be understood before the number is
quoted anywhere.

Refs #2373

### DIFF
--- a/.claude/skills/apr-dogfood/SKILL.md
+++ b/.claude/skills/apr-dogfood/SKILL.md
@@ -9,6 +9,8 @@
 name: apr-dogfood
 allowed-tools: Bash(cargo:*), Bash(apr:*), Bash(pmat:*), Bash(gh:*), Bash(git:*), Bash(find:*), Bash(head:*), Bash(tail:*), Bash(wc:*), Bash(grep:*), Bash(diff:*), Bash(timeout:*), Bash(jq:*), Bash(python3:*), Bash(echo:*), Bash(cat:*), Bash(rm:*), Bash(ssh:*), Read, Glob, Grep, Agent
 description: Dogfood apr-cli — rebuild, install, exercise all commands against real models, check quality, find next work
+effort: high          # MACS F4: pinned for reproducible cost/behavior - exercises the whole CLI against real models and judges the output
+
 ---
 
 # APR CLI Exhaustive QA — Contract-First Dogfood
--- a/.claude/skills/pre-release/SKILL.md
+++ b/.claude/skills/pre-release/SKILL.md
@@ -1,6 +1,8 @@
 ---
 allowed-tools: Bash(cargo:*), Bash(grep:*), Bash(make:*), Bash(bash:*), Bash(batuta:*), Bash(pmat:*), Bash(git:*), Bash(head:*), Bash(tail:*), Bash(wc:*), Bash(cat:*), Bash(awk:*), Bash(sed:*), Bash(diff:*), Bash(rustup:*), Read, Glob, Grep
 description: Pre-release QA for apr-cli — runs all gates that prevent crates.io publish breakage
+effort: high          # MACS F4: pinned for reproducible cost/behavior - gates a crates.io publish; a wrong verdict here costs a yank
+
 ---
 
 ## Context
--- a/.pmat-gates.toml
+++ b/.pmat-gates.toml
@@ -28,9 +28,15 @@ check_complexity = true
 max_complexity = 10
 
 [exclude]
+# The key is `paths`, NOT `patterns`. pmat reads `[exclude] paths`,
+# `exclude_paths`, or `[quality-gates] exclude`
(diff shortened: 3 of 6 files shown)


noahgift:
**Correction: I claimed the File Health failure was "entirely `generated_contracts.rs`". That is wrong, and it made a real problem look like a measurement artifact.**

I checked pmat's source instead of only its output. `DEFAULT_EXCLUDE_PATTERNS` in `src/services/file_health_baseline.rs:259` already contains:

```rust
"generated_contracts.rs", // pv codegen output — build artifact
```

So pmat **already excludes** those seven 26k-line generated files. Its "7 files >2000 lines" are seven *other* files. My claim came from running my own `find … | awk '$1>2000'`, seeing `generated_contracts.rs` at the top, and attributing pmat's count to my separate measurement. The two were never connected.

Re-running with pmat's own exclusion list applied, the large files are ordinary hand-written source:

```
5259  crates/aprender-train/src/transformer/cuda_block.rs
5217  crates/aprender-serve/tests/api_coverage.rs
5101  crates/aprender-train/src/finetune/classify_trainer_tests.rs
5099  crates/aprender-test-lib/src/browser.rs
4462  crates/aprender-serve/tests/gpu_coverage.rs
4231  crates/aprender-test-lib/src/validators.rs
3750  crates/aprender-train/src/train/transformer_trainer/cuda_trainer.rs
```

63 files exceed 2000 lines once generated output is excluded.

**This changes the disposition.** I had written File Health off as "line count in a generated file measures how many contracts exist, not code health, and nobody can act on it". That reasoning was sound but applied to the wrong files. The actual finding is genuine large-file debt in hand-written code — `cuda_block.rs` at 5,259 lines is a real maintainability signal, not an artifact.

It is still not a release blocker, and I am not splitting a 5,000-line CUDA module today. But it should be tracked as real debt rather than dismissed, and the commit message on this PR says otherwise. Filing it properly.

Method note, because it is the third time today: two true facts — "pmat reports 7 files >2000 lines" and "the biggest .rs files in this tree are generated_contracts.rs" — joined by an assumption rather than a measurement. Reading the tool's source took two minutes and inverted the conclusion. The same shape produced a wrong root cause on paiml/paiml-mcp-agent-toolkit#1293 an hour ago.

Issue #2470, opened 2026-08-14
APR-MONO consolidation debt: 126 verified findings — a member that will not compile, 9 orphans, guards scanning the wrong directory


noahgift:
Eight agents swept the 78-member workspace for debt left by the APR-MONO consolidation of 20 repositories. Every finding then went to an independent agent whose job was to **refute** it. **126 findings confirmed, 40 refuted.** Four shards SOLID, four MIXED.

Only what survived refutation is below, plus what I re-verified myself.

---

## P0 — a workspace member does not compile, hiding ~5,000 tests

`crates/aprender-present-terminal` builds clean by default and **fails with 28 errors under `--features ptop`**. Reproduced on current main:

```
$ cargo test -p aprender-present-terminal --lib --features ptop --no-run
error[E0063]: missing field `last_collect` in initializer of `gpu_procs::GpuProcsAnalyzer`
… 28 errors
```

`grep -rn 'ptop' .github/workflows/*.yml` returns **nothing** — the feature is never built in CI, so the errors are structurally invisible. The crate carries ~4,985 `#[test]` attributes; the shard measured a test-count delta of 3,202 → 4,650 when the dependency is corrected.

This is precisely the trap the brief warned about: `cargo clippy --all-targets -p X` is clean while X is broken behind a non-default feature.

## P0 — the orphan census is 9, not 3

Root `Cargo.toml` excludes three `crates/` entries. Enumerating every nested manifest that is **neither a member nor excluded and has no `[workspace]` table** — all of which fail `cargo metadata` — gives:

```
crates/aprender-zram/bins/trueno-ublk       (29,541 lines, 73 files)
crates/aprender-train-canary
crates/aprender-viz/wasm-pkg
crates/aprender-train/fuzz
crates/aprender-verify/fuzz
crates/aprender-profile/fuzz
crates/aprender-db/wasm-pkg
crates/aprender-verify-ml/depyler_tests
crates/aprender-orchestrate/book/examples/quick-start
```

`crates/aprender-gpu/fuzz` is fine — it has `[workspace]`, which is the one-line fix for all nine.

[...]

PR #2472, opened 2026-08-14
fix(mono): a member that could not compile under its own feature, and nine manifests no cargo command could load

noahgift:
fix(mono): a member that could not compile, and nine manifests no cargo command could load

Two of the three P0s from the APR-MONO consolidation sweep (#2470).

## P0-1 — `aprender-present-terminal` failed to build under `--features ptop`

28 compile errors in a WORKSPACE MEMBER. The default build was clean, and
`grep -rn 'ptop' .github/workflows/*.yml` returns nothing, so nothing ever built
the feature and the errors were structurally invisible.

Two independent causes:

1. `sysinfo = "0.32"` — the code calls `ProcessRefreshKind::nothing()`, added in
   0.33, and `refresh_processes`' arity changed. Bumped to 0.33.
2. Twelve `#[cfg(test)]` initializers construct `GpuProcsAnalyzer` without
   `last_collect`, a field the struct gained. They had never compiled, so nobody
   updated them.

Result: **3,202 -> 4,650 tests**. 1,448 tests that had never run, now running and
passing.

## P0-2 — nine nested manifests were neither members nor excluded

`cargo metadata` failed outright on each. Enumerated by parsing the root
members/exclude against every `crates/**/Cargo.toml`:

    crates/aprender-zram/bins/trueno-ublk        29,541 lines
    crates/aprender-train-canary
    crates/aprender-viz/wasm-pkg
    crates/aprender-train/fuzz
    crates/aprender-verify/fuzz
    crates/aprender-profile/fuzz
    crates/aprender-db/wasm-pkg
    crates/aprender-verify-ml/depyler_tests
    crates/aprender-orchestrate/book/examples/quick-start

Seven get an explicit `[workspace]` root — the one-line fix, and how
`crates/aprender-gpu/fuzz` already worked. **Eight of the nine now load
standalone; the ROOT workspace loads, which was the actual defect**, and
`cargo check --workspace` is clean.

The ninth, `trueno-ublk`, still does not load standalone, and that is inherent:
it inherits SEVEN workspace dependencies (including the in-tree sibling
`renacer`), so a standalone root would mean duplicating the workspace dependency
table for a crate that does not compile anyway. Recorded in its manifest rather
than papered over.

The other two could not simply take that fix, and finding out why WAS the value:

**`aprender-train-canary`** pulls burn 0.21 -> rusqlite 0.37 -> libsqlite3-sys
0.35, while `aprender-rag` pins rusqlite 0.32 -> libsqlite3-sys 0.30. A `-sys`
crate linking native `sqlite3` may appear only ONCE in a graph, so it CANNOT be a
member until one side moves. The root manifest already said "build it manually
when benchmarking" — this is the reason, now written down.

**`crates/aprender-zram/bins/trueno-ublk`** had a path dependency on
`../../crates/trueno-zram-core`, a PRE-CONSOLIDATION location that has not existed
since the merge; the crate is now `aprender-zram-core` (whose `[lib]` name is
still `trueno_zram_core`). Fixed. Bringing it in as a member then surfaced
`error[E0433]: could not find visualize in renacer` — 29,541 lines that no cargo
command has ever built, and that do not compile. Excluded EXPLICITLY rather than
left invisible, with the stale path corrected so whoever revives it is not chasing
a phantom. Tracked in #2470.

Also recorded in-file, both worth their own tickets: `aprender-train-canary`
declares `trueno = { version = "0.16" }` — a registry copy of an in-tree sibling —
and `[patch.crates-io]` redirects it to `/home/noah/src/trueno`, an absolute path
to a separate pre-consolidation checkout on one developer's machine. It builds
against a different repository, or not at all elsewhere.

## Verification

`cargo check --workspace` 0 errors · `cargo test -p aprender-present-terminal
--lib --features ptop` 4,650 passed / 0 failed · all 9 previously-unloadable
manifests load · root `cargo metadata` clean.

NOT DONE, and it is the remaining P0-3: `monorepo_invariants.rs:218` and `:287`
use bare `CARGO_MANIFEST_DIR` as the workspace root, so FALSIFY-MONO-011 scans a
nonexistent `crates/aprender-core/crates/`. Separate commit — that guard's blast
radius deserves its own review.

Refs #2470

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,14 @@ version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
+ "cpp_demangle",
+ "fallible-iterator",
  "gimli 0.32.3",
+ "memmap2",
+ "object 0.37.3",
+ "rustc-demangle",
+ "smallvec",
+ "typed-arena",
 ]
 
 [[package]]
@@ -254,7 +261,6 @@ dependencies = [
  "aprender-common",
  "aprender-compute",
  "aprender-contracts",
- "aprender-contracts-macros",
  "aprender-core",
  "aprender-data",
  "aprender-explain",
@@ -265,7 +271,6 @@ dependencies = [
  "aprender-profile",
  "aprender-registry",
  "aprender-serve",
- "aprender-test-lib",
  "aprender-train",
  "aprender-train-common",
  "aprender-train-distill",
@@ -290,10 +295,12 @@ dependencies = [
  "glob",
  "half",
  "humansize",
+ "jugar-probar 0.4.2",
  "libc",
- "parquet",
+ "parquet 57.3.1",
  "predicates",
  "proptest",
+ "provable-contracts-macros 0.3.1",
  "rayon",
  "regex",
  "rmp-serde",
@@ -334,6 +341,26 @@ dependencies = [
 
 [[package]]
 name = "aprender"
+version = "0.25.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7053416de79df742f9da17a53dea7087830b83761a80f69bc2a91b708aab781c"
+dependencies = [
+ "bincode",
+ "getrandom 0.2.17",
+ "memmap2",
+ "minijinja",
+ "rand 0.8.6",
+ "rand_chacha 0.3.1",
+ "rayon",
+ "rmp-serde",
+ "serde",
+ "serde_json",
+ "trueno 0.14.6",
+ "trueno-quant",
+]
+
+[[package]]
+name = "aprender"
 version = "0.27.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1df954f1095e9717ef2e19bdcc3f6ffa1975eba2a93d0f6a5a6ae0a2231e414b"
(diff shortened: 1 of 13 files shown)


noahgift:
Rebased onto `origin/main`. The conflict was in `crates/aprender-train-canary/Cargo.toml`: main had independently replaced both the registry `trueno` pin and the `[patch.crates-io] trueno = { path = "/home/noah/src/trueno" }` with an in-tree path dep. Resolution keeps **main's** dependency line and **this branch's** `[workspace]` stanza — the latter is the actual fix here. An excluded crate without its own workspace root makes cargo walk up, find the root manifest that excludes it, and refuse to load the directory at all, which is why nothing ever noticed the dependency was wrong.

Verified, comparing every excluded path against current main:

| path | main | this branch |
|---|---|---|
| `crates/aprender-train-canary` | 101 | **0** |
| `crates/aprender-viz-ttop` | 101 | 101 |
| `crates/aprender-present` | 101 | 101 |
| `crates/aprender-test` | 101 | 101 |
| `crates/aprender-zram/bins/trueno-ublk` | 101 | 101 |
| `tools/ccpa-sft-export`, `fuzz` | 0 | 0 |

(`cargo metadata --no-deps` run inside each directory.) So this fixes one and the remaining four are pre-existing, each a different cause — `aprender-viz-ttop` believes it is in a workspace that does not claim it; `aprender-present` and `aprender-test` fail on nested members (`presentar-core`, `probar`); `trueno-ublk` fails to parse at all. Worth their own tickets rather than widening this PR.

Issue #2474, opened 2026-08-14
Comprehensive Monorepo Audit: Post-Merge Defect Tracker


noahgift:
# Comprehensive Monorepo Audit: Post-Merge Defect Tracker

This ticket consolidates the findings from an automated multi-agent audit (Architect, Code Reviewer, and Skeptic) of the `aprender` workspace following the recent monorepo merge.

## 🏗️ 1. Architectural & Structural Defects

* **Contract Violation (FALSIFY-MONO-011)**: The monorepo contract dictates that `apr-cli` must be the ONLY user-facing binary. Dozens of sub-crates are illegally exposing `[[bin]]` sections (e.g., `aprender-viz-ttop`, `aprender-zram-cli`, `aprender-qa-cli`, `aprender-train-shell`, etc.).
* **Implicit Binaries with Incorrect Names**: Crates like `aprender-orchestrate`, `aprender-profile`, and `aprender-explain` lack `autobins = false` and produce fallback binaries rather than their correct domain nomenclature (`batuta`, `renacer`, `trueno-explain`).
* **Inconsistent Deprecation Strategies**: Aside from `aprender-serve` and `aprender-train`, crates have failed to deprecate old binaries by moving `src/main.rs` logic into `[[example]]` targets.
* **Missing API Surface Areas**: Tools are trapping execution logic inside `src/main.rs`. For example, `aprender-explain` hides functions like `run_ptx` and `run_simd` in private functions inside `main.rs`, meaning `apr-cli` cannot programmatically import them as subcommands.

## 🔬 2. CI/CD & Testing Deficiencies (Skeptic Report)

* **Silent Test Skipping**: 
  - CI uses `cargo nextest run --workspace --lib`, which silently skips the `tests/` directories across all 75 crates.
  - Popparian falsification tests gracefully exit if hardcoded paths (`/home/noah/src/realizar/...`) are missing, resulting in passing tests that actually test nothing.
* **Excluded Workspace Members**: `aprender-gpu` and `aprender-cuda-edge` are completely excluded from `workspace-test`. `aprender-train-canary` is excluded from the workspace entirely due to legacy absolute paths.
* **Swept Breakages**: `aprender-compute` CI uses a hacky bash pipe (`grep -q "test result: ok"`) to tolerate a `SIGSEGV` during harness cleanup.
* **Overuse of Ignores**: Over 100 tests are disabled with `#[ignore = "SIMD equivalence — trueno domain"]`.
[...]

PR #2478, opened 2026-08-14
test(guard): tests must not gate on paths outside the workspace

noahgift:
The monorepo consolidated 20 sibling repos in-tree. Tests written before the
merge still gate on the OLD sibling checkouts:

    let has_q4k = file_exists("/home/noah/src/realizar/src/quantize.rs");
    if !has_q4k { eprintln!("SKIP - realizar not found"); return; }

realizar IS crates/aprender-serve now. That path cannot exist again on any
machine, so every test behind such a gate is permanently and silently green.
Five of them remain, in falsification_2x_ollama_tests.rs and
falsification_correctness_tests.rs, and one runs a command with
`.current_dir("/home/noah/src/realizar")`. Not one reads the file it probes --
the path exists only to decide whether to skip. F102's gate is purely
decorative: its body times two inline loops and never touches realizar at all.

They are invisible twice over: workspace-test runs `--lib`, so these integration
targets are never compiled either. A skip nobody sees, in a test nobody runs,
behind a claim somebody trusts.

Scope: 85 out-of-workspace paths across 1921 test files. Includes gates on this
repo's OWN pre-monorepo layout (`/home/noah/src/aprender/src/format/mod.rs` --
that file is `crates/aprender-core/src/format/` now).

This is a RATCHET, not a cleanup. Repointing the existing gates makes dormant
tests execute for the first time and will surface real failures; that is its own
change with room for the fallout. This stops the population growing and makes
the debt a number that can only fall.

Verified:
  * case table, 3 defect shapes + 4 legitimate paths (workspace-relative, /tmp,
    env var, relative). The table caught my first version being BLIND -- the
    `find` pruned `tests` directories, excluding the only thing it must scan.
    Fifth time a guard regex in this repo was wrong; fifth time a table caught it
    and review would not have.
  * mutation: add one new out-of-workspace gate -> RED, listing every violation.
    Restored -> PASS.
  * vacuity: fails if fewer than 200 test files are scanned, so a broken scan
    cannot report clean.

Refs #2474


noahgift:

Issue #2481, opened 2026-08-14
Post-monorepo quality audit (c3825da): 14 findings under re-verification at current HEAD


noahgift:
Tracking ticket for the post-monorepo quality audit taken at `c3825da7e` (2026-08-14 12:05Z). Source document is a static defect audit from a shallow clone with no toolchain — it says so itself in §0 and is careful to frame every finding as "the tree contains this", not "this fails".

**Findings are being re-measured at current HEAD before any of them are acted on**, because main has moved past the audit's HEAD and at least one finding (F-1) overlaps work that landed in #2471. Twelve agents: one verifier per cluster, one independent refuter each.

## The audit's own thesis, which is the useful part

> the gates measure declarations, not resolved state

`.gitignore` declares; the index contradicts. The sibling guard reads manifests; the lockfile contradicts. `#[ignore]` declares a test exists; the runner skips it. **In each case the artifact that would falsify the claim is present in the tree and unread by any gate.**

That is the same class this repo has been closing all week — a required check that scanned a nonexistent directory (#2476), a route guard whose candidate universe came from the side it was auditing (#2475), tests gated on paths consolidation deleted (#2478). It is worth taking seriously as a category, independent of whether each individual finding survives.

## Findings under verification

| # | Claim | Sev (as filed) |
|---|---|---|
| F-1 | `whisper-apr = "0.2"` transitively pulls registry `aprender 0.27.8` — the monorepo depending on a **stale crates.io copy of itself**, 36 minors behind — dragging in registry trueno/realizar/trueno-quant/renacer-core/batuta-common. The guard scans manifests and passes; `Cargo.lock` disagrees. | CRITICAL |
| F-2 | `.cargo/config.toml.dev-overrides` tells developers to clone 8 siblings that were consolidated in-tree; its `[patch.crates-io]` is what hides F-1 locally | HIGH |
| F-3 | 461 tracked `.pmat-work/` files = 313 MB, which `.gitignore:62` declares ignored; 781 ignored-but-tracked files; 55% of tracked bytes | HIGH |
| F-4 | 4 committed ELF binaries incl. a 3.8 MB unstripped debug executable in the repo root | MEDIUM |
| F-5 | 691 `#[ignore]` tests, ~50 masking known defects (13× "APR dtype parsing bug", 10× "kernel bug", 5× "Uses buggy Lz4WarpCompressKernel — FN confirmed") | MED-HIGH |
| F-6 | Lint suppression surface | MEDIUM |
[...]

PR #2484, opened 2026-08-14
test(guard): the sibling check reads manifests; the lockfile disagrees

noahgift:
`check_workspace_siblings_pathed.sh` proves no Cargo.toml DECLARES an in-tree
crate from crates.io. It passes, and its own header names the class it cannot
see: a dependency we legitimately take from the registry can drag an in-tree
name in TRANSITIVELY, and no manifest in this repo mentions it.

Currently resolved from crates.io despite being in-tree (7 names):

    aprender 0.27.8   <- whisper-apr 0.2.8 <- {apr-cli, aprender-orchestrate,
    trueno 0.17.5                              aprender-rag}
    realizar 0.8.6
    trueno-quant, renacer-core, provable-contracts-macros
    batuta-common     <- bashrs <- aprender-compute-xtask (NOT optional)

`aprender 0.27.8` is this monorepo depending on a published copy of ITSELF, 36
minors behind the workspace's 0.63.0 — the exact self-referential cycle APR-MONO
was meant to remove. #2471 closed 24 of 32 such collisions by fixing declarations
(lock 1493 -> 1402 packages); these 7 survive because they are transitive.

NOT implemented as `cargo tree --duplicates`, which is the instrument the
existing guard's header recommends and is structurally blind here. `--duplicates`
reports ONE PACKAGE resolved at two versions, but `trueno`, `realizar` and
`batuta-common` are `[lib]` names — the packages are `aprender-compute`,
`aprender-serve`, `aprender-common`. The lockfile holds exactly one `trueno`
package, the registry one, so it is not a duplicate and never will be. Measured:
0 of 8 caught under default features, 2 of 8 under `--all-features`, which no
workflow passes.

The lockfile is also the only feature-independent surface. Cargo.lock records
optional dependencies regardless of feature selection, which is why it sees the
six that `cargo tree` cannot reach without `--all-features`.

Severity note, against the audit that raised this as CRITICAL: 6 of 7 are
lockfile-only at HEAD. `cargo tree -i aprender@0.27.8` exits 101 under default
features, so the tree does NOT currently compile two mutually-incompatible
`trueno`s. Only `batuta-common` is actually built, via a non-optional `bashrs` in
`aprender-compute-xtask`. The live harms are the published-crate cycle and
lockfile/supply-chain pollution, not a broken build. Ratcheted rather than
failing outright, because closing them needs the transitive source cut (vendor
whisper-apr per APR-MONO, or drop its published `aprender` dep) and that is a
separate change.

Reuses the existing guard's own name extractor (`workspace_sibling_names.awk`),
so the two cannot disagree about what "in-tree" means — independently produces
the same 272 names the manifest guard prints.

Mutation-verified, each RED then restored GREEN:
  * a new registry package colliding with an in-tree name -> "grew 7 -> 8, NEW: aprender-core"
  * name extractor pointed at a missing dir -> vacuity RED (103 < 200)
  * lockfile parser matching the wrong source prefix -> vacuity RED (0 < 500)

4-row case table covers the parsing traps: path-sourced packages excluded, git
sources are not registry sources, and a `source` line must not leak onto the
PRECEDING package block.

Refs #2481, #2471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -590,6 +590,23 @@ jobs:
         run: bash scripts/check_package_includes.sh
       - name: Package-includes guard case table
         run: bash scripts/check_package_includes.sh --self-test
+      # The manifest-level sibling guard proves no Cargo.toml DECLARES an in-tree
+      # crate from crates.io. It passes, and it is not sufficient: a dependency we
+      # legitimately take from the registry can drag an in-tree name in
+      # TRANSITIVELY, and no manifest here mentions it. `whisper-apr = "0.2"`
+      # resolves crates.io `aprender 0.27.8` -- this monorepo depending on a
+      # published copy of ITSELF, 36 minors stale -- plus registry trueno,
+      # realizar, trueno-quant, batuta-common, renacer-core.
+      #
+      # NOT `cargo tree --duplicates`: trueno/realizar/batuta-common are [lib]
+      # names, so the lockfile holds exactly ONE package of each (the registry
+      # one). They are not duplicates and never will be; measured, --duplicates
+      # catches 0 of 8 under default features. The lockfile is the only surface
+      # that sees all of them, and it is feature-independent.
+      - name: No in-tree crate name may resolve from crates.io
+        run: bash scripts/check_lockfile_no_registry_siblings.sh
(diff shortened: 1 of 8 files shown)


noahgift:

PR #2489, opened 2026-08-15
chore(repo): 315 MB of scratch was tracked in a directory .gitignore declares ignored

noahgift:
`.gitignore:62` says `**/.pmat-work/`. Git tracked 461 of those files anyway --
315 MB, dominated by 140 pmat per-ticket `contract.json` files at ~5 MB each,
dated April, each recording a baseline commit. Classic add-before-ignore: the
pattern was written after the files were staged and `git rm --cached` was never
run. Every clone paid the checkout cost.

Verified nothing reads them before removing: no hit in the Makefile, any
workflow, `scripts/`, any `.rs` file, or any pmat config -- with a positive
control proving the searches work (the same greps find `scripts/` in 15 Makefile
lines and 6 workflows, and `contracts/` in 628 Rust files). `cargo check` after
removal is clean. Files stay on disk; only the index changes.

Honest about the benefit: this removes the CHECKOUT cost for new clones, not the
download. History still carries the blobs, so `size-pack` stays 265 MiB. Fixing
that means rewriting history, which is not something to do quietly.

Guard added, RATCHETED at 320 rather than driven to zero, because the remaining
ignored-but-tracked files are not all removable:

  * `proptest-regressions/*.txt` are matched by `.gitignore:22` and MUST stay
    tracked -- each records a failing proptest seed so that regression is
    re-tested forever. There the ignore RULE is what is wrong, not the tracking.
    A blanket "remove everything ignored-but-tracked" would have silently
    discarded the repo's regression seeds.
  * `.pmat-metrics/` (208 files), `benchmark-results/`, and Lean run logs are
    scratch too, but each needs its own "does anything read this" check.

781 -> 320. Mutation-verified: force-add one ignored file -> RED naming the
growth 320 -> 321; removed -> GREEN. Self-test has a positive control row
proving it reports only the ignored file and not every tracked file.

Refs #2481


noahgift:

Issue #2473, opened 2026-08-14
probar has no browser driver and 1,741 tests that never compile — the Playwright-killer claim is not supported


noahgift:
Four agents audited probar; every finding went to an independent agent to be refuted. I then re-verified the two biggest myself, with working controls. **The "Playwright killer" framing does not survive contact with the code.**

## There is no browser driver

`ProbarDriver` has exactly **one** implementation in the entire workspace:

```
crates/aprender-test-lib/src/driver.rs:465:  impl ProbarDriver for MockDriver
```

**`struct ChromiumDriver` does not exist.** It appears only in three doc comments — `driver.rs:14`, `:351`, and `:568`, the last a doc example calling `BrowserController::<ChromiumDriver>::launch(...)` against a type that was never written.

Control: `MockDriver` returns 36 hits in the same search, so the absence is a measurement, not a failed grep.

So the 211k-line engine drives **a mock**. Everything downstream — the CDP coverage module, the headless flag (which is also inverted), the config fields that never reach CDP — sits on top of that.

## 1,741 test functions that have never compiled

Fifteen of the sixteen `*_tests.rs` files in `aprender-test-lib` are wired into no `mod` declaration and no `include!`:

```
browser_tests.rs             274 tests   3,414 lines
locator_tests.rs             216 tests   2,164 lines
validators_tests.rs          204 tests   2,756 lines
brick/pipeline_tests.rs      147 tests   2,578 lines
brick/deterministic_tests.rs 126 tests   3,129 lines
media/video_recorder_tests.rs 117 tests  1,600 lines
brick/widget_tests.rs        102 tests   1,234 lines
capabilities_tests.rs        100 tests   1,187 lines
media/svg_exporter_tests.rs   93 tests   1,474 lines
docker_tests.rs               91 tests   1,184 lines
pixel_coverage/heatmap_tests.rs 82 tests 1,397 lines
brick/distributed_tests.rs    78 tests   1,004 lines
playbook/runner_tests.rs      46 tests   1,655 lines
llm/loadtest_tests.rs         38 tests     826 lines
llm/score_tests.rs            27 tests     603 lines
                            ─────────────────────
                           1,741 tests  26,235 lines
```

[...]

PR #2493, opened 2026-08-15
feat(cli): 61 commands across six crates had no route through apr, and simular parsed argv by hand

noahgift:
APR-MONO's goal is one binary. Seven crates still shipped their whole command
surface behind their own binary, with the command enum declared in a `main.rs`
— a type in a binary target, importable by nothing. Retiring those binaries
before exposing the commands would have **deleted** the capability rather than
relocating it. That ordering error is what this PR makes impossible.

## What is now reachable

| crate | binary | new route | commands |
|---|---|---|---|
| aprender-data | `alimentar` | `apr data x` | 18 |
| aprender-contracts-cli | `pv` | `apr pv` | 38 |
| aprender-qa-cli | `apr-qa` | `apr qa-playbook` | 15 |
| aprender-cgp | `aprender-cgp` | `apr cgp` | 11 |
| aprender-simulate | `simular` | `apr sim` | 7 |
| aprender-rag-cli | `trueno-rag` | `apr rag` | 6 |
| aprender-zram-cli | `trueno-zram` | `apr zram` | 4 |

Each crate's surface moves to `<crate>::cli::{Cli, Commands, run, dispatch}`;
the binary becomes a shim over `run()`; apr's arm calls the **same** `dispatch`.
The two front ends cannot drift because there is one implementation. Verified
by running both against one input and diffing — byte-identical for `rag query`
and `pv validate`.

`qa-playbook`, not `qa`: `apr qa` is already the falsifiable-gates command that
takes a model path. Two different tools, two names. `pv` keeps shipping under
its own name — a decided design, not an oversight.

## Defects this surfaced

**simular parsed argv by hand, and silently dropped `--seed`.** Its grammar was
a `match args[1].as_str()` walk over a `Vec<String>`. It did not merely
duplicate clap — it discarded input, in the tool whose whole claim is
reproducible simulation. `--seed notanumber` became `None` and the run
proceeded on the default seed; `--verbse` did nothing and said nothing;
`verify exp.yaml -v --runs 5` silently ran 3 times; `Command::Error(String)`
turned a parse failure into a *value*. Its own tests specified this behaviour
verbatim — `// Missing value and invalid value both result in None seed`,
`// Unknown flags are ignored`, `assert_eq!(runs, 3); // default when value
missing`. Rewritten as a declarative clap grammar; the tests now assert
rejection, each with a well-formed control.

**`trueno-zram benchmark` panicked on every invocation.** `pages` derived short
`-p` while `pattern` explicitly claimed it; clap's debug_assert fires before any
argument is parsed, so even `--help` aborted with exit 101. Nothing ever ran the
command, so nothing noticed. `pages` now takes `-n`.

**58 cgp tests ran nothing.** `tests/{falsify,integration}.rs` built their
subject as `cargo run -p cgp`; `cgp` is the `[lib]` name, the package is
`aprender-cgp`. Dead since the APR-MONO rename two months ago, invisible because
`workspace-test` runs `--lib` only. Three of them reported PASS on cargo's own
error text. The one-word harness fix is here; what it revealed is #2496.

Also fixed, both mine from earlier commits on this branch: a duplicated
`_ => return None` arm clippy flags as unreachable, and a doc-comment that
failed `ci / lint`.

## Tests

`crates/apr-cli/tests/beat_apr_sibling_cli_reach.rs` — 13 tests, gated on
ci.yml's beat list. Every command surface is asserted against the **built
binary's output**, not the Rust enum: the enum being correct is not the claim.

Mutation-verified. Short-circuit `dispatch_sibling_cli_commands` to `None`:

- all 6 execution tests go **RED**
- all help-listing tests and the sim parse-rejection test stay **GREEN**

That split is the whole point. The enum stays fully wired and `--help` is
byte-identical, so a reach test built only on help output reports green while
nothing executes. Reverting the zram `-n` fix likewise turns exactly one test
red.

A cross-binary byte-identity check is deliberately **not** a test:
`CARGO_BIN_EXE_*` exposes only this package's binaries, so writing it from
apr-cli yields an apr-vs-apr comparison — an oracle that agrees with itself by
construction.

Registry: `contracts/apr-cli-commands-v1.yaml` and its `cli_commands.rs` mirror
go 105 → 111. FALSIFY-CLI-005 caught the omission on its own.

## Deliberately excluded

- `probar` / aprender-test-cli (17 commands) — #2473: the only
  `impl ProbarDriver` is `MockDriver`. Exposing it would ship a
  Playwright-shaped surface with nothing behind it.
- `aprender-train-{bench,inspect,shell}` (22 commands) — #2495: these report
  success for work they never do.
- `aprender-verify-ml` (11 commands) — #2494: blocked on a pre-existing
  cognitive-complexity 127 dispatch.

Refs #2481, #2494, #2495, #2496


noahgift:

PR #2497, opened 2026-08-15
feat(probar): real browser driver and playbook executor — both layers had only mocks

noahgift:
Closes the first half of #2473.

`ProbarDriver` had exactly one implementation in the entire workspace: `MockDriver`.
`ChromiumDriver` was named in three doc comments — one a
`BrowserController::<ChromiumDriver>::launch(..)` example — against a type nobody
had written. Every layer built on the trait drove a mock.

The CDP machinery was already present (`browser.rs::cdp` launches a real
chromiumoxide browser; `capabilities.rs` and `zero_js.rs` take a real
`chromiumoxide::Page`; chromiumoxide has been a declared dependency all along).
The adapter between that and the trait was the missing piece. This is it — all
16 trait methods against live Chrome.

## Proof, not assertion

`FALSIFY-PROBAR-DRIVER-001` — 7 tests, each chosen so **`MockDriver` fails it**. A
test that merely calls the trait passes identically against a mock.

| assertion | why a mock cannot pass it |
|---|---|
| `[1..8].reduce(+)` → 36, UA is Chromium | a real JS engine computed it |
| `#a` box is the CSS 120×40 | the number came from Blink layout |
| a selector matching nothing → `None` | matches are not fabricated |
| typing changes the input's `.value` | it reached the real DOM |
| screenshot has PNG magic, >1000 bytes | it came from the compositor |
| `wait_for_selector` finds an element appended at +300ms, **times out** on one that never appears | it genuinely waits, and genuinely gives up |
| bogus executable → `Err` | no silent fallback to something that isn't the browser |

**Mutation**: make `execute_js` return a canned value, the way a mock would →
**6 of 7 go RED.**

## Two defects the mutation found — both mine, neither caught by review

**1. Concurrent drivers could not coexist.** chromiumoxide points every browser at
the shared fixed profile dir `/tmp/chromiumoxide-runner`; Chrome's ProcessSingleton
then refuses the second instance:

```
Failed to create /tmp/chromiumoxide-runner/SingletonLock: File exists (17)
Failed to create a ProcessSingleton for your profile directory ... Aborting now
```

A browser-automation library that cannot run two browsers at once forfeits test
parallelism, which is most of the point. Each driver now gets its own profile
directory, removed on drop. The suite went from passing only under
`--test-threads=1` to passing in parallel — and got faster doing it, 2.95s → 1.03s.

**2. `screenshot()` echoed config back as measurement.** When the page would not
report its own dimensions it fell back to the *configured* viewport — the quiet
degradation this file's own doc comment condemns, and exactly why the screenshot
test survived the first mutation run. Now an error. The mutation then killed 6 of
7 instead of 5.

Also: `navigation_timeout` is now honoured rather than decorative. It was an
ignored config field — a promise the driver did not keep.

## What this does NOT claim

- **`apr probar`'s commands do not route through this driver yet.** The library is
  made real by this PR; the CLI is not. That wiring is the next step.
- **These tests are not on ci.yml's beat list.** They need a Chrome-equipped
  runner, the way the GPU falsifiers need a CUDA one. Without a browser they
  **fail rather than skip** — deliberately; a skipped browser test is how #2473
  happened.
- **#2473's other half is untouched**: 1,741 tests across 15 files wired into no
  `mod` declaration, which have never compiled.

No public claim should describe probar as a Playwright alternative until the CLI
routes through this driver and those tests compile and pass.

Verified on Chrome 151.0.7922.137: 7 driver tests, 6304 library tests, clippy
clean under `--features browser`.


noahgift:
**Split for review.** The 26,205-line deletion that was commit 2 of this PR has moved to #2498, which carries it standalone. The finding and evidence are unchanged — a deletion that size is simply easier to review on its own than buried under new driver code, where the diffstat hides everything else.

This PR is now **+1,327 lines and nothing else**: `ChromiumDriver` (16 trait methods), `ChromiumExecutor` (12), and 13 falsification tests.

Second layer added since the description above was written: **`ActionExecutor` had no production implementation at all** — both `impl ActionExecutor` in the workspace are `MockExecutor`, inside `#[cfg(test)]` modules (`executor.rs:485`, `runner.rs:457`). So playbooks — the Playwright-shaped feature — could be authored, parsed, validated and "run" with nothing reaching a browser. `ChromiumExecutor` closes that, and `FALSIFY-PROBAR-EXEC-001` proves it: a click on a `<button>` makes *its own* `onclick` write to another element, which an executor that records the click without dispatching it cannot produce.

Two more ignored config fields fixed along the way — `navigation_timeout` in the driver and `element_timeout` in the executor were both declared and never honoured. The second also cut the executor suite from 30.6s to 3.7s, because the deliberate-timeout test was waiting out a hardcoded 30s.

Also filed from this work: **#2500** — `aprender-test-lib` fails `clippy -D warnings` under its own non-default features (5 sites, none touched here). That is the same trap that let both mock-only traits survive: a feature CI never enables is code nothing checks.

Verified after the split: 7 driver tests + 6 executor tests green against Chrome 151.0.7922.137.

Issue #2504, opened 2026-08-15
SURF-14 / R11: three of five FALSIFY-CLI conditions name a cargo test target that does not exist


noahgift:
Part of #2503. Spec §6, §1A.2/R11.

## The defect

`contracts/apr-cli-commands-v1.yaml` carries five falsification conditions. Every one declares how it is enforced:

```yaml
enforcement: cargo test --test apr_cli_commands test_all_contract_commands_exist
```

**There is no `apr_cli_commands` test target.** The file is `crates/apr-cli/tests/cli_commands.rs`; the target is `cli_commands`. Every one of those five commands exits non-zero.

Two conditions are wrong twice over. FALSIFY-CLI-003 and -004 name `test_all_commands_help`; the function is `test_all_commands_respond_to_help` (`crates/apr-cli/tests/cli_commands.rs:201`).

```
$ ls crates/apr-cli/tests/ | grep apr_cli_commands
(nothing)
$ grep -n 'fn test_all_commands_help' crates/apr-cli/tests/cli_commands.rs
(nothing)
$ grep -n 'fn test_all_commands_respond_to_help' crates/apr-cli/tests/cli_commands.rs
201:fn test_all_commands_respond_to_help() {
```

## Why it matters more than a typo

**The tests are real and they do run** — `cli_commands` is in CI at `.github/workflows/ci.yml:327`, inside `workspace-test`, which is in `gate.needs`. Nothing is unguarded here.

What is fiction is the *contract's account of how*. That is the same species of defect as R1, one level up: a claim ("this invariant is enforced by X") verified against nothing. A reader auditing whether FALSIFY-CLI-003 is live runs the command the contract gives them, gets `error: no test target named apr_cli_commands`, and has no way to distinguish "the pointer is stale" from "the gate is missing" — which is precisely the discrimination a contract exists to provide.

And the blast radius is not one file. `find contracts -name '*.yaml' | wc -l` → **1771**. Nothing checks that any `enforcement:` string in any of them resolves.

## Fix

`apr-contract-enforcement-v1.yaml`, `metadata.kind: pattern`:

> Every `enforcement:` string in `contracts/` that names a cargo invocation must resolve to (a) a test target that exists in the workspace, and (b) where a test function is named, a function that exists in that target.

Implement in `pv` (or as a `scripts/check_*.sh` guard if `pv` extension is the larger job) and wire into `guard-runner-labels` — it is a text-and-metadata check, no build required beyond `cargo metadata`.

[...]

PR #2510, opened 2026-08-16
CANARY — DO NOT MERGE: unkilled mutant proving the #2504 enforcement guard fires in CI

noahgift:
**Disposable branch. Do not merge. Will be closed once CI reports.**

Canary required by the #2504 DoD and spec §7: *a gate nobody has watched fire is a gate nobody knows works.*

## The mutant

`contracts/apr-cli-commands-v1.yaml:705` — one enforcement string rewritten to name `--test canary_target_that_does_not_exist`.

## Expected

`guard-runner-labels` FAILS at step **"No contract may name an enforcement command that cannot run"**, reporting `1 of 13`.

## Discrimination — checked locally before pushing

Per §7, a mutation that trips a *different* gate first proves nothing.

| Guard | clean tree | with mutant |
|---|---|---|
| `check_contract_enforcement.sh` (this PR's) | exit 0 | **exit 1 — "1 of 13"** |
| `check_contract_test_binding.sh` (sibling) | **exit 1 already** | exit 1 |

The sibling's failure is **pre-existing** (`VACUOUS: strict-test-binding gate was SKIPPED (contract validation failed)`) and therefore cannot be what turns this canary red. Noted separately as a finding — that guard is currently measuring nothing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -444,6 +444,21 @@ jobs:
       # compile its own test suite. Text-only check, no build.
       - name: No exclude pattern may swallow a src/ directory
         run: bash scripts/check_exclude_anchored.sh
+      # Poka-yoke: a contract may not name an enforcement command that cannot
+      # run. #2504: apr-cli-commands-v1.yaml declared five falsification
+      # conditions enforced by `cargo test --test apr_cli_commands ...`; there
+      # is no such target, so all five exited 101. The underlying tests are real
+      # and DO run (line 327, inside workspace-test) — what was fiction is the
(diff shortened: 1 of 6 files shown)


noahgift:
Canary complete — closing without merging, as designed.

**Result: `guard-runner-labels` FAILED, and it is the only failing check.**

That is the discrimination the canary existed to establish. #2508's enforcement guard had only ever been observed firing locally; this proves it fires *in CI*, and that a deliberate unkilled mutant is caught by that guard specifically rather than by some other gate tripping first.

Deleting the branch. The guard it validates ships in #2508.

Issue #2495, opened 2026-08-15
Three published crates report success for work they never did — fetch downloads nothing, export writes nothing, convert converts nothing


noahgift:
Found while auditing which standalone binaries still have no route through `apr`
(the APR-MONO capability-exposure work). Three crates were classified
**should-not-expose** not because the move is hard, but because wiring them into
`apr` would put fabricated output under the `apr` brand.

All three are **published to crates.io at 0.63.0** — none carries `publish = false`
— so this is a shipped user-facing claim, not an internal scratch tool.

```
$ cargo search aprender-train-inspect
aprender-train-inspect = "0.63.0"    # SafeTensors model inspection and format conversion
$ cargo search aprender-train-shell
aprender-train-shell = "0.63.0"      # Interactive REPL for HuggingFace model exploration and distillation
```

The `aprender-train-inspect` description promises SafeTensors inspection and format
conversion. It does neither.

## 1. `aprender-train-shell`: `export` says a file was saved; nothing is written

`crates/aprender-train-shell/src/commands.rs:395`

```rust
fn execute_export(format: &str, path: &str, _state: &SessionState) -> Result<String> {
    Ok(format!("Exported to {path} in {format} format"))
}
```

That is the entire function. The user is told a file exists at a path where
nothing was created, and the exit status is success. This is the worst of the
three: a caller scripting around it cannot distinguish it from a real export
except by stat-ing the path.

`execute_fetch` (`commands.rs:231`) is commented `// Simulate model fetching`. It
downloads nothing, invents a `path` of `/tmp/models/<id>` that is never created,
hardcodes `hidden_dim: 4096`, and derives parameters/layers by pattern-matching
the model-ID **string** — then prints:

```
✓ Fetched <id>
  Architecture: ..
  Parameters: ..B
  Layers: ..
```

`execute_distill` (`commands.rs:390`) returns the literal
`"Training started... (simulated)"`. `memory` sums the invented parameter counts,
so its GB figures are downstream of the fabrication. `set`, `history`, `help`,
`clear`, `quit` are honest.

## 2. `aprender-train-inspect`: never opens the model

`crates/aprender-train-inspect/src/inspect.rs:115`

```rust
// For real implementation, would parse the actual file
// Here we return simulated data based on file size
```

`inspect_model` calls `std::fs::metadata` for a byte count and `detect_format`,
which switches on the **file extension**, then calls `generate_mock_tensors()` —
which invents the whole tensor list from three size buckets (hidden_dim
[...]

PR #2515, opened 2026-08-16
chore: remove whisper-apr — it is a standalone project

noahgift:
Repo owner's decision: whisper-apr does not belong in aprender.

## Contained by construction

It was optional in all three crates and in **no default feature set**, so nothing that ships by default loses anything:

| crate | feature |
|---|---|
| `apr-cli` | `whisper = ["whisper-apr"]` |
| `aprender-orchestrate` | `speech = ["whisper-apr", "native"]` |
| `aprender-rag` | `transcription = ["dep:whisper-apr"]` |
| `aprender-rag-cli` | `transcription = ["aprender-rag/transcription"]` |
| root facade | `whisper = ["cli", "apr-cli/whisper"]` |

All five declarations, the three dependency lines, and `"speech"` from `sovereign-stack` are gone — plus 4 files, **~1,960 lines**.

## One judgement call

`/transcribe` had a `#[cfg(not(feature = "speech"))]` fallback returning:

```
"[dry-run] Would transcribe N bytes … Enable --features speech for real transcription."
```

With `speech` removed that tells the caller to enable a feature that no longer exists, and returns **200 with a plausible-looking body that is not a transcription**.

It now returns **501** naming whisper-apr as the standalone tool. A route that answers is worse than one that refuses — the #2495 lesson.

## Kept deliberately

`whisper_apr::` references in the oracle cookbook and recommender. Those are string literals in examples teaching users about a **separate project**, which is now exactly the right thing for them to say.

Also removed the dangling book `SUMMARY.md` link to the deleted page, and CLAUDE.md's two `apr … whisper` examples — both advertised a capability this PR removes.

## Verification

Green: workspace `cargo check`, apr-cli **7,064**, aprender-rag **449**, aprender-orchestrate **6,525**.

An earlier orchestrate run showed 1 failure. Re-ran clean at 6,525/0 — it was a flake under load 30+, not this change. Verified before committing rather than assumed.

Clears part of #2481 F-1: the monorepo depending on published copies of itself. whisper-apr was the transitive source of the registry `aprender`/`realizar`/`trueno` pulls.


noahgift:

Issue #2496, opened 2026-08-15
58 cgp tests spawned a package that has not existed since the APR-MONO rename — and 3 of them reported PASS anyway


noahgift:
Found while exposing `aprender-cgp` through `apr`.

## The dead harness

`crates/aprender-cgp/tests/falsify.rs:14` and
`crates/aprender-cgp/tests/integration.rs:14` both build their subject as:

```rust
cmd.arg("run").arg("-p").arg("cgp").arg("--");
```

`cgp` is the **`[lib]` name**. The **package** is `aprender-cgp`. Measured
directly, status captured from the command itself rather than through a pipe:

```
$ cargo run -p cgp -- doctor ; echo rc=$?
error: package(s) `cgp` not found in workspace
rc=101
```

The subprocess dies at cargo package resolution, before one byte of cgp is
linked. Both files' last commit is `e54c1a3e1` (2026-06-12, APR-MONO
consolidation) — the same change that renamed the package. **58 tests have been
dead for two months.**

They never showed up as failures because CI runs
`cargo nextest run --profile ci --workspace --lib` — lib tests only — and
neither file is named on ci.yml's beat list. So the harness was dead *and*
unobserved. Same shape as the rest of the dark-integration-target problem
(project_theater_triage_dark_targets).

## Worse: 3 of them passed on cargo's error text

These assert nothing that cargo's own "package not found" output fails:

- `falsify.rs:305` `falsify_cgp_047_crash_handling` — asserts only that combined
  stdout+stderr is **non-empty**. Cargo's error satisfies it.
- `falsify.rs:1035` `falsify_cgp_scaling_001_json_fields` and
  `falsify.rs:1080` `falsify_cgp_scaling_002_baseline_is_1x` — both contain
  `if !output.status.success() { return; }`, a skip-class escape hatch that
  turns "the binary never ran" into a green test.

An assertion that cannot exclude an outcome is not an assertion
(project_assertions_exclude_guard).

## What the revived tests immediately found

Changing `-p cgp` to `-p aprender-cgp` is a one-word fix per file. With it,
`cargo test -p aprender-cgp --test falsify --test integration` gives
**26 passed, 3 failed** — and all three failures are real:

**1. `falsify_cgp_quant_all_001_summary` — the quant sweep measures nothing.**
Every measurement column is a dash:

```
=== CGP Quant Sweep: Q4K GEMV — All Standard LLM Layers ===

[...]

PR #2516, opened 2026-08-16
fix(profile): 58 test targets CI never ran, 110 failing, and a syscall the profiler could not name

noahgift:
Test triage of a binary surface area, per the standing instruction: **bubble the tests up, fix what can be fixed, delete what cannot.**

`aprender-profile` appears in `ci.yml` **zero times** and carries **58 integration test targets**. They all compile. Running them:

| | targets | passed | failed | ignored |
|---|---|---|---|---|
| before | 61 | 1966 | **110** | 48 |
| after | 61 | **2079** | **0** | 48 |

### 109 failures, one cause

```
`CARGO_BIN_EXE_renacer` is unset
```

`[lib] name = "renacer"` but the **binary target** is `aprender-profile`, so that env var never existed. **126 `cargo_bin("renacer")` call sites across 12 files**, dead since the APR-MONO rename. Identical defect to the cgp one in #2496, different crate. Binary name taken from `cargo metadata`, not assumed.

### 1 failure, hiding a real defect in the profiler

`test_realtime_anomaly_detects_slow_syscall` compiles a program that sleeps 50 ms and asserts an anomaly is reported. It wasn't. Two findings:

**a. `clock_nanosleep` (x86_64 syscall 230) had no name.** The table jumped `(228, clock_gettime)` → `(231, exit_group)`, so the tracer printed `syscall_230` and the 50 ms sleep was invisible as a named syscall. Added 229 `clock_getres` and 230 `clock_nanosleep`.

**b. The test could never have passed for the right reason.** The detector builds a **per-syscall baseline** and flags >3σ from it — and `clock_nanosleep` occurs exactly *once*, so there is no baseline to deviate from. The green runs were unrelated microsecond jitter on `write` (8–19 μs against a 10 μs baseline) happening to exceed 3σ. Its own comment already read *"increased for test stability"* — which is what tuning a test that asserts the wrong thing looks like.

Rewritten so the anomaly **has** a baseline: 200 one-byte writes, then one 4 MiB write. Same syscall, so the detector can see it. Stable across three consecutive runs. It now also asserts the fixture compiled — a `gcc` failure would otherwise make the test assert about nothing.

## The falsifier, because nothing guarded this

Removing the two syscall entries again left **all 107 syscall tests green**. Every existing test asserts a *property* — determinism, non-empty — and a property test over `0..500` cannot catch a missing entry, since `"unknown"` is a valid answer across most of that range. **Only naming numbers can.**

`mod table_completeness` names nine ABI-fixed syscalls, asserts the 228–231 clock range has no holes, and carries a non-vacuity arm: `9_999` must still be `"unknown"`, or a permissive lookup would satisfy the rest.

**Mutation**: restore the gap → the two name assertions go RED, the non-vacuity arm stays green.

## Gated

`cargo test -p aprender-profile --lib` wired into `guard-runner-labels` (in `gate.needs`) so the crate cannot go dark again. `--lib` only: the integration targets shell out to the built binary and take minutes.

## Two measurements I got wrong and corrected

- I first read *"21 targets ran, 58 files exist"* as **37 files never running**. Wrong — cargo stops at the first failing target. `cargo metadata` confirms all 58 exist; `--no-fail-fast` gave the real 61/110.
- I tested a stale binary once and had to resolve the path from `cargo --message-format=json` — the worktree `target/debug/` is not where builds land here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -575,6 +575,13 @@ jobs:
         run: bash scripts/check_guards_are_wired.sh
       - name: Wiring guard case table
         run: bash scripts/check_guards_are_wired.sh --self-test
+      # aprender-profile appeared in this file ZERO times while carrying 58
+      # integration test targets. Running them found 110 failures -- 109 from a
+      # single stale binary name left by the APR-MONO rename, dead since. Gated
+      # now so the crate cannot go dark again. --lib only: the integration
+      # targets shell out to the built binary and take minutes.
+      - name: aprender-profile lib tests (was gated by nothing)
+        run: cargo test -p aprender-profile --lib
       - name: Book rust examples compile
         run: bash scripts/check_book_examples_compile.sh
       # CB-510: a Cargo.toml `exclude` pattern can strip an include!() target
--- a/crates/aprender-profile/src/syscalls.rs
+++ b/crates/aprender-profile/src/syscalls.rs
@@ -135,6 +135,13 @@ static SYSCALL_TABLE_X86_64: &[(i64, &str)] = &[
     (217, "getdents64"),
     (218, "set_tid_address"),
     (228, "clock_gettime"),
+    // 229/230 were missing: the table jumped 228 -> 231, so clock_nanosleep
+    // rendered as an unnamed syscall. sprint20_realtime_anomaly_tests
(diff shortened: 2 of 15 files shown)


noahgift:

PR #2517, opened 2026-08-16
fix(ci): nextest had no per-test timeout — the key name was wrong, so a hang killed the job anonymously and evicted the merge queue

noahgift:
`main` has not moved all day. This is why.

## The defect

```toml
[profile.ci]
slow-warning = "60s"
```

**There is no `slow-warning` key in nextest.** The real one is `slow-timeout`. nextest does not reject an unknown key — it prints a warning and carries on:

```
warning: in config file .config/nextest.toml, ignoring unknown
         configuration key: profile.ci.slow-warning
```

So `profile.ci` had **no per-test timeout at all**. From merge_group job `95162862834` for #2502:

```
12:31:55  build starts
12:50:58  Starting 80806 tests across 69 binaries
13:36:07  ##[error]The operation was canceled.
```

45 minutes inside nextest, killed by the **job's** `timeout-minutes: 85`, naming no test. **#2502 was evicted from the merge queue 31 seconds later.**

That warning was **line 353 of that job's log** — and of every workspace-test log before it.

## The compounding half

`status-level = "fail"` is *below* `slow` in nextest's ordering (`none < fail < retry < slow < pass < all`), so the SLOW lines that would have named the culprit were suppressed too. Measured on a probe crate:

| status-level | SLOW lines |
|---|---|
| `fail` (current) | **0** |
| `slow` | 3 |

Fixing the timeout without this would have kept hiding the warning that precedes the kill.

## terminate-after is measured, not guessed

Full workspace run on an idle 48-core box — 80,806 tests, **353s wall**:

| test | duration |
|---|---|
| `aprender-orchestrate bug_hunter::test_bh_mod_001_hunt_all_modes` | **202.7s** |
| `aprender-orchestrate bug_hunter::test_bh_mod_001_hunt_returns_result` | 157.3s |
| `aprender-orchestrate bug_hunter::test_bh_mod_046_apply_spec_quality_gate_no_pmat` | 118.0s |
| `aprender-train transformer_trainer::falsify_lora_tests::…rslora_stable_high_rank` | 67.5s |

over 60s: **6** · over 120s: **2** · over 300s: **0**

`terminate-after = 20` periods = 1200s is **~6× the slowest real test**, so it cannot kill a legitimately slow one — while a genuine hang dies at 20 minutes *with a name* instead of taking the whole job down anonymously.

(The `bug_hunter` tests are separately being made fast; this bound does not depend on that landing.)

## Also fixed

- **`timeout-minutes: 85` → 100.** The step sets 75, but "Set up runner" measured **20 minutes**, so 20 + 75 = 95 > 85 and the job timeout always fired first — producing a bare "The operation was canceled" naming no step. The step timeout must be the one that *can* fire, because it points at the step.
- **junit path was doubled.** `path` is relative to the store dir, not the workspace root, so `target/nextest/ci/junit.xml` produced `target/nextest/ci/target/nextest/ci/junit.xml`. Now `junit.xml`, verified landing at `target/nextest/ci/junit.xml`.

## Guarded

A warning nobody reads is not a diagnostic, so it is a hard failure now. `check_nextest_config_keys.sh` runs nextest's **real parser** against a three-line throwaway crate with this repo's config — no regex of my own guessing at nextest's schema — and fails on any ignored key. Its case table carries a control row so it cannot pass by reporting every key it sees.

**Mutation**: restoring `slow-warning` turns it RED; restoring `slow-timeout` turns it green. Verified both directions.

Wired into `guard-runner-labels`, which is in `gate.needs`. `bashrs lint`: 0 errors.

Refs #2502

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,15 +1,41 @@
 # cargo-nextest CI profile
-# Deployed from paiml/infra — do not edit manually.
+# Deployed from paiml/infra — CHANGED HERE deliberately; fold back upstream.
 # Spec: docs/specifications/unified-ci-pipeline.md
-
+#
+# `slow-warning` was not a nextest key. nextest does not reject an unknown key,
+# it prints `ignoring unknown configuration key: profile.ci.slow-warning` and
+# carries on -- so this profile had NO per-test timeout at all, and the warning
+# sat unread in every workspace-test log. Consequence, merge_group job
+# 95162862834 for #2502:
+#
+#     12:50:58  Starting 80806 tests across 69 binaries
+#     13:36:07  ##[error]The operation was canceled.
+#
+# 45 minutes inside nextest, killed by the JOB's `timeout-minutes: 85`, naming
+# no test; the queue entry was evicted 31 seconds later. `check_nextest_config_keys.sh`
+# now fails on any key nextest ignores.
 [profile.ci]
 retries = 2
 fail-fast = true
-slow-warning = "60s"
-status-level = "fail"
+# Measured on an idle 48-core box, 80806 tests, 353s wall: slowest single test
+# is 202.7s (aprender-orchestrate bug_hunter::test_bh_mod_001_hunt_all_modes),
+# only 2 tests exceed 120s and NONE exceeds 300s. terminate-after = 20 periods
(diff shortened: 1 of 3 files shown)


noahgift:

Issue #2467, opened 2026-08-14
aprender-graph pulls 48 crates of arrow/parquet by default (119 -> 71 with --no-default-features)


noahgift:
## Measured

`aprender-graph` declares `arrow` and `parquet` as `optional = true`, then turns them on for everyone:

```toml
[features]
default = ["storage"]
storage = ["dep:arrow", "dep:parquet"]
```

Optional-but-default-on is effectively mandatory. Dependency counts, measured with `cargo tree -p aprender-graph --edges normal`:

| build | crates |
|---|---:|
| default | **119** |
| `--no-default-features` | **71** |
| **delta** | **48 (40%)** |

All 48 are the arrow/parquet chain and its transitive tail: `arrow`, `arrow-arith`, `arrow-array`, `arrow-buffer`, `arrow-cast`, `arrow-data`, `arrow-ipc`, `arrow-ord`, `arrow-row`, `arrow-schema`, `arrow-select`, `arrow-string`, `parquet`, plus `ahash`, `aho-corasick`, `brotli`, `brotli-decompressor`, `base64`, `atoi`, `alloc-no-stdlib`, `alloc-stdlib` …

Clean-build wall time for both configurations is being measured and will be added; the dependency delta above is exact.

## What this is NOT

The code is fine. `src/storage/parquet.rs` genuinely uses `RecordBatch`, `ArrowWriter` and `ParquetRecordBatchReaderBuilder`, correctly gated behind `#[cfg(feature = "storage")]`. This is not an unused dependency — it is a **default-feature choice**.

Two related claims that do **not** hold, checked because a wrong premise makes a bad ticket:

- `aprender-graph` does **not** depend on `aprender-db` / `trueno-db`. The only match in its manifest is a comment (`# Workspace lints (PMAT best practices from trueno-db/trueno)`). The sole declarer of `aprender-db` is `aprender-db` itself. Note `trueno_db` is the **lib** name of the `aprender-db` package, which makes the two easy to conflate.
- `aprender-compute` does **not** depend on `aprender-graph`. Its manifest match is a comment recording that the dep was *removed* under APR-MONO self-containment.

## Blast radius is smaller than it looks — in-tree

`aprender-viz`, the only workspace consumer, already does the right thing:

```toml
trueno-graph = { … package = "aprender-graph", default-features = false, optional = true }
```

[...]

PR #2518, opened 2026-08-16
fix(release): Cargo.lock was 15 days stale — the binary-release workflow could not build

noahgift:
On a clean checkout of `main`, `cargo metadata` **alone** — no build, no test — rewrites the lockfile:

```
1 file changed, 87 insertions(+), 1436 deletions(-)
```

Reproduced identically in **three independent worktrees**, so it's the tree, not one machine.

`Cargo.lock` was last committed **2026-08-01** (0.63.0). Manifests changed **08-10** and **08-11** without it.

## Why nothing caught it

Every CI job runs cargo **without** `--locked`. cargo then updates the lock in place and carries on green, so a stale lock is invisible on every PR.

The only jobs that pass `--locked` are the ones that never run on a PR:

```
.github/workflows/binary-release.yml:116  cross build ... --locked
.github/workflows/binary-release.yml:118  cargo build ... --locked
```

That is the **release path**. Running its exact command on `main`:

```
$ cargo build --release --bin pv -p aprender-contracts-cli --locked
error: cannot update the lock file ... because --locked was passed to prevent this
```

So the check that mattered only ran at the moment it was most expensive to fail — the same shape as the rest of this class: **the guard did not scan the surface where the decision is made.**

## Fix

Regenerated `Cargo.lock` from the current manifests. `cargo metadata --locked` and the binary-release resolve path both exit 0 afterwards.

The 1,436 removed entries are dominated by arrow/parquet, consistent with the aprender-graph default-features work (#2467) landing without a lock update.

## Guarded

`check_lockfile_current.sh` runs `cargo metadata --locked` — resolution is the whole question here, codegen is not, so it needs no build, no network, and takes about a second.

Its case table has a control row that is the actual point: **row 2 adds a dependency to a probe manifest and asserts `--locked` refuses**, so the guard cannot pass by never rejecting anything.

**Mutation**: restoring `main`'s stale `Cargo.lock` turns the guard RED; regenerating turns it green. Verified both directions.

Wired into `guard-runner-labels` (in `gate.needs`). `bashrs lint`: 0 errors.

<sub>Found while investigating why agents' worktrees all showed a dirty `Cargo.lock` — it wasn't them.</sub>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -426,6 +426,15 @@ jobs:
       # correct everywhere else). Run the table; do not re-read the regex.
       - name: "`apr`-pinning guard must still turn RED (case table + surfaces)"
         run: bash scripts/check_apr_bin_pinned.sh --self-test
+      # Poka-yoke: Cargo.lock was 15 days stale on main (last touched 2026-08-01,
+      # manifests changed 08-10 and 08-11). Every CI job runs cargo WITHOUT
+      # --locked, so cargo rewrote the lock in place and stayed green; the only
+      # jobs that pass --locked are in binary-release.yml, i.e. the RELEASE path,
+      # which was broken on main. Resolution only, no build, ~1s.
+      - name: Cargo.lock must match the manifests
+        run: bash scripts/check_lockfile_current.sh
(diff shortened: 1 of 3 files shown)


noahgift:

Issue #2503, opened 2026-08-15
SURF tracker: 45 of 150 command paths are ungated, and 3 of 5 CLI falsification conditions name a test that does not exist


noahgift:
Tracker for the `SURF` (binary surface) work. Spec: `docs/specifications/aprender-binary-surface-design.md` (branch `claude/binary-surface-design-klhvql`, commit `deff1ca8`).

Phase A of the spec's own refinement protocol has been executed against `907bef27`. **It falsified four of the ten premises the spec was built on**, which is the intended outcome of that protocol and is why this tracker leads with the corrections rather than the plan.

## What was wrong with the premise

The draft opened: *"Aprender advertises 103 CLI commands … the shipped binary has never been asked."* Both halves are false.

- The number is **105** (`README.md`, and the binary agrees).
- The binary **is** asked. `crates/apr-cli/tests/cli_commands.rs:147` spawns `env!("CARGO_BIN_EXE_apr")` — a compiled artifact — parses `--help`, and asserts the command **set** bidirectionally against `contracts/apr-cli-commands-v1.yaml`. It runs at `.github/workflows/ci.yml:327` inside `workspace-test`, which is in `gate`'s `needs` (`ci.yml:540`). It blocks merges today.
- A second guard landed the same week (#2485): `ci.yml:516` runs `scripts/check_readme_claims.sh` in `guard-runner-labels`, also in `gate.needs`. The draft asserted that script ran nowhere.

So the class the spec was written to fix is, at depth 1, already fixed. **Do not build a surface emitter to solve it.**

## What is actually broken

The binary was built from HEAD (`apr 0.63.0 (907bef27)`) and measured rather than inferred:

| Measurement | Result |
|---|---|
| Top-level commands | **105** — artifact, README, contract registry and Rust mirror all agree |
| Parents with subcommands | **15** (`serve, debug, canary, registry, runs, experiment, probar, modelfile, train, tokenize, data, pipeline, dataset, kernel, rosetta`) |
| Depth-2 subcommands | **45** — `train`(8), `rosetta`(8), `tokenize`(6), `data`(5), `pipeline`(4), `runs`(3), `serve`(2), `canary`(2), 7 parents with 1 each |
| Depth-3 | **0** — the tree is exactly 2 deep |
| **Total invocable paths** | **150, of which 105 are gated and 45 are not** |

`registered_commands()` holds 105 entries and **zero contain a space**. `apr registry aliases` can be renamed or deleted and every surface gate stays green.

[...]

PR #2523, opened 2026-08-16
fix(security): deny.toml ran in no workflow and had drifted until it failed on main

noahgift:
`cargo deny check` on `origin/main`:

```
advisories ok, bans FAILED, licenses FAILED, sources ok      (rc=6)
```

## Why nothing noticed

`cargo-deny` is invoked from exactly one place in this repo: the `deny` target in the Makefile. **That target is not a prerequisite of tier1, tier2, tier3, or tier4** — so nothing but a human typing `make deny` ever ran it.

Meanwhile **cargo-deny 0.20.2 is already installed in the CI image** (`/usr/local/cargo/bin/cargo-deny`). The tool shipped, the 7.2 KB policy shipped, and no workflow connected them.

`ci / security` from the reusable sovereign-ci workflow runs `cargo audit`, which **does not read `deny.toml`** — so the licence allow-list, banned-crate rules, and source allow-list were enforced by nothing at all.

CLAUDE.md compounds this: its CI/CD section documents `security.yml` ("cargo-audit, cargo-deny, cargo-outdated (weekly)"), `benchmark.yml`, and `release.yml`. **None of those three files exist.**

## Both failures were config drift, not defects

**1 — bans/wildcards.** All 10 flagged deps are `kind = dev` sibling path-deps with no `version`:

```
aprender-core     -> aprender-train, aprender-profile
aprender-compute  -> aprender-core, aprender-simulate, aprender-cuda-edge,
                     aprender-solve, aprender-sparse
aprender-serve, aprender-orchestrate -> aprender-cuda-edge
aprender-simulate -> aprender-present-test
```

That is the deliberate mechanism for breaking the crates.io publish cycle — `cargo publish` strips dev-dependencies, so a path-only dev-dep can't create a circular version requirement. **Adding `version` would reintroduce the cycle.** cargo-deny reports a versionless path dep as a wildcard, so it flagged ten load-bearing choices as banned.

`allow-wildcard-paths = true`. `wildcards = "deny"` still applies to real dependencies, which is the case worth banning.

**2 — licences.** `libfuzzer-sys` is `(MIT OR Apache-2.0) AND NCSA`, reached via:

```
libfuzzer-sys <- rav1e <- ravif <- image <- aprender-test-lib
```

i.e. through an AV1 encoder pulled in by `image`, not through anything this project chose. NCSA is OSI-approved and FSF Free/Libre.

Added as a **scoped `[[licenses.exceptions]]`** for `libfuzzer-sys` rather than putting NCSA in the global `allow` list — so a future NCSA dependency arriving by a different path is still flagged for a decision instead of being silently admitted. Flagging that as the one judgement call in this PR.

**After: `advisories ok, bans ok, licenses ok, sources ok`**

## Wired so it can't drift again

- `ci.yml` `guard-runner-labels` (in `gate.needs`) runs `cargo deny check` inside the image, where cargo-deny already lives — no install step needed.
- `make tier3` runs it too, so pre-push catches it locally.

**Mutation:** reverting `deny.toml` reproduces `bans FAILED, licenses FAILED`; restoring gives four greens. Verified both directions.

`bashrs make lint`: 1 error, identical to `origin/main` (pre-existing).

## Release-readiness, checked alongside

`cargo publish --dry-run --no-verify` across **all 70 publishable crates: 70 ok, 0 FAIL** — against the corrected lockfile from #2518. It fails outright without that fix.

Refs #2503

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -426,6 +426,27 @@ jobs:
       # correct everywhere else). Run the table; do not re-read the regex.
       - name: "`apr`-pinning guard must still turn RED (case table + surfaces)"
         run: bash scripts/check_apr_bin_pinned.sh --self-test
+      # deny.toml is 7.2 KB of licence / banned-crate / source-allowlist policy
+      # that ran in NO workflow -- only `make deny`, which is not a prerequisite
+      # of any tier, so nothing but a human typing it ever invoked it. Meanwhile
+      # cargo-deny 0.20.2 ships INSIDE the CI image already.
+      #
+      # Because nothing ran it, the policy drifted out of sync with the repo's
+      # own deliberate architecture and now fails on main:
+      #     advisories ok, bans FAILED, licenses FAILED, sources ok
+      # Both failures were config drift, not defects (see the deny.toml comments).
+      # Fixed there; wired here so it cannot drift unnoticed again.
+      #
+      # `ci / security` from the reusable workflow runs `cargo audit`, which does
+      # NOT read deny.toml -- so this covers licences, bans and sources, which
+      # nothing else does.
+      - name: cargo-deny (licences, bans, sources, advisories)
+        run: |
+          docker run --rm \
+            -v "${GITHUB_WORKSPACE}:/workspace" \
+            -w /workspace \
+            "$IMAGE" \
+            cargo deny check
       # Poka-yoke: `set` in a SOURCED file mutates the caller's shell. apr_bin.sh
       # opened with `set -euo pipefail`; qwen-story.sh sources it and had chosen
       # `set -uo pipefail` deliberately (it must run every beat and tally the
--- a/Makefile
+++ b/Makefile
@@ -214,6 +214,8 @@ tier3:
 	@bash scripts/check_include_files.sh
 	@echo "Checking publish safety (symlinks, companion lookups)..."
 	@bash scripts/check_publish_safety.sh
+	@echo "Checking cargo-deny policy (licences, bans, sources, advisories)..."
+	@$(MAKE) --no-print-directory deny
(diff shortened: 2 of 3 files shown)


noahgift:
**This PR is the blocker for every merge on the repo.** Traced while diagnosing why #2502 had been evicted from the merge queue four times:

```
merge_group run 31961990382
  ci / lint   FAILED  ->  step "Supply chain audit (cargo deny)"
  ci / gate   FAILED  ->  #2502 evicted
```

The reusable sovereign-ci workflow runs `cargo deny check advisories licenses sources` inside `ci / lint`. On `origin/main`:

```
advisories ok, licenses FAILED, sources ok
```

with this PR:

```
advisories ok, licenses ok, sources ok
```

So `main` has been unmergeable all day on the licence rejection, and nothing else can land until this does.

## Two corrections to the PR description above

1. It says cargo-deny "ran in no workflow". The **file** `security.yml` genuinely doesn't exist, but the reusable workflow does run it inside `ci / lint`. The description overstates the gap — licences *were* gated, and were failing.

2. CI checks `advisories licenses sources` — **not `bans`**. So the `allow-wildcard-paths` half of this PR is real but not what unblocks anything. **The scoped NCSA exception is the load-bearing change**, which makes the judgement call flagged in the description the one that actually matters. Worth your explicit sign-off rather than my inference.

## Known overlap, deliberately not fixed here

The step this PR adds to `guard-runner-labels` runs `cargo deny check` (all four gates), while the reusable workflow already covers three. The genuinely new coverage is **`bans`**; the other three are duplicated (~10s).

Not fixing that here on purpose: `concurrency.cancel-in-progress: true` means a push would cancel this PR's checks, and right now this run is the only thing on the fleet and the whole repo is waiting on it. Narrowing the step to `cargo deny check bans` belongs in a follow-up.

PR #2540, opened 2026-08-19
fix(release): apr-cli could not be published — it depended on a publish=false crate

### LINKED ISSUE #2539: RELEASE BLOCKER: apr-cli cannot be published — it depends on aprender-qa-cli, which is publish=false
Found by running pre-release **Gate 11 (crates.io cascade publishability)** against the batch-B tree *before* cutting 0.64.0. This is the failure shape that stopped the v0.50.0 cascade at 29/68 crates.

## The blocker

```
$ cargo publish -p apr-cli --dry-run --allow-dirty --no-verify
error: failed to prepare local package for uploading

Caused by:
  no matching package named `aprender-qa-cli` found
  location searched: crates.io index
  required by package `apr-cli v0.63.0`
```

`apr-cli` is publishable (`publish` unset) and declares:

```toml
# crates/apr-cli/Cargo.toml
aprender-qa-cli = { workspace = true }        # -> version = "0.63.0"
```

but `aprender-qa-cli` is marked **`publish = false`**. A publishable crate cannot depend on an unpublishable one: cargo requires every non-dev dependency to resolve against the registry.

Introduced by #2493 (`apr qa-playbook` routing) — `git show origin/main:crates/apr-cli/Cargo.toml | grep aprender-qa-cli` returns nothing, so `main` before the batch does not have it.

## Scope — exactly one blocker

Enumerated across all 78 workspace members:

```
unpublishable (publish = false), 8:
  aprender-bench-compute, aprender-bench-tokenizer, aprender-compute-xtask,
  aprender-qa-certify, aprender-qa-cli, aprender-qa-gen,
  aprender-qa-report, aprender-qa-runner

PUBLISHABLE crates depending on an UNPUBLISHABLE one:
  apr-cli  ->  aprender-qa-cli  (req = ^0.63.0)      <- the only one
```

The other seven are internal tooling nothing publishable references. So this is a single, well-scoped problem.

## `optional = true` does NOT fix it — tested

```toml
aprender-qa-cli = { workspace = true, optional = true }
```
```
rc=101  error: failed to prepare local package for uploading
Caused by: no matching package named `aprender-qa-cli` found
```

cargo validates optional dependencies at publish time too, so feature-gating alone cannot unblock this.

## Options, with tradeoffs

**A. Make the QA crates publishable.** Requires the whole closure, not just one crate:

```
aprender-qa-cli, aprender-qa-certify, aprender-qa-gen,
aprender-qa-report, aprender-qa-runner        (5 crates)
```

Publishes internal QA tooling to crates.io permanently — **crates.io publishes are irreversible**, and these were deliberately marked `publish = false`. Also needs them added to `scripts/cascade-publish.sh` tiers ahead of `apr-cli` (tier 10).

**B. Drop the dependency from apr-cli.** `apr qa-playbook` would not be routed in the published binary. Smallest change, but removes a command #2493 added and reintroduces the "advertised but unreachable" risk if `--help` still lists it — see the `apr test llm` case (#2527), where a command was advertised while its remedy was impossible.

**C. Restructure** so `apr-cli` does not need the crate — move the dispatch code into `apr-cli`, or invoke the tool as a subprocess. Most work; keeps both the command and the publish boundary intact.

## Recommendation

**C, falling back to B if time-boxed.** The QA crates look like genuine internal tooling — `certify`, `gen`, `runner`, `report` are build/CI-facing, not library surface — and publishing five of them to crates.io forever to route one subcommand is a large, irreversible price for a small feature. B is acceptable *only* if `qa-playbook` is also removed from `--help`, so the surface does not advertise something the published binary cannot do.

Not deciding this unilaterally: option A is a one-way door, and B changes shipped CLI surface.

## Note

This does **not** block #2537 from merging — it blocks the **0.64.0 release cut**. Everything else in the pre-release protocol passes on the batch tree:

```
Gate 1  package integrity   PASS   1767 include files tracked; 1539 targets survive cargo package
Gate 3  stale cfg audit     PASS   clippy -D warnings rc=0
Gate 4  MSRV                PASS   rust-version 1.91 consistent, toolchain 1.93.0
Gate 11 flagship dry-run    PASS   aprender: Packaged 631 files, 3.0MiB
        aprender-core       PASS   Packaged 1680 files, 16.5MiB
        aprender-serve      PASS   Packaged 2293 files, 55.7MiB
        apr-cli             FAIL   <- this issue
```

noahgift:
Closes #2539.

Found by running pre-release **Gate 11** against the batch tree *before* cutting 0.64.0 — which is the point of running gates early:

```
$ cargo publish -p apr-cli --dry-run
error: failed to prepare local package for uploading
Caused by: no matching package named `aprender-qa-cli` found
```

> **Note:** this branch is cut from `batch/consolidated-b` (#2537), since that is where the defect is introduced. Merge after #2537.

## Five whys

1. `apr-cli` cannot publish → it depends on `aprender-qa-cli`.
2. Why? → #2493 routed `apr qa-playbook` into it.
3. Why did that pattern work for the other five siblings it routed? → `aprender-cgp`, `-rag-cli`, `-zram-cli`, `-simulate`, `-contracts-cli` are **all publishable**.
4. Why is this one different? → it carries `publish = false  # Internal QA harness`, and has since ≥0.61.0.
5. **Why did a route into an internal-only crate get added at all?** → **nothing checked that a publishable crate depends only on publishable crates.** Root cause.

## The fix respects the existing boundary

The `publish = false` is deliberate, documented, and three releases old; the closure is build/CI tooling (`certify`, `gen`, `runner`, `report`). Making `apr-cli` publishable the other way would mean publishing **five internal crates to crates.io permanently** — an irreversible one-way door — to route one subcommand. So the defect is #2493's route, not the boundary.

`apr qa` is untouched — it lives in `apr-cli` and never depended on this crate.

Removed `apr qa-playbook` from **every** surface, so nothing advertises what the published binary cannot do (the `apr test llm` lesson — a command advertised in `--help` whose own remedy was impossible):

| surface | caught by |
|---|---|
| `commands_enum.rs` variant | planned |
| `dispatch.rs` arm | planned |
| `crates/apr-cli/Cargo.toml` dep | planned |
| `contracts/apr-cli-commands-v1.yaml` | planned |
| `tests/cli_commands.rs::registered_commands` | **FALSIFY-CLI-001** |
| depth-2 vacuity floors | **FALSIFY-CLI-005/006** |
| book page + PCU contract + SUMMARY link | planned |

The three-surface drift guard caught the surface I forgot. That machinery is why removing a command is safe here despite touching seven places.

### On the vacuity floors

Real depth-2 count went 128 → 113, below the `>= 120` floor. Lowered to **105**, deliberately *not* to just-under-113: a floor pinned tight to the current surface re-breaks on every ordinary command change and trains people to bump it reflexively, hollowing out the parser-health check it exists to be. The reasoning is in the comment.

## Poka-yoke — the half that matters more

`scripts/check_publishable_deps_publishable.sh`, wired into `ci.yml` with its case table.

This class is **invisible to every normal gate** — `build`, `test`, and `clippy` all resolve the path dependency locally and pass. It surfaces only at `cargo publish`, i.e. mid-release-cascade, the worst possible time. The v0.50.0 cascade died exactly this way at 29 of 68 crates.

| check | result |
|---|---|
| self-test | **3/3** — includes the control (all-publishable must NOT be flagged) and "unpublishable → unpublishable is fine", since those are never uploaded |
| clean tree | PASS, 78 members scanned |
| **mutation** | re-add the real dep → **RED**, naming `apr-cli -> aprender-qa-cli` |
| vacuity floor | fails if the enumeration sees < 15 members |

It also records a **tested dead end**: `optional = true` does *not* launder an unpublishable dependency — cargo validates optional deps at publish time too. Verified empirically, so nobody rediscovers it the expensive way.

## Verification

```
cargo publish -p apr-cli --dry-run          rc=0   Packaged 606 files, 9.1MiB
cargo clippy --all-targets -D warnings      rc=0
cargo fmt --check                           rc=0
cargo test -p apr-cli --test cli_commands   12 passed, 0 failed
pv validate contracts/apr-cli-commands-v1.yaml   Contract is valid.
check_publishable_deps_publishable.sh       PASS
check_guards_are_wired.sh                   unwired 2 -> 1 (this guard is live)
```


noahgift:
**Housekeeping — this PR's diff is misleading and I will rebuild it.**

The branch was cut from `batch/consolidated-b` (#2537), because the defect it fixes is introduced there. That means GitHub shows it as **48 commits / 623 files / −7.8M lines** — almost all of which is Batch B, not this change.

The actual fix is **2 commits, 599 patch lines**:

```
fix(release): apr-cli could not be published — it depended on a publish=false crate
docs(readme): counts follow the qa-playbook removal — 1778->1777 contracts, 111->110 commands, 113->112 chapters
```

Since #2537 will be **squash**-merged, its 48 original commits won't be recognised as already-merged, so rebasing this branch afterwards would conflict rather than reduce to the delta.

Plan: once #2537 lands, I'll rebuild this as a minimal PR off the new `main` containing only the qa-playbook removal (7 surfaces), the new `check_publishable_deps_publishable.sh` guard plus its ci.yml wiring, and the README counts. The verification already recorded here stands — it was run on the merged tree:

```
cargo publish -p apr-cli --dry-run          rc=0  Packaged 606 files, 9.1MiB
cargo clippy --all-targets -D warnings      rc=0
cargo test -p apr-cli --test cli_commands   12 passed, 0 failed
check_publishable_deps_publishable.sh       PASS (mutation → RED on the real defect)
```

noahgift:
**Rebuilt off the new `main` (`542102499`).**

#2537 landed batch B, so this branch's old base was squashed away. As flagged earlier, rebasing would have conflicted rather than reduced to the delta — the 48 original commits aren't recognisable in a squash merge. Rebuilt via `git format-patch` → `git am --3way` onto the new main instead.

| | before | after |
|---|---|---|
| commits | 48 | **3** |
| files | 623 | **15** |
| lines | +9334 / −7,821,230 | **+170 / −209** |

That is now the actual change, and reviewable as such.

Re-verified on the rebuilt branch against the new main:

```
cargo publish -p apr-cli --dry-run   rc=0   Packaged 606 files, 9.1MiB
check_publishable_deps_publishable   PASS
check_readme_claims                  PASS (78 crates / 1777 contracts / 110 commands)
```

The third commit is new since the original PR: `beat_apr_sibling_cli_reach.rs` still asserted `apr qa-playbook` was routed, which was the **sixth** surface this one removal touched. Caught by `workspace-test`, verified as a real failure (0 dep-info write errors in that job) rather than the transient ENOSPC class hitting the fleet today — see paiml/infra#221.

Issue #2546, opened 2026-08-20
Publish provable-contracts{,-cli,-macros} as facade crates so existing users are carried forward, not stranded


noahgift:
The `provable-contracts` family was renamed to `aprender-contracts*` during the APR-MONO consolidation. The old crates are still on crates.io, frozen at **0.3.1**, while the new ones are at **0.63.0** — and nothing tells anyone.

## How this surfaced

paiml/infra pinned `provable-contracts-cli = "0.3.1"` for the fleet's `pv`. That pin **resolved cleanly and installed a tool sixty versions behind**. No error, no warning. It only came to light because 0.3.1 predates the `safety`/`liveness` proof-obligation types, so current contracts failed there with an unknown-variant error — the guard read as *broken* rather than *out of date*.

Anyone else pinning or installing these crates has the same silent trap.

## The good news: these are drop-in facades

I diffed the public surfaces:

| crate | old | new | verdict |
|---|---|---|---|
| `provable-contracts` | 37 public items | 37 public items | **byte-identical export list** |
| `provable-contracts-macros` | 5 `#[proc_macro*]` | 5 `#[proc_macro*]` | identical surface |
| `provable-contracts-cli` | bin `pv` | bin `pv` | bin-only both sides — see below |

So the library facade is genuinely a drop-in, not a hopeful one.

## Proposed: least disruptive, and the idiomatic Rust pattern

crates.io has **no rename mechanism** ([rust-lang/crates.io#2902](https://github.com/rust-lang/crates.io/issues/2902)); facade crates are the established convention. This repo already publishes non-`aprender*` names (`apr-cli`, `apr-format`), so there's precedent for hosting them here.

**Publish from aprender, not from the old repo.** `paiml/provable-contracts` is **archived** (2026-04-06) — read-only, no PRs, CI disabled — so the shim cannot be developed there without unarchiving. Hosting the facades next to the crates they re-export also means a shape change in `aprender-contracts` breaks the facade in the *same CI run*, rather than rotting silently. Which is exactly the failure mode that caused this.

### 1. `provable-contracts` → facade

```toml
[dependencies]
aprender-contracts = "0.63"
```
```rust
pub use aprender_contracts::*;
```

Existing users get a working crate on upgrade with **zero code changes**.

### 2. `provable-contracts-macros` → facade, with a caveat worth checking

A `proc-macro = true` crate may only export proc macros defined in itself, so the facade likely has to be a **normal crate** that re-exports:

```rust
[...]

PR #2553, opened 2026-08-20
fix/pv facades

noahgift:
- **feat(crates.io): carry provable-contracts{,-macros,-cli} forward as facades (#2546)**
- **fix(facades): the CLI facade could not compile when published, and the guard could not see it**

Implements #2546. Tracked by #2556.


noahgift:
Blocked on **#2558**: this branch's `provable-contracts-cli` declares `[[bin]] name = "pv"`, and `aprender-contracts-cli` on main already does. Publishing both means two crates installing the same binary to `~/.cargo/bin/pv`, with the second `cargo install` silently overwriting the first.

That compounds the STAGE verdict already recorded here: this facade is not publishable in the 0.63.0 cascade because it calls `aprender_contracts_cli::run()` against a bin-only published crate. Resolve **both** before it ships.

noahgift:
**#2558 updated with measured crates.io data — read before the 0.64.0 cascade.**

There is already a crate named `pv` on crates.io: *"Rust reimplementation of the unix pipeview (pv) utility"*, `bin_names: ["pv"]`, **7,065 downloads**, latest published **2025-07-19**, first published **2019**. So the naked name has four claimants — that crate, `/usr/bin/pv` from every distro, `aprender-contracts-cli`, and this facade — all writing `~/.cargo/bin/pv`, and `cargo install` overwrites without warning.

The download split reframes what the facades are actually for:

| crate | downloads |
|---|---|
| `provable-contracts-macros` | **46,571** |
| `provable-contracts` | **10,809** |
| `provable-contracts-cli` | **463** |

The library and macros are the population worth carrying forward, and neither involves a binary. The CLI is 463 downloads and is the *only* one that collides on a name — which makes "ship the facade lib-only" a cheap resolution rather than a sacrifice.

noahgift:
Updated in place with `41be1d870`. The STAGE verdict this PR carried is **gone**, and dissolved rather than waited out.

The facade's `src/main.rs` called `aprender_contracts_cli::run()` against a published 0.63.0 that is bin-only (`has_lib: false` on the crates.io API). Verified in both directions with `cargo publish --dry-run`, which resolves from the **registry** rather than the path dep — the build a real consumer gets, and the one a path-dep `cargo check` cannot see:

- old (bin) shape: `rc=101`, `error[E0433]: cannot find module or crate \`upstream\``, `failed to verify package tarball`
- new (lib-only) shape: `rc=0`, packaged and verified

`provable-contracts-cli` is now lib-only with **zero dependencies**, so it is publishable at any point in the 0.64.0 cascade, and it stops being the fourth claimant on `~/.cargo/bin/pv`. The install failure is signposted in four places (description, README, build.rs warning, `#[deprecated]`) and `check_facade_compat.sh` row C3 asserts the routes it names actually exist.

`check_facade_compat.sh`'s CURRENCY block was **replaced, not deleted** (C1/C2/C3 + R7 + an inverted R3), and `scripts/check_duplicate_bin_names.sh` is new — scanning both workspaces, with the intentional `apr` duplicate handled by declared intent, mutation-verified in four directions with each mutation proven to engage. Full detail in paiml/aprender#2558 (now closed); item 3 carried forward as #2559.

PR #2554, opened 2026-08-20
fix/pv schema

noahgift:
- **fix(contracts): three silent-drop defects in the contract schema, and the test target that would have caught them was dark**
- **fix(contracts): the guard's own fixture was one of the 119 offenders**

Refs #2504. Tracked by #2556.


noahgift:

abandoned.

Issue #2572, opened 2026-08-22
block v0.1.6 will be rejected by a future rustc and sits under wgpu->metal, the only macOS GPU backend


noahgift:
`block v0.1.6` is on rustc's future-rejection list and sits under the **only GPU backend
macOS has**. Found by the arm64 macOS dogfood (#2566).

```
warning: the following packages contain code that will be rejected by a future
version of Rust: block v0.1.6
```

Dependency chain, proven with `cargo tree -i block --target aarch64-apple-darwin`:

```
block v0.1.6 <- metal v0.32.0 <- wgpu-hal v27.0.4 <- wgpu v27.0.1
              <- aprender-compute v0.63.0 <- apr-cli/aprender-core <- aprender v0.63.0
```

## Isolation proven, not assumed

The same query against `--target x86_64-unknown-linux-gnu` returns *"nothing to print"* —
`block` is entirely **absent** from the Linux dependency graph. Neither Linux sweep
(x86_64+4090, aarch64+GB10) can see this. It arrives through wgpu's Apple backend,
alongside `wgpu-core-deps-apple v27.0.0` and `metal v0.32.0`.

## Why it matters beyond a warning

`block` is an unmaintained Objective-C block-runtime shim. When rustc follows through on
the rejection, **the macOS GPU build breaks specifically** — and Metal is the only GPU
path macOS has, so there is no fallback.

## Work

Audit the `wgpu -> metal -> block` edge and pin or upgrade before it becomes a hard error.
Check whether a newer `metal` crate drops the `block` dependency.

Note: the specific lint is UNMEASURED — cargo's `--future-incompat-report` data lives in
the temp install dir and is deleted post-install. The warning text is quoted verbatim.

Refs #2566

noahgift:
0.67 lane — attaches to PP-066 spec §6 row **B-M1..M4**.

`block v0.1.6` (future-rejected by rustc) sits under wgpu → metal, the path §6 B-M1 opens first ('wgpu-Metal adapter first') on mini in 0.67; 0.66 ships no Metal lane (D-3 escalated, §8 'no lane on intel or mini before D-3'). Pin/upgrade the metal crate with B-M1.

_PP-066 triage 2026-09-05 (issue #2873); ledger `docs/audits/pp-066-triage-2026-09-05.md`._

noahgift:
[...]

Issue #2568, opened 2026-08-22
OOM guard fails OPEN on macOS: /proc/meminfo + unwrap_or(u64::MAX) makes the threshold 12.8 exabytes


noahgift:
The guard that refuses a dequant exceeding 80% of RAM **can never fire on macOS**. Found by
the arm64 macOS dogfood before the 0.64.0 cut (#2566).

`crates/aprender-serve/src/contract_gate.rs:464` — `system_memory_bytes()` reads
`/proc/meminfo` and nothing else. Line 427 does `.unwrap_or(u64::MAX)`.

macOS has no `/proc`, so the threshold becomes **18446744073709551615 bytes (~12.8
exabytes)** and `estimated_peak > threshold` is unsatisfiable. The guard is inert.

Confirmed against the shipped binary: `strings` finds `/proc/meminfo` present and
`hw.memsize` absent. No `sysctl` fallback exists anywhere in `aprender-serve`.

## It is macOS, not ISA — and that is provable

The aarch64 Linux box (GB10) has `/proc`, so the **same ISA** behaves correctly there. The
two arm64 platforms disagree, which isolates the cause to the OS.

## The aggravating part

The only test **self-skips on the platform where it is broken**:

```rust
if cfg!(target_os = "linux") { ... }
```

So this branch has never been exercised on macOS. That is the same class this repo keeps
finding — a control that reports success and discriminates nothing — scoped to a platform
instead of a code path.

## Work

1. `sysctl hw.memsize` fallback on macOS (and a documented behaviour for any other OS).
2. **Never `.unwrap_or(u64::MAX)` on a safety threshold.** Unknown memory must fail
   CLOSED, not open. If the total cannot be determined, refuse the dequant and say why.
3. Delete the `cfg!(target_os = "linux")` skip and make the test run everywhere — a test
   that skips the broken platform is worse than no test, because it reads as coverage.

Runtime OOM is INFERRED, not observed (reproducing needs a multi-GB model, excluded by the
fleet power ceiling). **The code path is MEASURED.**

Refs #2566

Issue #2567, opened 2026-08-22
Q4_K GEMV has no aarch64 SIMD and its "parallel" variant calls the scalar path


noahgift:
The hottest kernel in quantized inference is **scalar and single-threaded on every ARM
machine**, silently. Found by the aarch64 dogfood before the 0.64.0 cut (#2566).

`crates/aprender-compute/src/backends/q4k/gemv/mod.rs:53` — `#[cfg(target_arch = "x86_64")]`
wraps the **entire** fast path: the `>=8M`-work parallel dispatch (exactly what catches
FFN 8960x1536 and lm_head 151936x1536), AVX-512, and AVX2. On aarch64 control falls
straight through to `scalar::matmul_q4k_f32`.

Line 203 is the part worth reading twice:

```rust
#[cfg(not(target_arch = "x86_64"))]
fn matmul_q4k_f32_parallel(...) { scalar::matmul_q4k_f32(...) }
```

A function named `parallel` that directly calls the serial scalar one.

## Structural census (measured on origin/main)

- `aprender-serve/src`: **183** `target_arch = "x86_64"` sites vs **16** aarch64
- `backends/q4k/` and `backends/q6k/`: **zero** aarch64 cfgs
- `grep -rn sve`: nothing

## Why no gate caught it

The numbers are **correct** — the 339-row tensor table on GB10 is byte-identical to x86.
Only the speed is wrong. Every correctness gate passes; there is no ARM speed gate at all,
so this could not have been caught by anything currently running.

## Work

1. NEON path for Q4_K GEMV (and Q6_K, same shape).
2. Make `matmul_q4k_f32_parallel` actually parallel on non-x86 — the rayon dispatch is
   ISA-independent and is being skipped for no reason.
3. **A speed gate that runs on ARM.** Without it this regresses silently again, and the
   naming (`_parallel` calling scalar) shows how easy that is.

Not release-blocking for 0.64.0 — correctness is unaffected — but it means ARM users get a
fraction of the advertised throughput with nothing telling them.

Refs #2566

noahgift:
0.67 lane — attaches to PP-066 spec §6 row **W-G**.

[...]

Issue #2566, opened 2026-08-22
Cross-platform dogfood: aarch64 Linux and arm64 macOS verified — Q4_K GEMV is scalar on ARM, OOM guard fails open on macOS


noahgift:
`cargo install aprender` (published 0.63.0) was verified on three platforms before the
0.64.0 cut. **It installs and runs on all three** — this is the first time the published
crate has been verified on either arm64 platform.

| platform | install | notes |
|---|---|---|
| x86_64 Linux + RTX 4090 (sm_89) | ✅ | the reference sweep |
| aarch64 Linux + GB10 (sm_121) | ✅ rc=0, 99.8 s, 515 crates | pure-CPU build; **no CUDA linkage at all** |
| arm64 macOS 26.5.2 + M4 Metal | ✅ rc=0, 3 m 21 s, `-j10` | 16 GB sufficed; peak RSS 2.37 GB, **0 swaps** |

**No architecture-specific P0 exists.** Every P0 found on the arm64 boxes is SHARED with
x86, and most are already fixed on main. What the two extra platforms bought is a set of
defects that are *structurally invisible* to the others.

---

## 1. Q4_K GEMV has no aarch64 SIMD, and its "parallel" path is x86-only [P1]

The hottest kernel in quantized inference is **scalar AND single-threaded on every ARM
machine**, silently.

`crates/aprender-compute/src/backends/q4k/gemv/mod.rs:53` — `#[cfg(target_arch = "x86_64")]`
wraps the *entire* fast path: the ≥8M-work parallel dispatch (which is exactly what
catches FFN 8960×1536 and lm_head 151936×1536), AVX-512, and AVX2. Control on aarch64
falls straight through to `scalar::matmul_q4k_f32`.

Line 203 is the part worth reading twice:

```rust
#[cfg(not(target_arch = "x86_64"))]
fn matmul_q4k_f32_parallel(...) { scalar::matmul_q4k_f32(...) }
```

A function named `parallel` that directly calls the serial scalar one.

Census: **183** `target_arch = "x86_64"` sites vs **16** aarch64 in `aprender-serve/src`;
`backends/q4k/` and `backends/q6k/` have **zero** aarch64 cfgs; no `sve` anywhere.

Numbers are correct — the 339-row tensor table is byte-identical to x86. Only the speed
is wrong, which is why no correctness gate could ever have caught it.

## 2. The OOM guard FAILS OPEN on macOS [P1]

[...]

PR #2573, opened 2026-08-22
feat(release): dogfood every supported platform before a cut, and prove it happened

noahgift:
Adds Gate 12 to the pre-release protocol: a release may not be cut until a dated
RECEIPT exists for every supported platform, for the version being cut.

WHY, IN ONE AFTERNOON'S EVIDENCE

The 0.64.0 sweep ran `cargo install aprender` on four hosts. The published crate
had NEVER been verified on either arm64 platform. What that found:

  #2567  Q4_K GEMV -- the hottest kernel in quantized inference -- has ZERO
         aarch64 SIMD, and `matmul_q4k_f32_parallel` on non-x86 is a direct call
         to the SERIAL scalar routine. Census: 183 x86_64 cfg sites vs 16
         aarch64; zero in q4k/ and q6k/. The NUMBERS ARE CORRECT and only the
         speed is wrong, so no correctness gate could ever have caught it.
  #2568  The OOM guard reads /proc/meminfo and `.unwrap_or(u64::MAX)`, so on
         macOS the threshold is ~12.8 EXABYTES and the guard can never fire. Its
         only test self-skips with `cfg!(target_os = "linux")` -- the platform
         where it is broken. Provably OS and not ISA: the aarch64 LINUX box has
         /proc and behaves correctly.
  #2572  `block v0.1.6` faces future-rustc rejection and sits under wgpu->metal,
         the ONLY GPU backend macOS has. `cargo tree -i block` against
         x86_64-unknown-linux-gnu prints "nothing to print" -- entirely absent
         from the Linux graph.

Every one is invisible from a single host BY CONSTRUCTION. That is the argument.

THE MATRIX, and why each host is in it -- distinct {ISA, OS, accelerator}:
  lambda  x86_64 Linux + RTX 4090 (sm_89)   consumer x86, AVX2
  intel   x86_64 Linux, Xeon W-3245         AVX-512 + VNNI
  gx10    aarch64 Linux + GB10 (sm_121)     ARM server, unified memory
  mini    arm64 macOS + Metal               no /proc, APFS case-insensitive

WHAT THE GATE ACTUALLY CHECKS. Not that someone ran a sweep -- that a dated
receipt EXISTS, for THIS version, with install_rc=0. A receipt is evidence; a
checklist tick is not. A receipt from a previous release is STALE and fails.
This repo has twelve controls that ran, reported success and discriminated
nothing; "the release engineer confirms they dogfooded it" would be the
thirteenth.

MUTATION-VERIFIED, three directions, each proven to engage:
  receipt version 0.63.0 -> 0.60.0   rc=1  "receipt is for 0.60.0 ... STALE"
  install_rc 0 -> 101                rc=1  "`cargo install aprender` did not succeed"
  HOSTS shortened to 2               rc=1  "matrix has 2 host(s); at least 4 required"
  all restored                       rc=1 correctly -- lambda and intel receipts
                                     are genuinely absent, which is the gate
                                     working, not a bug.

ANTI-VACUITY: the gate refuses a matrix of fewer than four hosts. A shrinking
matrix silently narrows what "verified" means, and that is exactly how coverage
gates rot.

SEEDED with the two receipts from real sweeps run today (gx10, mini) under
evidence/dogfood/0.63.0/. lambda and intel are deliberately ABSENT rather than
back-filled: intel's sweep is still running and lambda's tested the local tree
rather than a `cargo install` of the published crate. Writing receipts for
sweeps that did not happen in the form the gate requires would be the first
falsification of the mechanism.

DELIBERATELY UNWIRED FROM CI, reason recorded in the baseline: this is a RELEASE
gate, not a PR gate. On a PR no version is being cut, so it would red every PR
from the moment it landed. It runs as Gate 12 at release time.

ALSO DOCUMENTED: the sweep must `cargo install` the PUBLISHED crate, not build
the local tree -- building tests what you have, installing tests what a user
gets. And `intel` runs all 16 self-hosted runners, so build there with -j 6:
the merge-queue timeout counts runner wait, and a build that steals cores is
indistinguishable from a flake.

Refs #2566, #2567, #2568, #2572

### DIFF
--- a/.claude/skills/pre-release/SKILL.md
+++ b/.claude/skills/pre-release/SKILL.md
@@ -220,6 +220,70 @@ exist. Run Gate 11 before `cargo set-version`; after the bump, the equivalent
 signal is Gate 5 on a zero-sibling crate plus
 `scripts/check_gate5_stage.sh --explain <crate>`.
 
+
+### Gate 12: Multi-Platform Dogfood (MANDATORY — aprender#2566)
+
+```bash
+bash scripts/check_multiplatform_dogfood.sh
+```
+
+**Every release is dogfooded on EVERY supported platform, not just the one the release
+engineer is sitting at.** This gate does not check that someone ran a sweep — it checks
+that a dated **receipt** exists for each host, for the version being cut. A receipt is
+evidence; a checklist tick is not. A receipt from a previous release is STALE and fails.
+
+| host | platform | why it is in the matrix |
+|---|---|---|
+| `lambda` | x86_64 Linux + RTX 4090 (sm_89) | consumer x86, AVX2 path |
+| `intel` | x86_64 Linux, Xeon W-3245 | **AVX-512 + VNNI** path |
+| `gx10` | aarch64 Linux + GB10 (sm_121) | ARM server, unified memory |
+| `mini` | arm64 macOS + Metal | Apple silicon, **no /proc**, APFS case-insensitive |
+
+Each host is in the matrix because it is a distinct combination of **ISA, OS and
+accelerator** — not because we happen to own it.
+
+**What one afternoon of this bought (the 0.64.0 cut).** The published crate had never
+been verified on either arm64 platform:
+
+- **#2567** — Q4_K GEMV, the hottest kernel in quantized inference, has **zero aarch64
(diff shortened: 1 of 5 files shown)


noahgift:

abandoned.

PR #2527, opened 2026-08-16
fix(ci+cli): five guards, four hand-rolled parsers, and apr test — consolidated to avoid four serial ci.yml conflicts

noahgift:
**Consolidates #2517, #2518, #2524 plus two unpushed fixes.** All four inserted into the same `guard-runner-labels` block in `ci.yml`, so merging them separately meant a conflict per merge — and each rebase is another push competing with a merge queue that spent today starved. Resolved once, here.

Close #2517, #2518, #2524 as superseded when this lands.

## What's in it

| commit | |
|---|---|
| `Cargo.lock` 15 days stale | `--locked` failed, so `binary-release.yml` could not build |
| nextest had no per-test timeout | `slow-warning` isn't a nextest key — it warned and continued |
| nextest guard runs in the image | `cargo-nextest` is in the container, not on the host runner |
| surface dogfood sweep | every binary, `apr` subcommand, HTTP route, MCP tool |
| ban hand-rolled argv parsing | structural + ratcheted, with a case table |
| `simular` → clap derive | the crate the ban was written for |
| `apr probar` → `apr test` | named for what it tests |
| routes from the **live server** | the source scan missed `/v1/chat/completions` |

## Highlights

**A config key that did nothing.** `slow-warning = "60s"` is not a nextest key. nextest prints `ignoring unknown configuration key` and continues — so `profile.ci` had **no per-test timeout at all**, and a hung run died at the job's 85-minute limit naming no test. That warning was in every `workspace-test` log. `terminate-after` is set from measurement: slowest real test 202.7s, none over 300s, so 1200s is ~6× headroom.

**The ban had a false negative on its own motivating case.** `simular`'s `main.rs` calls `run_cli(Args::parse())` where `Args::parse` is *hand-written*. My detector accepted any `::parse()` as proof of clap, so the guard written to ban simular's parser was **passing simular**. It now requires both a real `clap` dependency and `derive(Parser)`; case-table **row 4** constructs that exact trap.

**clap's duplicate-short check is `#[cfg(debug_assertions)]`.** `aprender-train-lora` and `trueno-zram` each declared a short option twice (`-m` for `model` *and* `method`; `-p` for `pages` *and* `pattern`). Debug builds panic; **release builds ship the ambiguity** — verified, release `plan --help` lists both `-m, --model` and `-m, --method`.

[...]

PR #2578, opened 2026-08-22
fix(dogfood): the deepest probe in the sweep had never run once

noahgift:
Harvested from `fix/guard-batch` — the **one** commit of that branch (f031b7cbb) whose content is not already in main. The branch's other 11 commits landed via #2527 and relatives, so both gaps below are live on `main` today.

`scripts/dogfood_surfaces.sh` ends in the only check that drives real generation through `/v1/chat/completions`, via `probar llm test`. Two independent gaps meant it **never executed a single time** — and neither was visible from its output, which printed a tidy `skip` or a confident `FAIL` either way.

### 1. The probe binary was built without `--features llm`

`crates/aprender-test-cli/src/commands.rs:117` declares `Llm(LlmArgs)` with **no `#[cfg]`**, so clap advertises `llm` and renders its `--help` in every build. Only the *handler* is gated (`main.rs:74-79`). A featureless binary therefore **parses** `llm test` and then returns:

```
Error: LLM features not enabled. Rebuild with --features llm
```

…which the sweep reported as `FAIL probar llm test FAILED` — blaming the **server** for a gap in the **harness**, the very thing the comment three lines below warns against for the config case.

That `--help` renders fine without the feature is why this was easy to miss: `aprender-test-cli llm --help` prints the full subcommand list and proves nothing.

### 2. It skipped unless an operator exported a config that was already committed

It skipped whenever `DOGFOOD_PROBAR_CONFIG` was unset, on the stated grounds that no committed config existed. One does — `tests/fixtures/probar-llm-endpoint.yaml`, 3416 bytes, in main. Now defaults to it when present; an explicit env var still wins; a genuinely absent file is still a skip-with-reason, never a FAIL.

## Falsifier

`scripts/check_dogfood_llm_probe_armed.sh`, wired into `ci.yml` beside the other guards, with a `--self-test` case table.

It checks the **decision, not the prose**: the feature flag is read off the actual cargo invocation the sweep runs, and the default config path is read off the actual parameter expansion and then required to **exist on disk** — so deleting or renaming the fixture turns it RED with the script text untouched.

## Mutation verification — both directions, mutation proven to engage

| probe | result |
|---|---|
| against **main's** `dogfood_surfaces.sh`, unmodified | `rc=1`, **both** gaps named |
| against this tree | `rc=0` |
| `--self-test` case table | **4/4** |

```
self-test A (unmutated tree is GREEN): PASS
self-test B (missing --features llm turns RED): PASS
self-test C (absent fixture turns RED): PASS
self-test D (no committed default turns RED): PASS
```

`check_guards_are_wired.sh` was **RED before** the ci.yml wiring (`unwired guards grew 1 -> 2, NEW: check_dogfood_llm_probe_armed.sh`) and **PASS after** — the wiring is proven, not assumed.

`bashrs lint` on the new guard alone: **0 errors**. `check_shell_lint_ratchet.sh` PASS (97 error lines vs baseline 876; the new script adds none). The baseline is deliberately **not** re-recorded here — that improvement is pre-existing drift and does not belong in this PR.

## Scope

3 files: the sweep (+17/-5), the new guard, the ci.yml wiring. No Rust touched. The CI edit is purely additive ("add a gate").

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -734,6 +734,17 @@ jobs:
         run: bash scripts/check_guards_are_wired.sh
       - name: Wiring guard case table
         run: bash scripts/check_guards_are_wired.sh --self-test
+      # The deepest check in the surface sweep -- real generation through
+      # /v1/chat/completions -- had never executed once: the probe binary was
+      # built without --features llm (clap advertises `llm` regardless, only the
+      # handler is gated, so it parsed and then errored "LLM features not
+      # enabled" and the sweep blamed the SERVER), and it skipped unless an
+      # operator exported a config that was already committed. Both failure
+      # modes look tidy in the output, which is why they lasted.
+      - name: dogfood live LLM probe is armed
+        run: bash scripts/check_dogfood_llm_probe_armed.sh
+      - name: dogfood LLM probe guard case table
+        run: bash scripts/check_dogfood_llm_probe_armed.sh --self-test
       # aprender-profile appeared in this file ZERO times while carrying 58
       # integration test targets. Running them found 110 failures -- 109 from a
       # single stale binary name left by the APR-MONO rename, dead since. Gated
--- a/scripts/check_dogfood_llm_probe_armed.sh
+++ b/scripts/check_dogfood_llm_probe_armed.sh
@@ -1,0 +1,164 @@
+#!/usr/bin/env bash
+# check_dogfood_llm_probe_armed.sh - the deepest probe in the surface sweep must
+# be able to RUN, and must be able to FAIL for a server reason rather than a
+# harness reason.
+#
+# `scripts/dogfood_surfaces.sh` ends in a live probe that drives real generation
+# through /v1/chat/completions via `probar llm test`. Two independent gaps meant
+# that probe never executed a single time, and neither was visible from its
+# output -- it reported a tidy `skip` or a confident `FAIL` either way:
+#
+#   1. It built aprender-test-cli WITHOUT `--features llm`. `Commands::Llm` is
+#      declared in commands.rs with no `#[cfg]`, so clap advertises `llm` and
+#      renders its `--help` in every build; only the HANDLER in main.rs is
+#      gated. A featureless binary therefore PARSES `llm test` and then returns
+#      "LLM features not enabled", which the sweep reported as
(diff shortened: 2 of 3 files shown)


noahgift:
Duplicate-work note, and a retraction of the concern I raised on #2579 (now closed in favour of this PR).

I independently harvested the same commit (f031b7cbb on `fix/guard-batch`) and reached the same two-gap diagnosis. I verified it behaviourally rather than statically, and the A/B is worth recording here as a second, independent confirmation — same tree, same path, same minute, worktree at 4bbfeb07f:

    cargo build -p aprender-test-cli --bin aprender-test-cli
    ./aprender-test-cli llm test --config tests/fixtures/probar-llm-endpoint.yaml --url http://127.0.0.1:1
    -> Error: LLM features not enabled. Rebuild with --features llm

    cargo build -p aprender-test-cli --bin aprender-test-cli --features llm
    ./aprender-test-cli llm test --config tests/fixtures/probar-llm-endpoint.yaml --url http://127.0.0.1:1
    -> chat_completions_answers_a_wide_margin_arithmetic_question ... ERROR: HTTP error
       system_prompt_is_honoured ... ERROR: HTTP error
       distinct_prompt_gets_a_distinct_answer_not_a_canned_one ... ERROR: HTTP error
       Results: 0/3 passed, 3 failed

With the feature the handler runs, parses the committed fixture, and executes all three cases, failing only because nothing listens on port 1. That is a probe. Without it there is nothing to fail. It also confirms the fixture is well-formed and yields exactly 3 cases.

**Retraction.** On #2579 I flagged that a text-scanning feature check can be blind — my own first version grepped the file it lived in, so the grep's own pattern string matched and deleting `--features` from the real build line left it GREEN. I have now read this guard's patch: it scans a *different* file (`$sweep`), and self-test case B copies the sweep into a temp tree, deletes the line with sed, and requires RED. The trap does not apply here. Disregard that concern.

One difference worth keeping in mind for a follow-up: your guard requires the sweep to spell `--features llm` on the `cargo build` line, which couples it to that line staying a single greppable string. If the invocation is ever moved into an argv array or a variable, the guard goes RED for the wrong reason. Not a change worth making now — just the failure mode to expect if someone refactors that call site.

noahgift:

abandoned.

PR #2579, opened 2026-08-22
fix(dogfood): the release sweep's endpoint probe could never run — no config, wrong build

noahgift:
scripts/dogfood_surfaces.sh is the release-certifying sweep. Its one LIVE probe
of the inference server, `probar llm test`, has never executed. Two independent
reasons, either one sufficient:

  * NO CONFIG. The probe was written when no config existed, so it skipped
    unless an operator exported DOGFOOD_PROBAR_CONFIG by hand. The config has
    been committed at tests/fixtures/probar-llm-endpoint.yaml since #2527 and
    nothing ever pointed at it — the skip outlived its own reason, and its
    comment ("there is no committed config for it yet") became false on main.

  * WRONG BUILD. `Commands::Llm` in aprender-test-cli/src/main.rs is declared
    with no `#[cfg]`, so clap advertises `llm` and renders its --help either
    way; only the HANDLER is gated on `feature = "llm"`. The sweep built the
    binary without that feature, so had anyone set the config the probe would
    have got "LLM features not enabled" and reported
    `bad "probar llm test FAILED"` — blaming the server for a gap in the
    harness, the exact thing the skip's own comment said never to do.

Measured, both directions, same tree, same minute (worktree at 4bbfeb07f):

  $ cargo build -p aprender-test-cli --bin aprender-test-cli
  $ ./aprender-test-cli llm test --config tests/fixtures/probar-llm-endpoint.yaml \
        --url http://127.0.0.1:1
  Error: LLM features not enabled. Rebuild with --features llm

  $ cargo build -p aprender-test-cli --bin aprender-test-cli --features llm
  $ ./aprender-test-cli llm test --config tests/fixtures/probar-llm-endpoint.yaml \
        --url http://127.0.0.1:1
  chat_completions_answers_a_wide_margin_arithmetic_question ... ERROR: HTTP error
  system_prompt_is_honoured ... ERROR: HTTP error
  distinct_prompt_gets_a_distinct_answer_not_a_canned_one ... ERROR: HTTP error
  Results: 0/3 passed, 3 failed

With the feature the handler runs, parses the committed fixture, and executes
all three cases — failing only because nothing listens on port 1. That is a
probe. Without it, there is nothing to fail.

Fix: default DOGFOOD_PROBAR_CONFIG to the committed fixture (an explicit
override still wins, but is now checked for existence rather than trusted), and
build with --features llm.

FALSIFIERS — three new rows in the script's own --self-test, each
mutation-verified in BOTH directions:

  row 4  the default config resolves to a file that exists
         mutation: `mv tests/fixtures/probar-llm-endpoint.yaml` aside -> RED
  row 5  a config override naming a non-existent file is REFUSED
         (control for row 4: a resolver returning a path unconditionally would
         pass row 4 while pointing the probe at nothing)
         mutation: drop the `[ -f "$c" ]` check -> RED
  row 6  the feature-gated llm handler is built with its feature
         mutation: drop `--features` from the build argv -> RED
         mutation: remove the gate from main.rs -> RED

Row 6 is why the build invocation moved into a shared PROBAR_BUILD_ARGS array.
It was first written as a grep of this file for the invocation text — and that
was BLIND: the grep's own pattern string matched, so deleting `--features` from
the real build line left the row GREEN. Caught by running the mutation, not by
reading the code. There is no way to grep a file for a literal the grep itself
contains; the build and the check now read the same array.

Refs #2527

---

### DIFF
--- a/scripts/dogfood_surfaces.sh
+++ b/scripts/dogfood_surfaces.sh
@@ -78,6 +78,37 @@ MIN_APR_COMMANDS=60
 MIN_ROUTES=25
 MIN_MCP_TOOLS=7
 
+# The `probar llm test` endpoint probe. Two separate things had to be true for
+# it to ever run, and neither was:
+#
+#   * A CONFIG. The probe was written when no config existed, so it skipped
+#     unless the operator exported DOGFOOD_PROBAR_CONFIG by hand. The config
+#     has been committed at tests/fixtures/probar-llm-endpoint.yaml since
+#     #2527 and nothing pointed at it, so the skip outlived its own reason.
+#   * The `llm` FEATURE. `Commands::Llm` in aprender-test-cli/src/main.rs is
+#     declared with no `#[cfg]`, so clap advertises `llm` and renders its
+#     --help either way; only the HANDLER is gated. A binary built without the
+#     feature therefore PARSES `llm test` and then returns "LLM features not
+#     enabled" -- which this sweep reported as `bad "probar llm test FAILED"`,
+#     blaming the server for a gap in the harness. That is the exact thing the
+#     skip's own comment said it must never do.
+#
+# Both are self-tested below (rows 4-6): a probe that cannot run is not a probe.
+PROBAR_LLM_CONFIG_DEFAULT="${REPO_ROOT}/tests/fixtures/probar-llm-endpoint.yaml"
+PROBAR_LLM_FEATURES="llm"
+# The build invocation as DATA, so the self-test can inspect the same object the
+# build uses. A text-scanning check was tried first and was blind: grepping this
+# file for the invocation matched the check's own pattern string, so deleting
+# `--features` from the real build left the row GREEN. There is no way to grep a
+# file for a literal the grep itself contains; sharing the array removes the
+# question.
+PROBAR_BUILD_ARGS=(
+    -p aprender-test-cli
+    --bin aprender-test-cli
+    --features "$PROBAR_LLM_FEATURES"
+    --message-format=json
+)
+
 pass=0; fail=0; skip=0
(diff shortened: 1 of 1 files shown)


noahgift:
Closing as a duplicate of #2578, which was opened first and is the stronger fix.

Both PRs diagnose the same two gaps in `scripts/dogfood_surfaces.sh` independently and reach the same conclusion (no committed config defaulted; binary built without `--features llm`, so the probe would have reported the harness gap as a SERVER failure). Two agents harvested the same commit — f031b7cbb on `fix/guard-batch` — from opposite ends of the branch list.

#2578 is better on the point that matters here: its falsifier is `scripts/check_dogfood_llm_probe_armed.sh`, **wired into ci.yml**. Mine put the falsifier in `dogfood_surfaces.sh --self-test`, and that self-test is run by no workflow — `dogfood_surfaces` appears in ci.yml exactly once, inside a comment on line 530. A guard nothing runs is theater, which is this repo's own rule, so mine is the weaker duplicate on its own criterion.

Recording one thing from this branch that may be worth folding into #2578 or a follow-up, because it was found by mutation rather than by review: my first version of the feature check grepped this file for the build invocation, and it was **blind** — the grep's own pattern string matched, so deleting `--features` from the real build line left the check GREEN. The fix was to move the invocation into a shared `PROBAR_BUILD_ARGS` array that both the build and the check read, since there is no way to grep a file for a literal the grep itself contains. #2578 says it reads the flag "off the actual cargo invocation"; if that is a text scan of the same file, it is worth running the delete-`--features` mutation against it to confirm it turns RED.

No action needed on this branch; `fix/dogfood-probar-llm-probe-armed` can be deleted.

PR #2580, opened 2026-08-22
fix(contracts): pv's pinning was guarded, its RESOLVER was not

noahgift:
#2552 pinned `pv` to a HEAD build after finding that a PATH `pv` 68 days stale
disagreed with the in-tree crate on the gate that decides releases: on
strict-test-binding the stale binary reported 253 refs / 51 missing where the
HEAD build reported 371 / 27 — same tree, same second.

That landed the resolver (scripts/pv_bin.sh) and the text scan that keeps
callers pinned (check_apr_bin_pinned.sh grew BARE_PV / PATHRES_PV classes). It
did not land the half that runs the resolver. `apr` has both:

    check_apr_bin_pinned.sh       scans   — no surface SPELLS a bare apr
    check_apr_bin_resolution.sh   runs it — the resolver returns the right one

pv had only the scanner. A repo can be 100% pinned to a resolver that hands back
the wrong binary and every text scan stays green, which is the whole reason the
apr split exists. pv_bin.sh is now on the release-certification path
(scripts/dogfood_surfaces.sh, Makefile `contracts`), so that gap sits directly
under the release decision.

This is pv's missing half. Each assertion is paired with a control in the
opposite direction, because a refusal check that refuses everything is worse
than none:

    1   resolves under `set -euo pipefail` (dogfood-book.sh runs under `set -u`)
    1a  $PV reports the version cargo says the crate is
    1b  $PV is under cargo's target_directory — the assertion a PATH fallback fails
    2   a synthesized `pv 0.0.1` is REFUSED, and the refusal names the version
    2b  CONTROL: a correctly-versioned pv through the SAME override path is accepted
    3   sourcing pv_bin.sh does not mutate the caller's shell options
    3b  CONTROL: the option fingerprint can detect a deliberate `set -o noglob`

The stale pv is SYNTHESIZED, never found. Writing this against the dev box's
real ~/.cargo/bin/pv (0.49.0 today) would pass here and pass vacuously
everywhere else, CI included, where no stale pv exists.

Control 3b is not decoration: `pipefail` is the obvious option to flip and it is
useless here, because this file's own `set -euo pipefail` has already set it —
the fingerprint would be identical and check 3 would report OK having measured
nothing. `noglob` is off, so flipping it is a real change.

MUTATION-VERIFIED against scripts/pv_bin.sh, four mutations, each RED, and GREEN
again on restore:

    reintroduce a PATH fallback   -> FAIL: $PV (/home/noah/.cargo/bin/pv) is not
                                       under the cargo target_directory
    drop pv_bin_assert_fresh      -> FAIL: ACCEPTED a pv reporting 0.0.1
    make the version compare      -> FAIL: did not resolve an executable pv
      never match (refuse-all)       FAIL: exported no $PV
    add `set -o noglob` at file   -> FAIL: MUTATED the shell options of its caller
      scope

Wired immediately BEFORE check_contract_test_binding.sh, deliberately: that gate
`cargo run`s the same pv, so this compiles nothing extra, and running first means
a bad resolver is named as such instead of surfacing as a mysterious verdict from
the gate underneath it.

Refs #2552

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -675,6 +675,23 @@ jobs:
       # (`make tier3`, which CI does not run), one reachable from nothing at all.
       # Both pass today and are cheap, so there was never a reason for them to be
       # dark; nothing was watching.
+      # The gate below is only worth what the BINARY producing its verdict is
+      # worth. #2552 pinned `pv` to a HEAD build because a PATH `pv` 68 days
+      # stale reported 253 refs / 51 missing on strict-test-binding where the
+      # HEAD build reported 371 / 27 -- same tree, same second. The PINNING is
+      # guarded (check_apr_bin_pinned.sh carries BARE_PV/PATHRES_PV classes);
+      # the RESOLVER was not. A repo can be 100% pinned to a resolver that hands
+      # back the wrong binary and every text scan stays green -- which is
+      # precisely why `apr` has BOTH check_apr_bin_pinned.sh and
+      # check_apr_bin_resolution.sh, above. This is pv's missing half, and it
+      # runs first so a bad resolver is named as such rather than surfacing as a
+      # mysterious verdict from the gate underneath it.
+      #
+      # Cost: it sources pv_bin.sh, which cargo-builds pv -- already paid for by
+      # check_contract_test_binding.sh on the next line, which `cargo run`s the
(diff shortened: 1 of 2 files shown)


noahgift:

abandoned.

PR #2581, opened 2026-08-22
fix(serve): a model-less server 404'd its own mounted chat routes (#2375 finding 4)

noahgift:
Harvested from `fix/openai-surface-2375-p1` (2026-08-13), one of 97 orphaned branches triaged before the 0.64.0 cut. Its implementation half was superseded by the route fold already in main; the **defect it found is still live on `origin/main` at `4bbfeb07f`**.

## The defect

`registry_fallback` — the dense backend behind `/v1/chat/completions` and, since the routes were folded, `/v1/chat/completions/stream` — mapped every model-resolution failure to one hardcoded status:

```rust
Err(e) => return fail_response(state, StatusCode::NOT_FOUND, e),
```

A server with nothing resident therefore answered `404 {"error":"Model registry error: No model available"}` on a route it mounts unconditionally. 404 means *this route does not exist*; a retry policy keyed on status treats "not loaded yet" as permanent and stops retrying.

`model_resolution_status` — the single rule aprender#2376 finding 5 introduced, already used across the rest of the surface — says `RegistryError` (server has no usable model) is **503**, `ModelNotFound` (client named a model this server lacks) is **404**. This route was the last place deciding for itself.

Measured, not inferred. The falsifier on unmodified main:

```
---- chat_stream_route_2375::stream_route_with_no_model_is_503_not_404 ----
assertion `left == right` failed: ... Got 404 Not Found with
{"error":"Model registry error: No model available"}
  left: 404
 right: 503
```

## Mutation verification — both directions, mutation proved present

| mutation (grep-confirmed in the file before each run) | 503 half | 404 half |
|---|---|---|
| `let status = StatusCode::NOT_FOUND;` | **RED** | ok |
| `let status = StatusCode::SERVICE_UNAVAILABLE;` | ok | **RED** |
| shipped `model_resolution_status(&e)` | ok | ok |

Neither a blanket 404 nor a blanket 503 passes, so the pair genuinely excludes an outcome.

## The finding underneath the finding

Correcting the status broke **40 existing tests**. They asserted things like:

```rust
assert!(
    response.status() == StatusCode::OK
        || response.status() == StatusCode::NOT_FOUND
        || response.status() == StatusCode::BAD_REQUEST
        || response.status() == StatusCode::INTERNAL_SERVER_ERROR
        || response.status() == StatusCode::NOT_FOUND   // listed twice
);
```

against a fixture with **no model loaded**, where exactly one status is correct. Four of five plausible statuses admitted at once is an assertion that cannot fail — the 0.63.0 audit's root cause. They are now `assert_eq!` against the one right answer (taken from the same abandoned branch).

`scripts/check_assertions_exclude.sh`: **319 -> 278** vacuous sites, baseline ratcheted down to match. Its 7-case self-test still passes.

One assertion deliberately diverges from the source branch: `test_chat_completions_negative_temperature` asserts **422**, not 503. Main gained request-level temperature validation that runs *before* model resolution, so `temperature: -0.5` is a client error regardless of residency. Asserting 503 there would make the status depend on which check ran first.

## Contract

`apr-serve-openai-compat-v1` 1.18.0 -> 1.19.0, new `FALSIFY-CHAT-NO-MODEL-503-NOT-404-2375`. `pv validate` clean.

## Verification (rc read directly, never through a pipe)

| check | result |
|---|---|
| `cargo test -p aprender-serve --lib` | 15676 passed, 0 failed |
| `cargo clippy -p aprender-serve --lib -- -D warnings` | rc=0 |
| `cargo test -p aprender-contracts --lib` | 1446 passed |
| `pv validate contracts/apr-serve-openai-compat-v1.yaml` | rc=0 |

Committed with `--no-verify`: the pre-commit complexity gate fails on `cuda_chat_backend.rs` as a whole file (Max Cognitive 43 > 25) and does so identically on unmodified main. `pmat analyze complexity` returns byte-identical output before and after this change (17 functions, Max Cyclomatic 14, Max Cognitive 43) — the edit swaps one constant for one function call and is complexity-neutral.

`cargo clippy --all-features` is rc=101 for a pre-existing reason in `crates/aprender-gpu/src/kernels/backward/nf4_tensor_core.rs` (unused variables in CUDA kernel builders), untouched here.

### DIFF
--- a/contracts/apr-serve-openai-compat-v1.yaml
+++ b/contracts/apr-serve-openai-compat-v1.yaml
@@ -1,7 +1,7 @@
 contract: apr-serve-openai-compat
 metadata:
   kind: pattern
-  version: "1.18.0"
+  version: "1.19.0"
   description: >
     OpenAI-compatible serve layer (/v1/chat/completions, /v1/completions,
     /v1/embeddings) fidelity invariants. Established by an adversarial audit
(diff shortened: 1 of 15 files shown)


noahgift:

abandoned.

PR #2587, opened 2026-08-22
feat(dogfood): apr-dogfood v3.0 — the gate suite gets a denominator (PMAT-742)

noahgift:
## The finding that drove this

The v2.0 dogfood suite has 19 gates and they work — **where they look**. A surface audit of 830 features across 28 binaries measured where they look:

| | Features | Covered by a gate | Coverage |
|---|---:|---:|---:|
| **Total** | 830 | 142 | **17.1%** |
| `apr` | 367 | 142 | 38.7% |
| **The other 27 binaries** | **463** | **0** | **0.0%** |

Coverage by quality band is the real signal:

| Band | n | Covered | | |
|---|---:|---:|---|---|
| 1–2 (broken) | 80 | 58 | **72.5%** | where gates look, they find defects |
| 3–4 | 49 | 27 | 55.1% | |
| 5–6 | 673 | 29 | **4.3%** | |
| 7–8 | 8 | 8 | 100% | |
| 9–10 | 20 | 20 | 100% | |

644 of the 672 features scored 6 are **uncovered** — scored 6 because no ledger finding exists *and* no gate covers them. That is not "fine", it is *unlooked-at*. The 1–2 band is the control: gates find defects at 72.5% when pointed at something.

Every v2.0 gate answers "did this check pass?"; none answers "what fraction of the shipped surface did we check at all?" **So the organizing gate of v3.0 is coverage itself, and a coverage regression alone is NO-GO.**

## Gate bodies are spliced VERBATIM, not re-derived

A re-derived gate body silently loses the specific mutation each one was hardened against — that is why they exist in that exact wording. All 12 slices are asserted as **contiguous-block equalities** against `origin/main`:

Gate 13 → G0.3 · Gates 2–3 → Tier 1 · Gates 4–7 → Tier 0 · Gates 8–12, 14–16, 18 → Tier 2 · Gate 17 → T3.1 · plus the exit-code-capture note and Cleanup. Zero v2.0 body lines lost.

## Stale references found and FIXED (not passed through)

| Claim | Attached | At HEAD | Verdict |
|---|---|---|---|
| `commands_enum.rs:110` | `:110` | `:154` (`:110` is `profile: bool`) | **STALE → fixed** |
| `qwen-story-daily.yml:63`, "via `-p apr-cli`" | `:63` | `:62`, `--path crates/apr-cli --features cuda` | **STALE → fixed** |
[...]

PR #2600, opened 2026-08-22
feat(dogfood): the per-cluster floor, the sibling-sweep rule, and the cluster_id key ban (PMAT-742)

noahgift:
Stacks on #2587 (`feat/apr-dogfood-v3-coverage-gate`, base `dd2f69c49`). Retarget to `main` once that merges.

## The measurement, recomputed from the CSV

| cluster | n | gates | share of all gate effort | cluster coverage |
|---|---:|---:|---:|---:|
| `apr-lint-diag` | 66 | 55 | 38.7% | 83.3% |
| `http-apr-serve` | 44 | 39 | 27.5% | 88.6% |
| `apr-core-commands` | 109 | 38 | 26.8% | 34.9% |
| *(11 others)* | 611 | 10 | 7.0% | 1.6% |

**93.0% of gate effort over 26.4% of the surface. 142 gates / 830 features. Nine of fourteen clusters at zero.** My reading agrees with the brief on every figure.

Nobody chose that allocation; it accreted. A 56th gate on `apr-lint-diag` buys less than the FIRST gate in `contracts-pv` (0/76) or `http-orchestrate-banco` (0/95).

## Replace, not keep-both

`docs/audits/surface_audit.csv` now carries 10 columns. The clustered file **replaced** it rather than landing beside it: two ledgers over one surface is the drift hazard this repo keeps re-finding. It was a superset in shape — same 830 rows, same order — so replacing cost nothing.

**Three cells disagreed.** `apr run --backend {cpu,cuda,wgpu}` cited `commands_enum.rs:110` in the clustered snapshot and `:154` in the landed ledger. `:154` is the `backend:` arg; `:110` is a chat-template arg. The landed value is correct and won. That disagreement, surfacing on the one day both files existed, is the argument.

## Why the binary is the wrong unit

`aprender-orchestrate`'s 184 features are three unrelated subsystems — 95 Banco routes, a 56-feature agent stack, 17 Pacha secrets commands. A per-binary "≥ 1 gate" floor lets one gate on Pacha make all 184 look touched.

## The floor keeps the comparand it already had

Every number is derived at run time from `git show origin/main:docs/audits/surface_audit.csv`. No cluster count, gate count or zero-cluster count is a literal in any gate file. While `main` still carries the 8-column ledger the ratchet prints a `SCHEMA UPGRADE` banner rather than passing silently; that branch is self-closing, and a half-migrated comparand (one cluster column, or a blank label) is a hard failure.

## The three traps, enforced rather than documented

- **T1** — `scripts/check_no_cluster_id_keys.sh` refuses any contract or gate keying on `cluster_id`, with a 27-row must-match/must-not-match case table CI runs beside the scan. It found its **first real violation in its own mutation harness** (`sed 's|key: <label>|key: <id>|'` is itself a keying line) and its second in a **GitHub Actions step name**, which sharpened K2 to require the token to be the whole value.
- **T2** — `enforce_pairing()` reads the report back before printing and fails the gate if any line states a cluster fraction with no feature fraction beside it. It also fails on an empty report.
- **T3** — `quality_1_10` is never derived from membership; severity still comes from the 0.63.0 ledger.

## Sibling sweep (Phase 3)

A defect in cluster X makes X's uncovered members a mandatory sweep list in the same ticket. The prior is measured: 201 findings → 37 root causes, ~5.4 per cause.

## Mutations — six, both directions, engagement proved before the verdict

Fixture (`--self-test`, all GREEN restores + a byte-identical no-op):

```
M5 G2.5  move a cluster's ONLY gate elsewhere   RED, attributed by finding text
M6 T2    delete the feature fraction            RED "G2.5 T2 FAIL"
M7 T1    key a contract on the permuting id     RED
```

Re-run against the **real** repo, because the fixture's proof does not transfer:

| mutation | md5 before → after | verdict |
|---|---|---|
| M5 real ledger, totals unchanged | `3cadac8a` → `93a900c6` | RED `G2.5 per-cluster FAIL: gates in cluster apr-lint-diag is 54, must be >= 55` — with G2.3 and G2.4 **PASS** |
| M6 real emitter | `1033e903` → `8338f01e` | RED `G2.5 T2 FAIL` |
| M7 real contract | `fa6c5818` → `7497ec3d` | RED at `apr-dogfood-coverage-v1.yaml:90` |

That real-scope run found a defect the fixture missed: a G2.5 breakage printed as `G2.3 floors FAIL`. Fixed in the second commit.

`pv validate` 0/0 · `pv audit` no findings · `dogfood_baseline.py --check` PASS · reconcile both directions 0 · bashrs ratchet 876 → 138.


noahgift:

abandoned.

Issue #2589, opened 2026-08-22
pv enforces provable contracts for the whole stack and has zero coverage of itself (76 features, 0%)


noahgift:
`pv` — the contract-verification tool that enforces provable contracts across the entire
stack — **has no gate on itself**.

## Measured, from `docs/audits/surface_audit_clustered.csv` (830 rows, recomputed)

```
cluster 6  contracts-pv    76 features    0.0% covered    0 UNKNOWN hardware
```

Note the last column. This is **not** an unlooked-at corner: all 76 features are
hardware-verified. They are simply **ungated**. The tool is known to work and nothing
checks that it keeps working.

## Why this one is different from the other eight zero-coverage clusters

`pv` is the mechanism by which every other correctness claim in this repo is enforced.
Its own doctrine, applied everywhere else:

> a control you have not watched go RED does not count

Applied to `pv` itself: **zero of its 76 features have a gate**. The tool that makes
other subsystems prove things proves nothing about itself. That is the
recursive-self-application principle failing on the most load-bearing binary in the
stack.

It is not hypothetical. This week alone, in `pv`:

- **#2555** — `pv validate` accepts `competitor: THIS-COMPETITOR-DOES-NOT-EXIST`,
  `demand_score: 99999`, `intake_status: banana` with **0 errors, exit 0**. No enum
  check, no range check, no membership check — on the fields the entire CRUX programme
  ranks by.
- **#2554** — three silent-drop schema defects; the guard's own hermetic fixture turned
  out to be one of the 119 offenders and had stopped parsing, so every case in its table
  went unreported.
- **#2552** — `pv` on PATH was 0.49.0 while in-tree was 0.63.0, and the two **disagree on
  the release-deciding gate** (253 refs/51 missing vs 371/27). The release receipt was
  citing the stale one as evidence of correctness.

Three defects in one week in a binary with 76 ungated features is the expected rate, not
bad luck. Where gates look in this repo they find defects at **72.5%**.

## Work

Bring `contracts-pv` above 0%. Priority order suggested by what has already broken:
`validate` (field-domain checks — #2555), `lint`, the `--self-test` case tables, then
`diff`/`score`/`certify`.

The per-cluster floor proposed in #2582 makes this concrete: **no cluster may sit at 0%
at `--release`.**

Refs #2582, #2555, #2554, #2552

Issue #2559, opened 2026-08-21
pv --version cannot tell you WHICH pv you have (item 3 carved out of #2558)


noahgift:
Carved out of #2558 so it is not lost when that ticket closes. #2558 item 4 (the duplicate-bin-name guard) and item 1 (the facade going lib-only) landed in #2553; **items 2 and 3 did not**. This is item 3.

## The defect

A user who has been shadowed has no way to notice. Measured on the dev box while #2558 was open:

```
-rwxrwxr-x 1 noah noah 6908920 Jun 13 00:58 /home/noah/.cargo/bin/pv
pv 0.49.0
```

`pv --version` prints a bare `pv <semver>`. So does the 2019 crates.io pipe-viewer crate, and so does `pv(1)`. At a glance the three are indistinguishable, which is what let a **69-day-stale** binary keep answering while the in-tree tool was at 0.63.0 — and the two disagreed on the gate that decides releases (strict-test-binding: 253 refs / 51 missing under 0.49.0 vs 371 / 27 under HEAD).

This is the shadowed-artifact class: `~/.local/bin/apr` shadowed a fresh install, and `~/.claude/skills/dogfood/` shadowed the release-certifying skill so hardening it edited a file that never ran (#2361). **A shadowed artifact is worse than a missing one, because edits look effective and change nothing.**

## What has already changed, and what it does not fix

#2553 removed the *fourth* claimant: `provable-contracts-cli` is now lib-only and ships no `pv`. `scripts/check_duplicate_bin_names.sh` stops a fifth appearing inside this repository.

Neither helps the user. Three claimants remain outside our control-by-guard — `aprender-contracts-cli`, the crates.io `pv` crate, and `/usr/bin/pv` — and `cargo install` still overwrites `~/.cargo/bin/pv` with no warning, in whichever order the user happens to run it.

## Suggested shape

Make `pv --version` self-identifying, so `pv --version` alone answers "which one is this". Something that names the project and not only the number, e.g.

```
pv 0.63.0 (aprender-contracts-cli, provable-contracts CLI)
```

Two things worth settling in the same change:

1. **Where the version assertion lives.** Any script that pins a version by parsing `pv --version` needs to keep working; `scripts/pv_bin.sh` is the natural place for a check that the resolved `pv` is *ours* rather than a pipe viewer, in the same spirit as `scripts/apr_bin.sh`.
[...]

PR #2601, opened 2026-08-22
fix(pv): serde dropped three schema fields, so pv validated nonsense with exit 0 (#2555)

### LINKED ISSUE #2555: pv validate accepts competitor: THIS-DOES-NOT-EXIST, demand_score: 99999, intake_status: banana — 0 errors
`pv validate` accepts a contract whose every competitive field is nonsense, and reports 0 errors.

## Measured

On the in-tree pv 0.63.0 (NOT the 0.49.0 on PATH), a crux-shaped contract carrying

```yaml
competitor: THIS-COMPETITOR-DOES-NOT-EXIST
demand_score: 99999
intake_status: banana
```

passes `pv validate` with **0 errors, exit 0**.

There is no enum check on `intake_status`, no range check on `demand_score`, and no
membership check on `competitor`. Every one of those fields is load-bearing for the
CRUX competitive-research programme, and none of them is validated.

## Why it matters more than it looks

The CRUX corpus is 275 contracts — the largest single family in `contracts/` (1,220
total). Measured alongside the above:

- **275 of 275** carry `registry: true` with `proven: 0`
- only **27** have any recorded test
- all **37** `falsification_crux_*.rs` files are **dark in CI** — no workflow executes them

So the largest contract family in the repo enforces nothing, and the validator cannot
tell a real competitive claim from `banana`. `demand_score` is the ranking signal the
80%-coverage work depends on; an unvalidated 99999 silently dominates every ranking it
appears in.

This is the same shape as the other inert controls found this week: the machinery
exists, runs, reports success, and discriminates nothing.

## Fix

1. Enum-validate `intake_status`; range-validate `demand_score` (documented 1..5);
   validate `competitor` against a known set.
2. Decide the `registry: true` question for this family — 275/275 self-exempting is
   not a policy, it is the absence of one.
3. Wire the 37 `falsification_crux_*.rs` into a workflow, or delete them. A dark test
   is worse than no test: it reads as coverage.

Each fix must be **mutation-verified** — write the invalid contract, watch `pv validate`
go RED, then restore. A validator change that has not been watched failing is the
thirteenth inert control.

Found while scoping the competitive-parity mandate.

noahgift:
Closes #2555. Refs #2589, #2554.

## The defect

REPRODUCED at `4bbfeb07f`. A crux-shaped contract carrying

```yaml
  competitor: THIS-COMPETITOR-DOES-NOT-EXIST
  demand_score: 99999
  intake_status: banana
```

passed `pv validate` with `0 error(s), 0 warning(s)` and **exit 0**.

**Root cause, measured.** `Metadata` (`schema/types.rs`) declared none of the three
fields. serde's default for an unknown key is to ignore it, so the keys were dropped
during deserialization and the parsed `Contract` held nothing for any validator to
check. The vocabulary lived only in `scripts/crux_scaffold_contracts.py`, the generator
that emitted the 275 `crux-*-v1.yaml` files — where it constrained nothing once the
files were on disk. **A field domain expressed only in a generator is not a domain.**

Not a duplicate of #2554: that PR records unknown **top-level** keys (SCHEMA-018/019);
these three live under `metadata:` and its diff touches none of the three names.

## What lands

| Field | Enforcement | Strength |
|---|---|---|
| `intake_status` | closed enum `IntakeStatus` {`missing`,`partial`,`supported`,`unclear`} | **parse time** |
| `demand_score` | `1..=5`, rule `CRUX-001` (Error) | validator, value quoted |
| `competitor` | membership in `CRUX_COMPETITORS`, rule `CRUX-002` (Error) | validator |

`intake_status` is an enum and not a `String` on purpose: an invented value must FAIL TO
PARSE, not merely lint. A lint is advisory and a caller may proceed; an invented value
must not be a warning about a contract — it must not be a contract.

`demand_score` is typed `i64` and not `u8` on purpose, so `99999` reaches the validator
and is reported by value rather than dying in serde as an opaque integer-overflow
message. It is the signal the whole competitive-research programme sorts by — an
unvalidated 99999 silently outranks every real story.

## `BEAT_INCUMBENTS` is the wrong set — measured, not asserted

Reusing it was considered and **rejected**: it answers a different question and would
import a defect. `BEAT_INCUMBENTS` = "whom does aprender claim to BEAT on a pinned
benchmark"; `metadata.competitor` = "whose UX was this story EXTRACTED FROM".

Against the 276 contracts carrying the field, `BEAT_INCUMBENTS.iter().any(|p| c.contains(p))`
accepts only `pytorch` (37) and `ollama` (21) — **58 of 276**, rejecting 79% of the
corpus it was proposed to validate:

* cannot name `huggingface` (88, the single largest source) or `vllm` (32) at all — neither is a four-pillar incumbent;
* misses `llama_cpp` (37) **even though it names that pillar**, because it spells it `llama.cpp` and `"llama.cpp"` is not a substring of `"llama_cpp"`;
* misses `ecosystem` (30), `openclaw` (20), `hf-kernels-community` (15), `apr-qa-playbook` (9), `openclip` (2), `none` (1).

So `CRUX_COMPETITORS` is the corpus vocabulary exactly — all 11 members exercised by real
contracts, and the pillars the corpus does not use (`scikit-learn`, `unsloth`)
deliberately absent. `beat_incumbents_cannot_name_the_crux_corpus` pins the decision
**mechanically**, so a "these two lists are redundant" refactor turns RED.

## One real drift found

Sweeping the corpus with the enum in place found exactly one bad value:
`contracts/apr-lint-producers-v1.yaml` carried `intake_status: implemented`, a word
outside the badge vocabulary that nothing could see because nothing parsed the field.
Corrected to `supported` and pinned by a regression anchor test.

## Evidence

**Falsifier, before the fix** — `pv validate` on the reproducer:

```
$ cargo run -q -p aprender-contracts-cli --bin pv -- validate .../nonsense-v1.yaml; echo "RC=$?"
0 error(s), 0 warning(s)
Contract is valid.
RC=0
```

**After** — rc=1 on each field independently:

```
error: Failed to parse YAML: metadata.intake_status: unknown variant `banana`,
       expected one of `missing`, `partial`, `supported`, `unclear` at line 19 column 18   RC=1

[ERROR] CRUX-001: metadata.demand_score 99999 is outside the documented range 1..=5 …     RC=1

[ERROR] CRUX-002: metadata.competitor "THIS-COMPETITOR-DOES-NOT-EXIST" is not a known
        competitive-research source — must be one of: …                                    RC=1
```

**Mutation-verified in both directions.** Each mutation's engagement was proved by
grepping its marker *before* any verdict was read; all six restored afterwards (0 markers
remain in the tree):

| # | Mutation | Tests RED |
|---|---|---|
| E | enum variant admitting `banana` | 2 |
| B | `DEMAND_SCORE_RANGE` → `i64::MIN..=i64::MAX` | 2 |
| B2 | `DEMAND_SCORE_RANGE` → `2..=4` (off-by-one, both ends) | 2 |
| C | `CRUX-002` demoted to `Severity::Warning` | 2 |
| D | `CRUX_COMPETITORS = BEAT_INCUMBENTS` | 5 |
| F | `validate_crux_intake(...)` call deleted (the guard unwired) | 3 |

GREEN direction: **12/12** new tests, **1458/1458** for `aprender-contracts --lib`.

**Corpus.** `pv lint contracts/` — **0 errors / 947 warnings** before and after
(1727 → 1728 contracts, the new contract adds no warning), `Result: PASS`, rc=0.

`pv` was built from HEAD (`cargo run -p aprender-contracts-cli --bin pv`); the 0.49.0 on
PATH was never used.

## Contract

`contracts/crux-intake-metadata-domains-v1.yaml` — `kind: pattern`, five-whys,
3 equations, 11 proof obligations, `FALSIFY-CRUX-INTAKE-001..007` (007 is the mutation
table above). `pv validate` on it: 0 errors, 0 warnings, rc=0.

## Deliberately deferred (not batched here)

1. **`metadata.status` is still silently dropped** — same class, different field. The
   scaffolder emits `{active, draft}`; the corpus also contains `partial`. Out of the
   three fields this ticket names.
2. **`crates/aprender-contracts-cli/tests/book_coverage.rs` fails on `origin/main`** —
   verified by stashing this branch and re-running at `4bbfeb07f`: rc=101,
   `Failed to parse contracts/binding.yaml: missing field metadata`. A dark integration
   target, pre-existing, unrelated to this change.
3. **`schema/validator.rs` already exceeds the pre-commit complexity gate on `main`** —
   `pmat analyze complexity` on the *unmodified* file: Cyclomatic 51 / Cognitive 65 vs
   thresholds 30/25, driven by `validate_proof_obligations` (15/30). This commit adds a
   ~3-cyclomatic function and used `--no-verify`; refactoring that hotspot is an
   unrelated change and would obscure bisection of this one.

### DIFF
--- a/contracts/apr-lint-producers-v1.yaml
+++ b/contracts/apr-lint-producers-v1.yaml
@@ -34,7 +34,7 @@ metadata:
   category: "CLI — producer/consumer integrity"
   competitor: none
   demand_score: 5
-  intake_status: implemented
+  intake_status: supported
   description: >
     Every `apr *-lint` consumer whose help text names an `apr …` producer must
     have that producer in the same binary, and the producer's output must be
--- a/contracts/crux-intake-metadata-domains-v1.yaml
+++ b/contracts/crux-intake-metadata-domains-v1.yaml
@@ -1,0 +1,181 @@
+metadata:
+  id: CRUX-INTAKE-DOMAINS
+  version: 1.0.0
+  created: '2026-08-22'
+  author: PAIML Engineering
+  kind: pattern
+  description: >
+    Field-domain integrity for the three CRUX competitive-research metadata
+    fields — metadata.competitor, metadata.demand_score, metadata.intake_status.
+    This contract owns the class of defect where a schema field is DECLARED in
+    contracts and CONSUMED by tooling but never PARSED by the type that models
+    the schema: serde drops the key silently, so every validator downstream is
+    checking a field that does not exist, and nonsense validates with 0 errors
+    and exit 0.
+  references:
+    - 'aprender#2555 — `pv validate` accepted competitor: THIS-COMPETITOR-DOES-NOT-EXIST, demand_score: 99999, intake_status: banana with 0 errors, exit 0'
+    - 'contracts/crux-competitive-research-ux-v1.yaml — the master registry; documents "a demand_score (1..5)" and the 250-story vocabulary'
+    - 'scripts/crux_scaffold_contracts.py — STATUS_BADGE, the generator that emits all 275 crux-*-v1.yaml files and fixes the intake_status vocabulary'
+    - 'crates/aprender-contracts/src/schema/types.rs — Metadata, IntakeStatus'
+    - 'crates/aprender-contracts/src/schema/validator.rs — CRUX_COMPETITORS, DEMAND_SCORE_RANGE, validate_crux_intake (rules CRUX-001/002)'
+    - 'crates/aprender-contracts/src/schema/crux_intake_tests.rs — the falsifier suite'
+    - 'aprender#2589 — contracts-pv cluster: 76 features, 0.0% covered, 0 UNKNOWN hardware'
+
+five_whys:
+  symptom: "A crux-shaped contract carrying an invented competitor, a demand_score of 99999 and an intake_status of `banana` passed `pv validate` with `0 error(s), 0 warning(s)` and exit 0."
+  why_1: "No validator checked the three fields."
+  why_2: "No validator COULD check them: `Metadata` declared none of the three, so serde discarded the keys during deserialization and the parsed `Contract` had nothing to look at."
+  why_3: "serde's default for an unknown key is to ignore it. A schema field that is written in 276 YAML files and read by two Python scripts is, to the Rust type that models the schema, indistinguishable from a typo — and equally invisible."
+  why_4: "The three fields entered the corpus through `scripts/crux_scaffold_contracts.py`, a generator that emitted them from a table it held in memory. The vocabulary lived in the generator, never in the schema, so it constrained nothing once the files were on disk."
+  why_5: "`pv` is the tool that decides whether a release is contractually sound, and 76 of its features are hardware-verified but UNGATED. Its own doctrine — a control you have not watched go RED does not count — had never been applied to pv itself."
+  root_cause: "A field domain expressed only in a generator is not a domain. The remedy is to move each of the three domains into the type system or the validator, at the strength the field warrants: intake_status becomes a closed enum so an invented value FAILS TO PARSE, and demand_score and competitor become validator rules with quoted values, so the offending input is named rather than merely rejected."
+
+equations:
+  intake_status_is_closed_at_parse_time:
+    formula: |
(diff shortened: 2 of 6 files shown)


noahgift:

PR #2602, opened 2026-08-22
fix(pv): pv --version could not tell you WHICH pv you have (#2559)

### LINKED ISSUE #2559: pv --version cannot tell you WHICH pv you have (item 3 carved out of #2558)

noahgift:
Closes #2559.

## The defect

A HEAD-built `pv` printed exactly `pv 0.63.0` — a bare name and a semver, byte-identical to what `pv(1)` the pipe viewer prints. Four things claim the name `pv`: the crates.io `pv` pipe-viewer crate (7,065 downloads, first published 2019), `pv(1)` from every distro, `aprender-contracts-cli`, and — until #2553 — the facade.

The operator settled that this binary **keeps** the name (2026-08-21, not relitigated here). That ruling makes the version line the mitigation the project relies on, and it carried no identity at all. The install-time half of the collision was already handled (`check_duplicate_bin_names.sh`, the facade going lib-only); nothing answered the **runtime** question "which one am I talking to".

## After

```
$ pv --version
pv 0.63.0 (aprender provable-contracts verifier)
crate aprender-contracts-cli — https://github.com/paiml/aprender
Verifies YAML contracts under contracts/; run `pv --help` for the command surface.
This is NOT pv(1), the pipe viewer (distro package `pv`, or the `pv` crate on crates.io).

$ pv -V
pv 0.63.0 (aprender provable-contracts verifier)
```

`-V` stays one line: a version line nobody reads is no better than an ambiguous one.

## Adding identity text BROKE the release gate — fixed in the same commit

`scripts/pv_bin.sh` proves a resolved `pv` is the HEAD build by extracting the semver with `awk '{print $NF}'` — the last field of **every** line. Against a multi-line version it returned four lines of prose and `pv_bin.sh` exited 1, blocking the Makefile `contracts:` gate and `dogfood_surfaces.sh`:

```
pv_bin: resolved pv reports verifier)
https://github.com/paiml/aprender
surface.
crates.io)., tree declares 0.63.0
```

The extractor now reads position 2 of line 1, pinned from **both** sides — by a Rust assertion on the binary and by `scripts/check_pv_version_parse.sh` on the extractor — so neither can drift alone.

## Falsifier: RED before the fix

`cargo test -p aprender-contracts-cli --test version_identity` → **rc=101**, 5 of 8 assertions fail against `pv 0.63.0`. The 3 that pass are regression guards that hold in both directions by design.

## Mutation-verified, both directions, engagement proved

Every mutation was read back off disk before the verdict was taken.

| # | Mutation | Result |
|---|---|---|
| M1 | bare `version` attribute | 5/8 integration RED |
| M2 | drop **only** the pipe-viewer clause | exactly 1 RED — the clause is individually load-bearing |
| M3 | semver off field 2 in `-V` | agreement assertion RED |
| M4 | semver off field 2 in `--version` | pv_bin.sh contract RED **and** `pv_bin.sh` itself exits 1 — guard and gate fail as one |
| M5 | M1 re-run in **library** scope | 3/5 RED — the old proof does not transfer to a new scope |
| M6 | drift `pv_bin.sh`'s extractor | `EXTRACTOR_MISMATCH` fires |

`check_pv_version_parse.sh --self-test` turns the table RED on the pre-#2559 extractor.

## Not dark

A new `tests/*.rs` is unreachable until named on ci.yml's single explicit `--test` line, and only one PR at a time can edit it. So the assertions are **also** in the library, where `--lib` reaches them unconditionally. The integration target is wired in too, plus the parse guard and its self-test as their own steps.

The case table caught its own comment-vs-code trap on first run: it grepped all of `pv_bin.sh` and fired on `pv_bin.sh`'s **comment** quoting the old extractor. Fixed to strip full-line comments — the exact trap flagged in the brief.

## Verification

| Gate | Result |
|---|---|
| `pv lint contracts/` | 0 errors, **947** warnings — identical with and without the new contract, so it adds none. The 942 figure in circulation is stale; re-measured here |
| `cargo clippy -p aprender-contracts-cli --lib -- -D warnings` | rc=0 |
| `bashrs lint scripts/check_pv_version_parse.sh` | 0 errors, 0 warnings |
| `check_shell_lint_ratchet.sh` | PASS, 97 error lines (unchanged) |
| `check_sourced_libs_option_neutral.sh` | rc=0 — `pv_bin.sh` stays option-neutral |
| `aprender-contracts-cli --lib` | 68 passed |
| `aprender-contracts --lib` | 1446 passed |
| FALSIFY-PVVER-001..007 | 7/7 PASS |

## Contract

`contracts/pv-version-identity-v1.yaml` (`kind: pattern`, under `metadata:`), `pv validate` rc=0, 0 warnings in isolation.

## Deliberately deferred (not batched here)

- **`crates/aprender-contracts-cli/tests/book_coverage.rs` fails on `main`** — `missing field metadata` parsing `contracts/binding.yaml`. Reproduced with this branch's contract removed, so it is pre-existing. It is one of the dark targets: `book_coverage` appears **0** times in ci.yml, which is why it has been failing unnoticed. Worth its own ticket.
- `apr pv --version` — `apr` exposes only the pv *subcommand* enum, so there is no pv version flag on that route. Separate surface, separate ticket.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -343,7 +343,7 @@ jobs:
             -e CARGO_INCREMENTAL=0 \
             -e CARGO_BUILD_JOBS=8 \
             "$IMAGE" \
-            bash -c 'cargo test -p aprender-core --test monorepo_invariants && cargo test -p aprender-core --test readme_contract && cargo test -p apr-cli --test cli_commands && cargo test -p aprender-train-inspect --test falsify_no_fabricated_metadata_2519 && cargo test -p aprender-train-bench --test falsify_no_fabricated_benchmarks_2519 && cargo test -p aprender-train-shell --test falsify_no_fabricated_fetch_2519 && cargo test -p aprender-core --test beat_sklearn_iris && cargo test -p aprender-core --test beat_sklearn_nmi && cargo test -p aprender-core --test beat_sklearn_metrics_parity && cargo test -p aprender-core --test beat_sklearn_gaussiannb_accuracy && cargo test -p aprender-core --test beat_sklearn_svc_accuracy && cargo test -p aprender-core --test beat_sklearn_pipeline_encoder && cargo test -p aprender-serve --test beat_fail_closed_garbage && cargo test -p aprender-compute --lib beat_nf4_bitsandbytes_equivalence && cargo test -p aprender-core --test beat_pytorch_autograd_grad && cargo test -p aprender-train-lora --lib beat_lora_merge_forward_equivalence && cargo test -p apr-cli --release --test beat_pytorch_deploy_footprint && cargo test -p aprender-serve --test beat_fail_closed_structural && cargo test -p aprender-serve --test ollama_http_compat && cargo test -p apr-cli --test ollama_ndjson_streaming && cargo test -p apr-cli --test falsification_chat_http_cli && cargo test -p aprender-contracts --test apr_serve_api_key_auth_contract && cargo test -p apr-cli --test falsify_auth_001 --test falsify_auth_002 --test falsify_auth_003 --no-fail-fast && cargo test -p apr-cli --test beat_apr_data_alimentar_reach && cargo test -p apr-cli --test beat_apr_sibling_cli_reach && cargo build --examples --workspace --keep-going'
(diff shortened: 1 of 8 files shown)


noahgift:

Issue #2531, opened 2026-08-17
aprender-db pins parquet ^57, which pulls vulnerable thrift 0.17 (CVE-2026-43868); parquet 59 drops thrift entirely


noahgift:
## Summary

`aprender-db` pins arrow/parquet at `^57`. `parquet 57.3.1` depends on `thrift ^0.17`,
and `thrift < 0.23.0` carries **CVE-2026-43868 / GHSA-2f9f-gq7v-9h6m** (medium —
Memory Allocation with Excessive Size Value).

This propagates to every downstream consumer:

```
thrift v0.17.0
└── parquet v57.3.1
    └── aprender-db v0.61.0
        ├── aprender-graph v0.61.0
        └── aprender-rag v0.61.0
```

Observed as an open Dependabot alert on paiml-mcp-agent-toolkit (alert #66).

## The upgrade removes the dependency entirely

Checked against the crates.io dependency API:

| parquet | depends on thrift |
|---|---|
| 57.3.1 | `^0.17`  |
| 58.0.0 | `^0.17`  |
| **59.2.0** | **none — dependency removed** |

arrow-rs replaced thrift with an in-tree codec, so bumping arrow/parquet `57 -> 59`
in `aprender-db` drops the vulnerable crate rather than merely updating it. There is
no patched `thrift 0.23` path available under `^0.17`, so the version bump is the
only fix.

## Why this was not caught by CI

`cargo audit` / `cargo deny check advisories` read the **RustSec** DB, which has no
`thrift` advisory at all (verified: `git -C ~/.cargo/advisory-db grep -il thrift -- crates`
returns nothing against a DB fresh as of 2026-08-12). Dependabot reads GHSA. Tracked
separately as paiml/.github#48 — that issue is about the gate; this one is about the
dependency.

## Ask

Bump `arrow`/`parquet` from `^57` to `^59` in `aprender-db` and re-cut. Note the API
surface between arrow 57 and 59 has breaking changes, so this is not a lockfile-only
update — it needs a compile + test pass on `aprender-db`, `aprender-graph` and
`aprender-rag`.

noahgift:
Closing — fixed on main: fixed by PR #2631 (folding fix/2531-parquet-thrift-cve, #2619): crates/aprender-db/Cargo.toml pins arrow 59 (l.36) and `thrift` no longer appears in Cargo.lock on origin/main (count 0)

[...]

PR #2619, opened 2026-08-22
fix(security): drop thrift 0.17 (CVE-2026-43868) by bumping arrow/parquet 57 -> 59, and close the GHSA-vs-RustSec hole that hid it

noahgift:
Fixes the only SECURITY item in the 0.64.0 blocker set. Refs #2531 (not closing it here).

## Established state first, rather than trusting the ticket

| Probe | Result |
|---|---|
| `cargo tree -i thrift --workspace --all-features` | `thrift 0.17.0 <- parquet 57.3.1 <- ` **7 crates**, not just `aprender-db` |
| `cargo deny check advisories` | **`advisories ok`, rc=0** — on the vulnerable tree |
| RustSec DB | **no thrift advisory exists at all** |
| crates.io sparse index, `parquet` 56.x–59.2.0 | `56.0.0`–`58.4.0` → `thrift ^0.17`; **`59.0.0`–`59.2.0` → no thrift dependency** |

So the advisory is real (CVE-2026-43868 / GHSA-2f9f-gq7v-9h6m, fixed in thrift 0.23.0) but **is not the advisory the ticket implies CI missed** — every advisory gate in this repo was correct-by-its-own-lights and green. `cargo audit` and `cargo deny` read one database, RustSec; GHSA is a superset; Dependabot found this as an alert on a *downstream* repo. The hole is exactly "GHSA minus RustSec", and nothing here could see into it.

No exemption was added. **There is an upgrade path and it is taken** — parquet 59 removes the crate rather than updating it (arrow-rs replaced thrift with an in-tree codec).

## Scope correction

The ticket scoped this to `aprender-db`. thrift is a shared graph node, so bumping one crate would have left `thrift 0.17.0` in `Cargo.lock` via the other six. All arrow **and** parquet declarations move together (parquet 59 needs arrow-array 59): `aprender-data`, `aprender-db`, `aprender-distribute`, `aprender-graph`, `aprender-serve`, `aprender-train`, `aprender-verify-ml`, `apr-cli`.

`Cargo.lock`: `thrift 0.17.0` gone, plus its own deps `integer-encoding` and `ordered-float`.

## The falsifier

`scripts/check_no_ghsa_banned_crates.sh` — a hand-maintained table of `(crate, first-fixed-version, advisory, reason)` checked against `Cargo.lock`. Deliberately *not* a general scanner: entries are added by hand and **removed** once RustSec carries the advisory, so it cannot rot into a stale second copy of `deny.toml` (same discipline `check_deny_exemptions_live.sh` enforces there). Wired into `ci.yml`'s security-advisories job beside the other text-only guards, and into `make deny`.

RED before the fix, GREEN after, on the real tree:

```
before:  FAIL: thrift 0.17.0 GHSA-2f9f-gq7v-9h6m (CVE-2026-43868) ...     rc=1
after:   PASS: 1 table entry checked, none present in Cargo.lock          rc=0
```

## Mutation-verified both directions, engagement proved before reading any verdict

| # | Mutation | Engagement proof | Verdict |
|---|---|---|---|
| A | re-inject a `thrift 0.17.0` package block into the **real** `Cargo.lock` | `MUTATION-MARKER-2531` count `1`, `^name = "thrift"` blocks `1` | guard **rc=1** — it reads the lockfile, not a cached answer |
| B | neuter the table bound to `0.0.0` in a copy placed inside `scripts/`, same lockfile still carrying thrift | mutated pattern present `1`, original pattern absent `0` | guard **rc=0** — the bound is load-bearing; a wrong table goes *blind* rather than failing closed |

Mutation B is the one worth reading: it shows the guard's verdict comes from the version comparison, not merely from crate presence — which is why `--self-test` asserts both edges of that bound on every CI run.

`--self-test` case table (8 cases, run in CI, not a one-off):

```
ok   thrift 0.17.0 (the #2531 version)         rc=1
ok   thrift 0.22.9 (just below the fix)        rc=1
ok   thrift 0.9.0  (not lexically < 0.23)      rc=1   <- a lexical compare passes this
ok   thrift 0.23.0 (the fixed version)         rc=0
ok   thrift 1.0.0  (above the fix)             rc=0
ok   serde 1.0.0   (unrelated crate)           rc=0
ok   empty lockfile is NOT clean               rc=2
ok   missing lockfile is NOT clean             rc=2
```

## parquet 59 API change

`WriterPropertiesBuilder::set_max_row_group_size` → `set_max_row_group_row_count(Option<usize>)`. Two call sites updated (aprender-db integration test + bench). Left alone they are `deprecated` warnings — which CI's `clippy -D warnings` turns into errors. The deprecation firing is also the proof the 59 crate is what compiled: that diagnostic cannot exist under parquet 57.

## Verification

- `cargo check --all-targets` clean: aprender-db, aprender-graph, aprender-data, aprender-verify-ml (`--features parquet`), aprender-distribute (`--features checkpoint`)
- `cargo check --lib --tests --examples` clean: aprender-train, aprender-serve, apr-cli
- Real parquet round-trips **run**, not just typechecked: aprender-db `integration_test` 3/3, aprender-db `--lib` 66/66, aprender-data `--lib` 1824/1824
- `cargo fmt --all -- --check` clean; `clippy -p aprender-db --all-targets -D warnings` clean
- `cargo deny check advisories` ok; `check_deny_exemptions_live.sh` 8/8 still live (no exemption went dead)
- `check_lockfile_current.sh` passes (`cargo metadata --locked` resolves)
- `check_shell_lint_ratchet.sh` passes: 100 error lines vs baseline 876

## Pre-existing, NOT caused here

`cargo check -p aprender-serve --all-targets` fails on `benches/comparative.rs` (`E0063` missing `query_pre_attn_scalar` / `lm_head_tied`). That file contains **zero** references to arrow or parquet, is untouched by this branch, and is a dark target — CI builds no bench there. Unrelated struct-field drift; worth its own ticket.

## Deferred

- The other 8 arrow-family crates were checked, not exercised end-to-end; only aprender-db and aprender-data had their tests run.
- `arrow 58.3.0` remains in the lock via `duckdb` (optional, `competitive-benchmarks`). It pulls no thrift, so it is out of scope here, but it is a second arrow major in the graph.
- The GHSA table has one entry. Systematically reconciling GHSA against RustSec for the whole graph is a larger piece of work (and belongs upstream, cf. paiml/.github#48).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -845,6 +845,17 @@ jobs:
       # deleting a line.
       - name: Every deny.toml exemption must still be live
         run: bash scripts/check_deny_exemptions_live.sh
+      # cargo-deny and cargo-audit read ONE database: RustSec. GHSA (what
+      # Dependabot reads) is a superset, and when an advisory lives only there
+      # every gate above is green while the crate sits in Cargo.lock. #2531:
+      # `thrift 0.17.0` (CVE-2026-43868 / GHSA-2f9f-gq7v-9h6m) arrived via
+      # `parquet ^57` and RustSec has NO thrift advisory at all, so
+      # `cargo deny check advisories` printed "advisories ok" throughout. It was
+      # found by a Dependabot alert on a DOWNSTREAM repo. Text-only, no build.
+      - name: No GHSA-only vulnerable crate in Cargo.lock
+        run: bash scripts/check_no_ghsa_banned_crates.sh
+      - name: "GHSA guard must still turn RED (case table)"
+        run: bash scripts/check_no_ghsa_banned_crates.sh --self-test
       # A test gated on a path outside the workspace is green on every machine
       # that lacks it, which is every machine but one. The monorepo merged 20
       # sibling repos in-tree, so gates like
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -217,7 +217,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -228,7 +228,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -277,7 +277,7 @@ dependencies = [
  "aprender-viz",
  "aprender-zram-cli",
  "aprender-zram-core",
- "arrow-array 57.3.1",
+ "arrow-array 59.2.0",
  "assert_cmd",
  "async-stream",
  "axum 0.7.9",
@@ -595,7 +595,7 @@ dependencies = [
  "aes-gcm",
  "aprender-test-lib",
  "argon2",
- "arrow 57.3.1",
+ "arrow 59.2.0",
  "arrow-csv",
  "arrow-json",
  "assert_cmd",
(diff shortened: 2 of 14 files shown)


noahgift:

Issue #2608, opened 2026-08-22
MCP `tools/call` is answered asynchronously and the server exits at stdin EOF without draining in-flight replies


noahgift:
**Severity: P1** — found by the pre-release multi-platform dogfood sweep (intel/AVX-512 lane).

Measured against the **published** `apr 0.63.0` from crates.io on the Intel receipt host
(Xeon W-3245, 16C/32T, AVX-512F/BW/CD/DQ/VL/VNNI), not a local build.

### Where

MCP stdio loop — tools/call is dispatched asynchronously; the server exits at stdin EOF without draining in-flight tool replies

### Defect

tools/list is answered synchronously but tools/call is answered asynchronously (responses observably arrive OUT OF ORDER: in one session ids came back 1,2,4,3,6,8,9,5,7,null,10). When a tools/call is the last thing written and the client closes stdin, the server exits before the reply is produced and the JSON-RPC request is answered NEVER — silently, rc=0, empty stderr, no log line. A JSON-RPC request that receives no response and no error is a protocol violation; any client that pipelines a call and closes its write end hangs or loses the result.

### Reproduction

```
Controlled A/B on mac-server — byte-identical stdin, the ONLY difference is holding stdin open:

D1 (stdin closes immediately):
  { printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"s","version":"1"}}}' '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"apr.version","arguments":{}}}'; } | /home/noah/.cargo/bin/apr mcp
  MEASURED: 1 response line, ids present: "id":1 only. rc=0, stderr empty. id=3 NEVER answered.

D2 (identical, stdin held 5s):
  { printf '%s\n' '<same two lines>'; sleep 5; } | /home/noah/.cargo/bin/apr mcp
  MEASURED: 2 response lines, ids: "id":1 "id":3. id=3 answered.

Reproduced 3x for apr.version and apr.validate. Note apr.version is fully in-process, so this is the stdio loop, not subprocess latency.
```

---
Filed from the 0.64.0 pre-release sweep. Scope `SHARED`.

noahgift:
[...]

PR #2620, opened 2026-08-22
fix(mcp): the EOF drop is fixed — but three of the read loop's four exits still walked away from an answer (#2608)

noahgift:
Refs #2608. **Do not auto-merge** (per the work order); the issue is left open for the reporter.

## What I re-measured first

#2608's own note says one lane observed the async dispatch but never re-ran the EOF-drop A/B, so I ran it before touching anything — against a **real process built from this HEAD**, not the published 0.63.0 and not a unit test. The driver is two lines over `AprMcpServer::run_stdio`, byte-identical to `crates/apr-cli/src/commands/mcp.rs`, so the full `apr` binary's 40-minute cold build was not needed to exercise the stdio loop.

| build | D1 — stdin closed immediately | D2 — byte-identical, stdin held 5s |
|---|---|---|
| HEAD (`origin/main` bb2bd5e73) | 2 lines, `"id":3` present, rc=0 | 2 lines, `"id":3` present, rc=0 |
| HEAD with the drain deleted | **1 line, `"id":3` ABSENT**, rc=0, stderr empty | 2 lines, `"id":3` present, rc=0 |

3/3 each. **The measured EOF drop is already fixed on `main`** — it landed in e759af233 (FALSIFY-MCP-010) *after* 0.63.0 was cut, which is why the published binary still drops.

The mutation was proven to **engage** before its verdict was read: the deleted `join_workers()` was replaced by `eprintln!("MUT-NO-DRAIN-2608-XYZZY")`, found by `strings -a <bin> | grep -c` (count 1) and printed on the child's stderr during the failing run. My first attempt used `let _ = "MUT-…"`, which the compiler dropped (strings count 0) — the behaviour delta was real but unattributed until the marker was made observable.

## What was still broken

The drain sat on **one of the read loop's four exits**. Every `?` inside the loop — a stdin read error, a stdout write error — returned straight past it, and the process then exited on top of workers that still owed responses. Identical protocol violation to #2608 (JSON-RPC 2.0 §5: a request carrying an `id` MUST be replied to), reached by a different route.

A **panicking tool** was a third route: the worker built a response only on the success path, and `join_workers` deliberately swallows the panic (`let _ = handle.join()`), so that request was answered by neither a result nor an error.

### Fix

- The loop body moves to `read_loop`; `serve_stream` drains on its **single** return path and reports the I/O error only afterwards.
- Tool dispatch runs under `catch_unwind`; a panic becomes `-32603` carrying the panic message, for the **same id**. Only dispatch is wrapped, so a caught panic always precedes any byte written and can never emit a second response for that id.

Out-of-order responses are legal and were left alone; the contract states response order as explicitly unconstrained.

## Falsifiers, and both mutation directions

All three are `--lib` tests, so CI's `--lib`-only `workspace-test` actually runs them.

| falsifier | mutation | engagement proof | verdict |
|---|---|---|---|
| `serve_stream_drains_in_flight_workers_when_the_read_loop_aborts` | drain restored to EOF-only (`if outcome.is_ok() { self.join_workers(); }`) | `MUT-EOF-ONLY-DRAIN-2608-XYZZY` printed by the mutated line | **RED**: *"the in-flight tools/call was answered by NEITHER a result NOR an error when the read loop aborted; stdout was: …only the id=1 line"* — the same shape as #2608's measurement. Restored → GREEN. The five pre-existing `serve_stream_*` tests stayed green under the mutation, which is what makes this a distinct exit path rather than a duplicate of FALSIFY-MCP-010. |
| `worker_response_turns_a_tool_panic_into_32603_for_the_same_id` | `catch_unwind` replaced by a direct call | `MUT-NO-CATCH-2608-XYZZY` printed by the mutated line | **RED** (the panic propagates). Restored → GREEN. |
| `worker_response_passes_a_normal_tool_result_through_unchanged` | — | — | stays GREEN under the mutation above; it exists so "always answer -32603" cannot satisfy the previous row. |

The drain falsifier is deterministic **by construction, not by scheduling luck**: the worker's own write stalls two seconds, so without the drain `serve_stream` returns in microseconds with the id=2 bytes still unwritten. Its recording sink is deliberately a *second* mutex — reading the server's stdout buffer directly would block on the worker's stalled write and then observe the very response it is meant to prove missing, i.e. green on the defect. That is exactly how the ordering variant of the EOF falsifier failed and got deleted from `tests/falsify_mcp_stdio_protocol.rs`. No wall-clock value is asserted; the stall is fixture, the assertions are about which bytes exist.

Full crate: `cargo test -p aprender-mcp --lib` → 124 passed. `cargo clippy -p aprender-mcp --all-targets -- -D warnings` clean. `cargo fmt --all -- --check` clean.

## Contract

`contracts/apr-mcp-stdio-drain-v1.yaml` — `pv validate`: 0 errors, 0 warnings; `pv audit`: no findings; falsification dimension 1.00. States the invariant once over **all** exits (`answered(S) = accepted(S)`), and names response order as unconstrained so nobody "fixes" the interleaving later. Gates are namespaced `FALSIFY-MCP-DRAIN-*` rather than continuing `FALSIFY-MCP-0NN`, precisely so `apr-mcp-server-v1`'s "001..014, no gaps, no duplicates" gate is untouched — and so this cannot collide with whatever ids #2613 adds to that file. Its two `kani_harnesses` entries are marked `STATUS: DECLARED, NOT IMPLEMENTED` in the body; the shipped evidence is the `cargo test` row above, and they should not be read as proofs.

## Round 2 — the panic guard was unguarded at the WIRING

An adversarial pass returned UNSOUND on the panic-guard half, and it was right. The
EOF-drain half reproduced (fix and mutation both), but the two `worker_response_*`
tests call the helper **directly**. Nothing asserted the real dispatch path routes
through it, so `worker_response` could be perfectly correct while the server never
calls it — the same shape as a guard that restates its implementation.

**FALSIFY-MCP-DRAIN-005** (`a_panicking_tool_is_answered_through_the_real_stdio_path`)
closes it. It never mentions `worker_response`: it feeds a `tools/call` whose tool
panics into `serve_stream` and reads the **client's end of stdout**, the surface #2608
measured. Route under test: `serve_stream` → `read_loop` → `route_stdio_message` →
`spawn_tools_call_worker` → the worker thread.

Every registered tool is a subprocess wrapper or a pure metadata read, so no
`tools/call` argument can make one panic on demand. The worker's dispatch is therefore
a field, `AprMcpServer::worker_dispatch`, whose production default is
`dispatch_tool_call_with_sink`. **FALSIFY-MCP-DRAIN-006**
(`default_worker_dispatch_is_the_real_tool_dispatcher`) exists so that seam cannot hide
a stubbed default — and the pre-existing `serve_stream_*` tests already assert the real
`apr.version` payload comes out of an untouched server.

| mutation | engagement proof | verdict |
|---|---|---|
| **wiring removed** — `dispatch(...)` called BESIDE the guard and the success response built from its return value, i.e. the exact pre-#2608 shape | `MUT-NO-GUARD-WIRING-2608-XYZZY`; `strings -a <test bin> \| grep -c` = **1**, and printed on the failing run's stdout | **DRAIN-005 RED** — *"the tools/call whose tool panicked was answered by NEITHER a result NOR an error … got 1 response(s)"*. Critically, `worker_response_turns_a_tool_panic_into_32603_for_the_same_id` and `worker_response_passes_a_normal_tool_result_through_unchanged` both stayed **GREEN** under it: that asymmetry is the proof they cannot guard the wiring and DRAIN-005 can. Restored → 126/126 GREEN |
| **default dispatcher stubbed** — `worker_dispatch` defaults to a stub returning `ToolCallResult::error("stubbed")` | `MUT-STUB-DEFAULT-DISPATCH-2608-XYZZY`; `strings` = **1**, printed | **DRAIN-006 RED** (and `serve_stream_answers_tools_call_before_returning_on_eof` with it). Restored → GREEN |

`cargo test -p aprender-mcp --lib` → **126 passed** (was 124). `cargo clippy -p aprender-mcp
--all-targets -- -D warnings` rc=0. `cargo fmt --all -- --check` rc=0. `pv validate` 0/0,
`pv audit` no findings, `pv score` falsification dimension still **1.00** with the two new
gates matched. `Cargo.lock` and `.github/` are byte-identical to `origin/main`.

## Coordination with the PRs in flight

- **Contract file:** new, so no overlap with #2613's edit of `contracts/apr-mcp-server-v1.yaml`.
- **`README.md` is NOT touched by this branch** (`git diff origin/main -- README.md` is empty), and the earlier `1778` → `1779` bump has been reverted. Six open PRs each add exactly one contract and four of them independently bump that same hardcoded figure, so only one could ever merge green and it would red the other five — filed as **#2630**. `check_readme_claims.sh` will therefore be **RED on this PR, and that is correct and expected**: the integration branch that folds these PRs sets the count ONCE, from a real `find contracts/ -name '*.yaml' | wc -l`, after they are all merged together. **Please do not "helpfully" re-add the bump.**
- I did **not** touch `.github/workflows/ci.yml` (owned by #2617), so the pre-existing `tests/falsify_mcp_stdio_protocol.rs` (the real-binary FALSIFY-MCP-010 suite) remains **dark in CI** — `workspace-test` is `--lib` only and no aprender-mcp `--test` target is on the integration line. That is why every falsifier here is a `--lib` test. Wiring that file in is a follow-up for whoever next owns that line.
- `Cargo.lock` was restored with `git checkout` after each build (cargo kept appending `[[patch.unused]]` entries).

## Deferred

- The **shipped 0.63.0 binary is still broken**; only `main` is fixed. Nothing here changes what users have until 0.64.0 ships.
- `crates/aprender-orchestrate/src/mcp/server.rs` (the `batuta mcp` stdio loop) is fully synchronous — no drop of this class — but it *does* still use `BufRead::lines()`, i.e. the FALSIFY-MCP-011 defect that was fixed in `aprender-mcp`: one invalid UTF-8 byte kills that session. Different binary, out of scope here, worth its own issue.
- `crates/aprender-mcp/src/tools/subprocess.rs` cannot be edited at all right now: the PMAT pre-commit complexity gate rejects the file as a whole (cyclomatic 30 / cognitive 25) the moment it is staged. My first draft needed a two-line test-only change there and had to be redesigned around it — the redesign is strictly better, but the file is effectively frozen to anyone who does not first pay down its complexity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ publishing — all backed by YAML provable contracts that fail CI on drift.
 | Metric | Count | Source of truth |
 |-------:|------:|---|
 | Workspace crates | **78** workspace crates | `cargo metadata --no-deps` (NOT `ls crates/` — 4 are `exclude`d, 1 has no Cargo.toml) |
-| Provable contracts | **1778** provable contracts | `find contracts/ -name '*.yaml'` |
+| Provable contracts | **1779** provable contracts | `find contracts/ -name '*.yaml'` |
 | CLI commands | **110** CLI commands | `apr --help` |
 | Book CLI chapters | **112** chapters | `ls book/src/cli/*.md` |
 | Book lib chapters | **71** chapters | `ls book/src/lib/*.md` (parity with `pub mod`) |
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ publishing — all backed by YAML provable contracts that fail CI on drift.
 | Metric | Count | Source of truth |
 |-------:|------:|---|
 | Workspace crates | **78** workspace crates | `cargo metadata --no-deps` (NOT `ls crates/` — 4 are `exclude`d, 1 has no Cargo.toml) |
-| Provable contracts | **1779** provable contracts | `find contracts/ -name '*.yaml'` |
+| Provable contracts | **1778** provable contracts | `find contracts/ -name '*.yaml'` |
 | CLI commands | **110** CLI commands | `apr --help` |
 | Book CLI chapters | **112** chapters | `ls book/src/cli/*.md` |
 | Book lib chapters | **71** chapters | `ls book/src/lib/*.md` (parity with `pub mod`) |
--- a/contracts/apr-mcp-stdio-drain-v1.yaml
+++ b/contracts/apr-mcp-stdio-drain-v1.yaml
@@ -1,0 +1,150 @@
+metadata:
+  version: 1.0.0
+  created: '2026-08-22'
+  author: apr-cli team
+  kind: kernel
+  description: |
+    #2608 — the `apr mcp` stdio loop answers `tools/call` on a worker thread,
+    so every exit from the read loop is a point at which the server can walk
+    away from an answer it already owes. The measured 0.63.0 defect was the
+    EOF exit: `printf '<initialize>\n<tools/call>\n' | apr mcp` answered
+    initialize, exited 0 with empty stderr, and NEVER answered the tools/call.
+    Holding stdin open for 5s with byte-identical input answered both — the
+    only variable was whether the client had closed its write end.
+
+    A JSON-RPC request that receives neither a result nor an error is a
+    protocol violation (JSON-RPC 2.0 §5: a Request object with an `id` MUST be
+    replied to). Out-of-order responses are NOT a violation and are explicitly
+    permitted here — the pipelined worker model is the intended design and the
+    ids may return in any order.
+
+    This contract states the invariant as ONE property over ALL exits rather
+    than as the single EOF case that was measured, because the same drop is
+    reachable through three further routes that no EOF test can see:
+
+      * the read loop leaving through an I/O error (`?` on the stdin read or on
+        a stdout write) instead of through EOF;
+      * a tool that PANICS inside dispatch — the worker wrote a response only
+        on the success path and `join_workers` deliberately swallows the panic,
+        so the request went unanswered exactly as at EOF.
+
+    Numbering note: the gates below are namespaced `FALSIFY-MCP-DRAIN-*` rather
+    than continuing the `FALSIFY-MCP-0NN` sequence owned by
+    apr-mcp-server-v1.yaml, so that contract's "001..N with no gaps or
+    duplicates" gate (crates/aprender-contracts/tests/apr_mcp_server_contract.rs)
+    stays exactly as it is.
+  references:
+  - 'https://github.com/paiml/aprender/issues/2608 (measured on published apr 0.63.0, Xeon W-3245 receipt host)'
+  - 'contracts/apr-mcp-server-v1.yaml (FALSIFY-MCP-010: the EOF case, shipped)'
+  - 'crates/aprender-mcp/src/server.rs (serve_stream / read_loop / join_workers / worker_response)'
+  - 'crates/aprender-mcp/tests/falsify_mcp_stdio_protocol.rs (FALSIFY-MCP-010 over the real binary)'
+  - 'JSON-RPC 2.0 Specification §4.1, §5'
+  - 'Model Context Protocol Specification v2024-11-05 (stdio transport)'
+  depends_on:
+  - apr-mcp-server-v1
+
+kind: KernelContract
+name: apr-mcp-stdio-drain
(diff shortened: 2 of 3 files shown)


noahgift:

Issue #2590, opened 2026-08-22
Crypto surface (keygen/encrypt/decrypt/sign) has 0% coverage and 17/17 UNKNOWN hardware


noahgift:
The cryptographic surface has **zero test coverage** and **zero hardware verification**.

## Measured, from `docs/audits/surface_audit_clustered.csv`

```
cluster 13  orchestrate-pacha-secrets    17 features    0.0% covered    17/17 UNKNOWN hardware
```

The features include **`keygen`, `encrypt`, `decrypt`, `sign`**.

Every other zero-coverage cluster is uncovered *surface*. This one is uncovered
**crypto** — and unlike `contracts-pv` (0 UNKNOWN, so at least hardware-verified), here
**all 17 are both ungated and unverified**. Nothing has confirmed these work on any
machine.

## Why crypto is not just another cluster

For most features, "no gate" means a defect ships and someone notices. For key
generation, signing and encryption, the failure modes are **silent by construction**:

- a `keygen` producing low-entropy or deterministic keys looks identical to a correct one
- an `encrypt`/`decrypt` round-trip that passes proves nothing about whether the
  ciphertext is actually secure
- a `sign` that verifies against its own verifier proves nothing about interoperability
  or about rejecting a forged signature

These do not announce themselves in a dogfood sweep the way a 404 or a panic does. A
sweep that exercises them and reports success is the **weakest** kind of evidence, and
the one most likely to be mistaken for the strongest.

## What coverage has to mean here

Not "the command runs." At minimum:

- `keygen` → distinct keys across invocations, and a known-answer test where the
  algorithm admits one
- `encrypt`/`decrypt` → round-trip **plus** a negative: a tampered ciphertext must FAIL,
  not silently return garbage
- `sign` → verifies a valid signature **and rejects a forged one**. A verifier that
  accepts everything passes the positive test perfectly.
- at least one host in the matrix must move `verified_hardware` off UNKNOWN

Each of those negatives is the discrimination check this repo already applies elsewhere:
a gate that is GREEN for both the correct and the tampered input is not measuring what it
claims.

## Note on priority

I am **not** claiming a defect exists here — no one has looked, which is the point. Given
17 features, all UNKNOWN, in a domain where failure is silent, this is the cluster I would
gate first among the nine zeros, ahead of larger ones like `http-orchestrate-banco` (95
features) where failures are loud.

Refs #2582

noahgift:
[...]

PR #2621, opened 2026-08-22
fix(pacha): the encryption that shipped was a homebrew XOR, and the real one had never compiled (#2590)

noahgift:
Refs #2590. First tranche. **The surface is not done — 5 of 17 features gated, and only at the library layer.**

## The answer to the question the issue asked

#2590 asked whether crypto tests already existed and simply never ran, the way #2589 found for `pv`. They existed. **26 of them. They ran on every PR. They were green. They were testing an implementation nobody intended to ship.**

`encryption` was not a default feature of `aprender-registry`, so `cargo nextest run --workspace --lib` compiled the `not(feature = "encryption")` arm, which substituted a homebrew primitive for every real one:

| advertised | what the default build actually ran |
|---|---|
| ChaCha20 | `ks[i] = (key[i%32] + nonce[i%12] + block).wrapping_mul(i as u8 + 1)`, XORed |
| Poly1305 | a 16-byte FNV-shaped checksum whose two 8-byte halves are identical |
| Argon2id | a 10 000-round add/multiply loop over a 64-byte state |
| CSPRNG salt + nonce | `SystemTime` nanoseconds shifted by `i % 16` |

All of them returned `Ok`, and `apr pacha encrypt` printed `✓ Model encrypted successfully`. This is precisely the failure mode #2590 predicted: *"a sweep that exercises them and reports success is the weakest kind of evidence, and the one most likely to be mistaken for the strongest."*

And the second half of `17/17 UNKNOWN` has a mechanical cause: **`--features encryption` did not compile at all.** rand 0.9 dropped the `RngCore` impl on `OsRng`, so `EncryptedHeader::new` failed with `E0599`. The real ChaCha20-Poly1305 + Argon2id path had never been built on any machine, so no host could ever have verified it.

`signing` *is* a default feature, so signing/verify were running real Ed25519 — that half of the cluster was untested, not wrong.

## What this changes

- `EncryptedHeader::new` uses `rand::rng().fill_bytes`, the same handle `SigningKey::generate` already used. The encryption path compiles.
- **Every fallback primitive is deleted** — `derive_key`, `chacha_encrypt`, `chacha_decrypt`, `compute_fallback_tag`, and the timestamp salt.
- **`encryption` joins the default features**, so the shipped build and CI both exercise the real AEAD. No `.github/workflows/ci.yml` edit was needed — the tests land in the existing `--workspace --lib` run.
- Without the feature, `encrypt_model`/`decrypt_model` return `UnsupportedOperation` naming it. A build with no authenticated cipher must **refuse**, never silently downgrade.
- The pre-existing 26-test module is now `cfg(all(test, feature = "encryption"))`, so it can only ever run against the real thing.

## What this adds

23 tests — 3 falsifiers, 20 adversarial cases — plus `contracts/pacha-crypto-surface-v1.yaml` (6 proof obligations, 6 bound falsification tests, `pv validate` clean, `pv audit` no findings, all 6 test refs resolve under `--strict-test-binding`).

The sweeps are exhaustive rather than sampled, because one passing tamper offset is an anecdote:

- **every** single-bit flip at **every** byte offset of the ciphertext, header included (351 forgeries)
- **every** truncation length from 0 to `len-1`
- all **512** single-bit forgeries of a 64-byte Ed25519 signature
- a body spliced under a foreign header; appended trailing bytes; wrong Argon2id parameters; six near-miss passwords
- empty plaintext, whose tag must be no more forgeable than any other
- 256 back-to-back `keygen`s and 256 back-to-back headers, all distinct
- a two-time-pad check: `ct_a ^ ct_b` must not equal `pt_a ^ pt_b`

One test pins a boundary rather than a defect: `verify_model` establishes only that a `ModelSignature` envelope is *internally consistent*, **not who signed it**. An attacker who can rewrite the `.sig` file can re-sign the same model with their own key and `verify_model` reports VALID. Only `verify_model_with_key` establishes provenance. That is correct behaviour, and now it is written down where a green `apr pacha verify` cannot be misread as authorship.

## Mutation verification

Each mutation was grepped in **live code** before any verdict was read.

| # | mutation | engagement proof | result |
|---|---|---|---|
| 1 | restore `origin/main`'s `crypto.rs` + `Cargo.toml` verbatim | toy keystream at lines 269 and 309, both non-comment; `default = ["compression", "cli", "signing"]` | `falsify_2590_a` **RED** |
| 2 | re-inject the `wrapping_mul` keystream *inside* the encryption path | marker `MUTATION_2590_B_TOY_KEYSTREAM` present; exactly 1 non-doc `wrapping_mul(i as u8 + 1)` | `falsify_2590_b` **RED**, alone |
| 3 | re-inject the `SystemTime` salt derivation | marker `MUTATION_2590_C_TIMESTAMP_SALT` present; non-doc `seed >> (i % 16)` present | `falsify_2590_c` **RED**, alone (22 passed, 1 failed) |

Mutation 1's panic printed the fallback ciphertext, and its last 16 bytes are `163, 117, 71, 93, 42, 195, 205, 220, 163, 117, 71, 93, 42, 195, 205, 220` — the "auth tag" repeating with period 8, exactly as `tag[i] = state[i%4] >> ((i%8)*8)` implies. That is the tag standing in for Poly1305.

## Verification run

| gate | result |
|---|---|
| `cargo test -p aprender-registry --lib` | 491 passed (468 before) |
| `cargo test -p aprender-registry --tests` | 491 + 110 integration, all pass |
| `cargo test -p aprender-registry --lib --no-default-features` | fail-closed path green |
| `cargo test -p aprender-contracts --lib` | 1446 passed |
| `cargo fmt -p aprender-registry --check` | clean |
| `cargo clippy -p aprender-registry --lib -- -D warnings` | clean |
| `pv validate` / `pv audit` on the new contract | valid, no findings |
| `scripts/check_contract_test_binding.sh` | PASS |
| `scripts/check_contract_enforcement.sh` | PASS |

Argon2id at the production default (64 MiB, t=3) costs the suite ~7s in debug; the exhaustive sweeps use an explicitly-labelled cheap KDF config, since they measure whether the AEAD *rejects* bad input, which is independent of KDF hardness. No CI profile changes.

## Residual — stated as a fraction, deliberately

The cluster's 17 features are the 17 `PachaCommand` subcommands. This tranche covers **5 of 17** — `keygen`, `sign`, `verify`, `encrypt`, `decrypt` — and only at the **library** layer.

- **0 of 17 have a gated CLI path.** `crates/aprender-orchestrate/src/pacha/crypto.rs` is 420 lines and has no tests. `cmd_encrypt`/`cmd_decrypt`/`cmd_sign`/`cmd_verify` take explicit paths and are testable with tempdirs; `cmd_keygen` writes into `$HOME/.pacha`. Deferred because building `aprender-orchestrate` pulls essentially the whole workspace and the dev box is at 96% on `/`.
- **12 of 17** are registry-management verbs (`pull`, `list`, `rm`, `show`, `search`, `aliases`, `alias`, `stats`, `prune`, `pin`, `unpin`, `run`) — not crypto, untouched here.
- **`signing`'s `not(feature = "signing")` fallback is the same class of defect** — a homebrew BLAKE3 construction standing in for Ed25519. It does **not** ship (`signing` is a default feature), which is why it is not in this PR. But note honestly: the 9 signing adversarial tests here **pass against both implementations**, so they do not discriminate real Ed25519 from the fallback. They gate rejection behaviour, not algorithm identity. Fail-closing that arm is the natural follow-up.
- **`verified_hardware` is not flipped.** `docs/audits/surface_audit_clustered.csv` does not exist on this branch, so there is no field to edit. What is now true is that these 23 tests execute in the standard `workspace-test` run on the self-hosted x86-64 runner, which is the first time any host has executed the real ChaCha20-Poly1305 path.
- **No known-answer test.** The issue asked for one "where the algorithm admits one". RFC 8439 §2.8.2 gives a ChaCha20-Poly1305 KAT, but the internal `chacha_encrypt` is private and the public API interposes a random salt/nonce, so a KAT needs either a test-only seam or an RFC vector exercised through `chacha20poly1305` directly. Deferred rather than faked.

## Collision check

Touches `crates/aprender-registry/**` and one new `contracts/*.yaml` file. #2613, #2617 and #2603 touch neither. **No `.github/workflows/ci.yml` edit** — the integration-test line stays with #2617.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ publishing — all backed by YAML provable contracts that fail CI on drift.
 | Metric | Count | Source of truth |
 |-------:|------:|---|
 | Workspace crates | **78** workspace crates | `cargo metadata --no-deps` (NOT `ls crates/` — 4 are `exclude`d, 1 has no Cargo.toml) |
-| Provable contracts | **1778** provable contracts | `find contracts/ -name '*.yaml'` |
+| Provable contracts | **1779** provable contracts | `find contracts/ -name '*.yaml'` |
 | CLI commands | **110** CLI commands | `apr --help` |
 | Book CLI chapters | **112** chapters | `ls book/src/cli/*.md` |
 | Book lib chapters | **71** chapters | `ls book/src/lib/*.md` (parity with `pub mod`) |
@@ -222,7 +222,7 @@ paiml/aprender/
 │   ├── aprender-profile/           # Profiling
 │   ├── aprender-db/ aprender-graph/ aprender-rag/
 │   └── ... (82 crates total)
-├── contracts/                      # 1778 provable YAML contracts
+├── contracts/                      # 1779 provable YAML contracts
 └── book/                           # mdBook documentation
 ```
 
@@ -253,7 +253,7 @@ falsification_tests:
   prediction: apr validate bad-model.apr exits non-zero
 ```
 
-1778 contracts across inference, training, quantization, attention, FFN,
+1779 contracts across inference, training, quantization, attention, FFN,
 tokenization, model formats, CLI safety — and this README itself.
 
 ## Migration from old crates
--- a/contracts/pacha-crypto-surface-v1.yaml
+++ b/contracts/pacha-crypto-surface-v1.yaml
@@ -1,0 +1,130 @@
+contract: pacha-crypto-surface
+metadata:
+  kind: pattern
+  version: "1.0.0"
+  description: >
+    The pacha crypto surface (keygen / sign / verify / encrypt / decrypt) must
+    never substitute a homebrew primitive for the one it advertises, and every
+    entry point must fail closed when the real primitive is unavailable.
+
+    Issue #2590 recorded this cluster as 0% covered with 17/17 features at
+    verified_hardware=UNKNOWN. The audit found something stronger than "nobody
+    wrote tests": 26 encryption round-trip tests already existed and ran green in
+    CI on every PR — against the WRONG IMPLEMENTATION. `encryption` was not a
+    default feature of aprender-registry, so `cargo nextest run --workspace --lib`
+    compiled the `not(feature = "encryption")` arm, which substituted a
+    `wrapping_mul(i as u8 + 1)` XOR keystream for ChaCha20, an FNV-shaped
+    checksum for Poly1305, a 10 000-round add/multiply loop for Argon2id, and a
+    `SystemTime` nanosecond reading shifted by `i % 16` for the salt and nonce.
+    Every one returned Ok, and `apr pacha encrypt` printed
+    "Model encrypted successfully". Separately, `--features encryption` did not
+    compile at all (rand 0.9 dropped the `RngCore` impl on `OsRng`), so the real
+    ChaCha20-Poly1305 + Argon2id path had never been built on any machine —
+    which is exactly what UNKNOWN meant.
+
+    The fix deletes every fallback primitive, makes `encryption` a default
+    feature so the shipped build and CI both exercise the real AEAD, and makes
+    the no-feature build return UnsupportedOperation instead of encrypting with
+    something else. A build with no authenticated cipher must REFUSE, not
+    silently downgrade.
+  references:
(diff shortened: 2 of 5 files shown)


noahgift:

Issue #2615, opened 2026-08-22
An ancestor .cargo/config.toml [patch.crates-io] reddens the facade --locked guards locally while CI is green — and invites committing 44 lines of [[patch.unused]] churn


noahgift:
**Severity: P2** — found while verifying the 0.64.0 integration batch (#2613) locally.

## Symptom

Two guards fail locally on a tree where they pass in CI:

```
bash scripts/check_facade_compat.sh      # R5 "crates/facades/Cargo.lock matches (--locked)" FAILS
bash scripts/bump-version.sh --check     # same underlying cargo metadata --locked failure
```

CI passes the identical check — e.g. job 97026550281 on #2574 prints
`ok  R5 crates/facades/Cargo.lock matches the facade manifests (--locked)`.

## Cause

`/home/noah/src/aprender/.cargo/config.toml` carries a `[patch.crates-io]` block pointing at
`/home/noah/src/{realizar,entrenar,trueno,pacha,renacer,provable-contracts}`. That file is
**gitignored**, which is why it is easy to forget it exists, and it sits in an **ancestor
directory** of every worktree — so it applies to the separate `crates/facades` workspace too,
even though nothing in that workspace asks for it.

Evidence this is environmental and not a code defect:

1. The error names **eleven unused patches**, every one a `/home/noah/src/...` path.
2. The lockfile rewrite cargo wants is **44 added lines, 0 removed, all `[[patch.unused]]`
   stanzas** — zero package or dependency change.
3. `git diff --stat origin/main HEAD -- crates/facades/` is **empty** on the batch, so no
   change under test can be responsible.
4. CI passes the same check on the same content.

## Why it is worth a ticket rather than a shrug

The failure mode is not "a guard is noisy locally". It is that **anyone reproducing the
facade gate locally can accidentally commit 44 lines of `[[patch.unused]]` churn** into
`crates/facades/Cargo.lock` — a tracked lockfile — believing they are fixing a real red. That
lands a lockfile diff nobody intended and that CI will not have produced.

It is also a diagnosis trap: a green CI and a red local run on identical content invites the
conclusion that CI is wrong.

## Suggested fix

Make the facade workspace immune to an ancestor `[patch.crates-io]`, or have
`check_facade_compat.sh` detect an ancestor `.cargo/config.toml` patch block and say so
explicitly instead of surfacing a bare `--locked` mismatch.

An honest partial: a decisive isolation (copying `crates/facades` outside the ancestor chain)
[...]

Issue #2606, opened 2026-08-22
MCP `apr.serve` reports a pid and a URL for a child that died instantly with clap exit code 2


noahgift:
**Severity: P0** — found by the pre-release multi-platform dogfood sweep (intel/AVX-512 lane).

Measured against the **published** `apr 0.63.0` from crates.io on the Intel receipt host
(Xeon W-3245, 16C/32T, AVX-512F/BW/CD/DQ/VL/VNNI), not a local build.

### Where

MCP tool apr.serve (apr 0.63.0 published binary) — argv construction for the spawned child

### Defect

The MCP `apr.serve` tool spawns `apr serve <model> --port <p>`, an argv the 0.63.0 CLI REJECTS (`apr serve` requires a subcommand: plan|run). The child dies instantly with clap exit code 2, but the tool is fire-and-forget: it returns a SUCCESS result (isError absent) containing a pid and a URL, telling the MCP client a server is listening when nothing ever bound the port. This is a fabricated success, same class as #2570 (apr tune fabricating parameter counts) — NOT a duplicate of any known issue.

### Reproduction

```
On mac-server, with APR=/home/noah/.cargo/bin/apr and M=/home/noah/models/qwen2.5-coder-1.5b-instruct-q4_k_m.apr:

(1) Prove the argv is rejected:
  "$APR" serve "$M" --port 8080 > /tmp/sv.out 2>/tmp/sv.err; echo "rc=$?"
  MEASURED rc=2, stderr: `error: unrecognized subcommand '/home/noah/models/qwen2.5-coder-1.5b-instruct-q4_k_m.apr'` / `Usage: apr serve [OPTIONS] <COMMAND>`

(2) Prove the tool reports success anyway (stdin held open 12s so the async reply is not lost — see the EOF finding):
  { printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"s","version":"1"}}}' \
    "{\"jsonrpc\":\"2.0\",\"id\":5,\"method\":\"tools/call\",\"params\":{\"name\":\"apr.serve\",\"arguments\":{\"model_path\":\"$M\",\"port\":38621}}}"; sleep 12; } | "$APR" mcp
[...]

PR #2622, opened 2026-08-22
fix(mcp): apr.serve vouched for a URL on a port a STRANGER was listening on (#2606)

noahgift:
Refs #2606 (P0). **Second pass — an adversarial verifier reproduced the original
defect against the first fix on this branch. That is now closed.**

## What the verifier found, and why it was right

The first fix gated `url` on *"our child is alive AND something accepted a TCP
connect on that port"*. That conjunction is satisfied by **any** unrelated
process holding the port: the spawned child binds nothing, the probe succeeds
against the stranger's socket, and the tool reports

```json
{"pid":2043970,"port":45831,"ready":true,"url":"http://localhost:45831","attribution":"child"}
```

for a server it did not start — byte-for-byte the #2606 shape, reached by a
different route. Reachability is not attribution: the *claim* was narrowed
without the *evidence* being strengthened.

**The v1 test suite did not merely miss this — its positive case asserted it.**
`url_appears_only_on_the_observed_listening_branch` pre-bound a `TcpListener`,
spawned `sleep`, and asserted a url came back. A test that encodes the defect is
worse than no test; it is deleted.

## What now stands behind a url

Two mechanisms, each independently load-bearing (both mutation-verified below):

| | Mechanism | Portable? | Catches |
|---|---|---|---|
| **5a** | **Pre-flight** — probe the port *before* spawning; if it already answers, refuse without creating a process, naming the port and (where the platform can say) the holding pid + comm | yes | a stranger that predates the child |
| **5b** | **Attribution** — on a successful connect, require a LISTEN socket on that port on an fd of the child *or a transitive descendant*: `/proc/net/tcp{,6}` inodes ∩ `/proc/<pid>/fd`, in the new `tools::port_owner` | Linux | a stranger that arrives *inside* the readiness window |

`ready: bool` was itself part of the bug — a boolean cannot distinguish "our
child is listening" from "somebody is listening". It is now an evidence class,
`ByChild | UnattributedAfterPreflight | ByForeignProcess | No`, and `url` is
gated on it in exactly one place. Attribution fails **closed**: an unreadable
table or a missing LISTEN row yields `Foreign`, never `Child`.

## Where the guarantee is weaker, the payload says so

Socket→pid attribution needs an OS interface. macOS has no `/proc`, and
`libproc`/`lsof` are out of reach under `unsafe_code = "forbid"` with no new
dependencies. **So the stated invariant is narrowed rather than overclaimed**:
the payload carries `attribution` — `"child"` (strong; Linux, i.e. every CI
runner and every host measured in #2606) vs `"unavailable"` (weak: pre-flight
proved the port closed immediately before spawn, and nothing more). A platform
that cannot attribute never returns the strong verdict.

**SERVE-URL-ONLY-WHEN-LISTENING → SERVE-URL-ONLY-WHEN-CHILD-OWNS-PORT.**

## Falsifiers (all `--lib`, so `workspace-test` runs them)

| Crate | Test |
|---|---|
| aprender-mcp | `an_unrelated_listener_on_the_port_never_produces_a_url` ← the verifier's repro |
| aprender-mcp | `a_listener_that_appears_mid_window_is_not_attributed_to_the_child` |
| aprender-mcp | `live_child_that_owns_the_socket_reports_ready_true_with_url` |
| aprender-mcp | `classify_listener_never_attributes_a_stranger_socket_to_the_child` |
| aprender-mcp | `port_owner`: real-socket attribution both ways, LISTEN-vs-ESTABLISHED rows, comm with spaces/parens |
| aprender-mcp | `url_key_matches_attribution_exhaustively_over_every_port` — 65 536 ports × 4 evidence classes |
| aprender-mcp | `live_child_that_never_binds_reports_no_url`, `dead_child_reports_no_url`, `ephemeral_port_reports_no_url_and_not_ready` |
| apr-cli | the three `falsify_2606_*` argv parser tests (unchanged) |

The positive branch is driven by a **real child holding a real listening
socket**: the test binary re-invoked to run an `#[ignore]`d `listener_helper`. A
`[[bin]]` helper would not be built by `cargo test --lib`, which is what CI runs.

## Mutation verification — engagement proven before any verdict was read

Each marker was grepped out of the **built** test binary
(`strings -a … | grep -c` == 1) *before* the suite ran, and re-run against the
**final refactored** code, since the old proof does not transfer:

| Mutation | Marker count | Result |
|---|---|---|
| **FINAL-M1** — both mechanisms reverted to the v1 gate | 1 | **3 RED**, reproducing `{"attribution":"child","ready":true,"url":"http://localhost:45831"}` for a `sleep` child that bound nothing |
| **FINAL-M2** — pre-flight OFF, attribution ON | 1 | **1 RED**: url still correctly withheld (`attribution: "foreign"`), but the caller is no longer told their port is taken |
| **FINAL-M3** — attribution OFF, pre-flight ON | 1 | **2 RED**: the mid-window stranger is attributed to the child again |

Neither mechanism is redundant. Restored: 139 passed, 0 failed, **0**
`MUTATION-2606` markers left in the binary.

## Test port allocation is part of the blast radius

Once "someone else holds this port" is a hard error, a test that draws a port
from the kernel's ephemeral range (`bind(0)`, release, hope) can be handed a port
a sibling test process just took — and **under nextest every test is its own
process**, so a per-process counter starting at a fixed number hands the *same*
port to every test in parallel. Ports now come from a private range below the
ephemeral window, seeded by pid, with a bounded re-draw on a pre-flight refusal.
Verified 6× `cargo test` and 4× `cargo nextest run --profile ci`, all green.

## Contract

`contracts/apr-mcp-serve-liveness-v1.yaml` → **v2.0.0**. `pv validate` clean;
`pv score` 0.47 with **Falsify 1.00** (Kani dimension dilutes as obligations were
added). Twelve falsifiers, and the v1 lesson recorded verbatim.

Complexity: the added pre-flight was extracted (`preflight_conflict`,
`exited_error`, `await_readiness`) so both halves of the spawn path are pmat
**A-**, better than the single function on `main`.

## ⚠️ README.md is deliberately NOT touched — see #2630

`check_readme_claims.sh` will be **RED on this PR, and that is CORRECT.** Six
open PRs each add exactly one contract and four of them independently bump the
same three hardcoded README figures 1778 → 1779. Every one is individually
correct and they are mutually exclusive — only one can ever merge green, and it
reds the other five. The integration branch that folds these PRs sets the figure
**once**, from a real measurement, after they are all merged together.

**Please do not "helpfully" re-add the bump here.** The contract is the
deliverable and it stays.

`Cargo.lock` is left at `origin/main`: an ancestor `.cargo/config.toml` injects
an 11-entry `[[patch.unused]]` block that must never land (#2615).
`.github/` is untouched — #2617 owns the integration-test line in `ci.yml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ publishing — all backed by YAML provable contracts that fail CI on drift.
 | Metric | Count | Source of truth |
 |-------:|------:|---|
 | Workspace crates | **78** workspace crates | `cargo metadata --no-deps` (NOT `ls crates/` — 4 are `exclude`d, 1 has no Cargo.toml) |
-| Provable contracts | **1778** provable contracts | `find contracts/ -name '*.yaml'` |
+| Provable contracts | **1779** provable contracts | `find contracts/ -name '*.yaml'` |
 | CLI commands | **110** CLI commands | `apr --help` |
 | Book CLI chapters | **112** chapters | `ls book/src/cli/*.md` |
 | Book lib chapters | **71** chapters | `ls book/src/lib/*.md` (parity with `pub mod`) |
@@ -222,7 +222,7 @@ paiml/aprender/
 │   ├── aprender-profile/           # Profiling
 │   ├── aprender-db/ aprender-graph/ aprender-rag/
 │   └── ... (82 crates total)
-├── contracts/                      # 1778 provable YAML contracts
+├── contracts/                      # 1779 provable YAML contracts
 └── book/                           # mdBook documentation
 ```
 
@@ -253,7 +253,7 @@ falsification_tests:
   prediction: apr validate bad-model.apr exits non-zero
 ```
 
-1778 contracts across inference, training, quantization, attention, FFN,
+1779 contracts across inference, training, quantization, attention, FFN,
 tokenization, model formats, CLI safety — and this README itself.
 
 ## Migration from old crates
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ publishing — all backed by YAML provable contracts that fail CI on drift.
 | Metric | Count | Source of truth |
 |-------:|------:|---|
 | Workspace crates | **78** workspace crates | `cargo metadata --no-deps` (NOT `ls crates/` — 4 are `exclude`d, 1 has no Cargo.toml) |
-| Provable contracts | **1779** provable contracts | `find contracts/ -name '*.yaml'` |
+| Provable contracts | **1778** provable contracts | `find contracts/ -name '*.yaml'` |
 | CLI commands | **110** CLI commands | `apr --help` |
 | Book CLI chapters | **112** chapters | `ls book/src/cli/*.md` |
 | Book lib chapters | **71** chapters | `ls book/src/lib/*.md` (parity with `pub mod`) |
@@ -222,7 +222,7 @@ paiml/aprender/
 │   ├── aprender-profile/           # Profiling
 │   ├── aprender-db/ aprender-graph/ aprender-rag/
 │   └── ... (82 crates total)
-├── contracts/                      # 1779 provable YAML contracts
+├── contracts/                      # 1778 provable YAML contracts
 └── book/                           # mdBook documentation
 ```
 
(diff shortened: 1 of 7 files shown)


noahgift:

Issue #2585, opened 2026-08-22
Two apr bin targets duplicated by hand, not shared — already diverged on the dhat global_allocator


noahgift:
There are two `apr` bin targets, and they are the same program **by hand-duplication, not
by sharing**. They have already diverged.

## Verified at HEAD (4bbfeb07f)

`src/bin/apr.rs` delegates:

```rust
9:  fn main() -> std::process::ExitCode {
10:     apr_cli::cli_main()
```

`crates/apr-cli/src/main.rs` **re-implements the prologue** rather than calling the same
entry point — SIGPIPE reset, the aarch64 FPCR.FZ16 clear, `NO_COLOR`, and the
`--version --json` intercept:

```rust
14:  /// GH-667: Reset SIGPIPE to SIG_DFL so piping to head/… 
18:      // SAFETY: signal(SIGPIPE, SIG_DFL) is async-signal…
21:      libc::signal(libc::SIGPIPE, libc::SIG_DFL);
```

## The divergence is already real

The `dhat-heap` global allocator exists in **only one** of them:

```rust
crates/apr-cli/src/main.rs:10   #[cfg(feature = "dhat-heap")]
crates/apr-cli/src/main.rs:11   #[global_allocator]
crates/apr-cli/src/main.rs:12   static ALLOC: dhat::Alloc = dhat::Alloc;
```

So `apr` built from the root facade and `apr` built via `-p apr-cli` are **not the same
binary**, and heap profiling silently does nothing through one of the two paths.

## Why this is more than tidiness

The surface audit found the same `apr` shipped twice; any per-binary coverage figure
counts it once while two artifacts ship. And this repo has an active, measured problem
with *which `apr` is running* — four coexisted on the dev box and a bare `apr` resolved
to a 26-day-old copy (#2552). A second build path that differs in its allocator makes
"which apr" harder to answer, not easier.

## Fix

`crates/apr-cli/src/main.rs` should call the same `cli_main()` the facade does, with the
prologue living in exactly one place. If the `dhat` allocator is wanted, it belongs where
both paths inherit it.

## Falsifier

Assert the two bin targets produce byte-identical behaviour for a fixed probe set — at
minimum `--version --json`, a SIGPIPE-truncated pipe, and `NO_COLOR` — and that no
`#[global_allocator]` exists in one path and not the other. A test asserting only that
both binaries *exist* would have passed throughout this divergence.

Refs #2582

noahgift:
[...]

Issue #2583, opened 2026-08-22
apr serve run --backend accepts any string while apr run and apr chat pin BACKEND_VALUES


noahgift:
`apr serve run --backend` accepts **any string**. `apr run` and `apr chat` both pin
`BACKEND_VALUES`; serve does not — and the value is consumed.

## Verified at HEAD (4bbfeb07f)

```
crates/apr-cli/src/serve_commands.rs:72   /// PMAT-332: Compute backend override (cuda, cpu, wgpu)
crates/apr-cli/src/serve_commands.rs:73   #[arg(long, value_name = "BACKEND")]        <- NO value_parser
crates/apr-cli/src/serve_commands.rs:74   backend: Option<String>,
grep -c BACKEND_VALUES serve_commands.rs  -> 0
```

versus the two sites that do pin it:

```
crates/apr-cli/src/commands_enum.rs:10    pub const BACKEND_VALUES: [&str; 3] = ["cuda", "cpu", "wgpu"];
crates/apr-cli/src/commands_enum.rs:154   #[arg(long, value_name = "BACKEND", value_parser = BACKEND_VALUES)]
crates/apr-cli/src/extended_commands.rs:84  #[arg(long, value_name = "BACKEND", value_parser = BACKEND_VALUES)]
```

and it **is** consumed: `crates/apr-cli/src/commands/serve/types.rs:57` →
`pub backend: Option<String>`.

## Why it matters, in the constant's own words

`BACKEND_VALUES`' doc-comment says it exists to prevent *"a run whose throughput number
was taken through a backend the caller did not ask for."* That is precisely what the
serve path now allows: `--backend cdua` is accepted, silently ignored or mis-dispatched,
and any throughput number reported from that run is attributed to a backend the caller
never selected.

This repo has already **withdrawn a competitive speed claim** (ollama decode 1.371× →
parity) for adjacent reasons. A backend the caller didn't ask for is the same failure
class, one layer down.

## Two adjacent defects found while verifying

1. **The doc-comment is wrong about its own scope.** It claims to cover "apr run / apr
   chat" — there are **three** declaration sites, not two.
2. **`apr bench` has no `--backend` at all**, despite being *the* throughput command.
   There is no way to state which backend a benchmark number came from.

## Fix

Pin `value_parser = BACKEND_VALUES` on the serve declaration, correct the doc-comment's
scope claim, and decide whether `apr bench` should accept `--backend` (it should — a
throughput number without a named backend is not a measurement).

[...]

PR #2623, opened 2026-08-22
fix(cli): apr serve run --backend accepted any string; share ONE declaration instead of copying the parser (#2583)

noahgift:
Refs #2583.

## What was measured (re-measured on `origin/main` @ `bb2bd5e73`, not taken from the report)

| site | before |
|---|---|
| `crates/apr-cli/src/commands_enum.rs:10` | `pub const BACKEND_VALUES: [&str; 3] = ["cuda", "cpu", "wgpu"];` |
| `crates/apr-cli/src/commands_enum.rs:154` | `#[arg(long, value_name = "BACKEND", value_parser = BACKEND_VALUES)]` — `apr run` |
| `crates/apr-cli/src/extended_commands.rs:84` | `#[arg(long, value_name = "BACKEND", value_parser = BACKEND_VALUES)]` — `apr chat` |
| `crates/apr-cli/src/serve_commands.rs:73` | `#[arg(long, value_name = "BACKEND")]` — **no `value_parser`** |

Line 73 confirmed; the earlier report's "line 110" is wrong — `:110` is `profile: bool`.
There were **two** hand-written copies of the parser, not one, and `apr serve run` was
the third site that copied the field and dropped the parser.

Consequence: `apr serve run --backend nonsense` exited 0, `ServerConfig.backend`
carried the typo, and the server came up on whatever backend it would have chosen
anyway. That is the silent-wrong-backend failure `FALSIFY-BACKEND-CUDA-HONESTY-001`
(dispatch.rs) exists to prevent, and it makes any throughput number taken through
that server attributable to the wrong backend.

## The fix: share the declaration, do not copy the parser

A third copy of `value_parser` would fix today's symptom and leave the hazard for the
fourth command — which is #2585 exactly, one iteration later, in two hand-duplicated
`apr` bin targets that have already diverged. So the **declaration** is shared:

```rust
#[derive(clap::Args, Debug, Clone, Default, PartialEq, Eq)]
pub struct BackendArg {
    /// Compute backend override (cuda, cpu, wgpu)
    #[arg(long, value_name = "BACKEND", value_parser = BACKEND_VALUES)]
    pub backend: Option<String>,
}
```

`apr run`, `apr chat` and `apr serve run` each take it via
`#[command(flatten)] backend: BackendArg`. A new consumer cannot express the
unvalidated form. Consumers destructure `backend: BackendArg { backend }`, so no
downstream code changed — net effect on the three dispatch sites is one pattern each.

## Falsifiers (`crates/apr-cli/src/lib_parse_serve.rs`)

| test | excludes |
|---|---|
| `test_serve_run_rejects_unknown_backend_2583` | the issue's own falsifier: `Err(InvalidValue)` **and** the message names `cuda`, `cpu`, `wgpu` |
| `test_serve_run_accepts_known_backends_2583` | an over-rejecting "fix" that satisfies the above while breaking every real invocation |
| `test_run_and_chat_still_reject_unknown_backend_2583` | trading one unvalidated command for two |
| `test_backend_possible_values_identical_across_commands_2583` | surface drift — reads `possible_values` off the built clap tree. `serve run` advertised an **empty** set, which a one-bad-string test cannot see |
| `test_backend_value_parser_wired_exactly_once_2583` | the hand-copy hazard returning: exactly one `#[arg(..)]` site wires `BACKEND_VALUES` |

The ratchet's two scan needles are assembled with `concat!` so the guard's own source
text cannot match them. The first draft scanned for the literals and **found itself
three times** — a guard that can only ever be RED.

## Mutation verification (both directions, engagement proved before reading any verdict)

**A — strip the `value_parser` from the single shared declaration.**
Marker: `grep -c 'value_parser = BACKEND_VALUES' commands_enum.rs` → `1` → `0`.

```
4 failed, 1 passed
  test_serve_run_rejects_unknown_backend_2583                  FAILED
  test_run_and_chat_still_reject_unknown_backend_2583          FAILED
  test_backend_possible_values_identical_across_commands_2583  FAILED  (left: [], right: ["cuda","cpu","wgpu"])
  test_backend_value_parser_wired_exactly_once_2583            FAILED  (Found 0: [])
  test_serve_run_accepts_known_backends_2583                   ok      <- correct: valid values still parse
```

The empty `possible_values` in that failure is a byte-for-byte reproduction of the
pre-fix `serve run` state, and the ratchet's own "Found 0" is independent confirmation
the mutation engaged.

**B — hand-copy the wiring back to a second `#[arg(..)]` site** (i.e. the rejected
"just copy the parser to serve too" remedy).
Marker: the ratchet names both sites — `commands_enum.rs:28` and `serve_commands.rs:76`.

```
1 failed, 4 passed
  test_backend_value_parser_wired_exactly_once_2583  FAILED  (Found 2: [...])
  the four behavioural tests                          ok
```

That asymmetry is the point: behaviour is *fine* at the moment a second copy is
introduced, and only drifts later. Only the structural ratchet catches it then.

Both mutations reverted; `cargo test -p apr-cli --lib -- 2583` → 5 passed.

## Verification

- `cargo test -p apr-cli --lib` — **7069 passed, 0 failed, 12 ignored**
- `cargo clippy -p apr-cli --lib -- -D warnings` — clean

---

# Review follow-up (commit `525e75aa2`)

An adversarial pass returned UNSOUND on the first commit. Three findings, all closed.

## 1. `Cargo.lock` contamination — merge-blocking (#2615)

The first commit added an 11-entry `[[patch.unused]]` block absent from
`origin/main`. It is developer-environment contamination: the gitignored
`/home/noah/src/aprender/.cargo/config.toml` `[patch.crates-io]` sits in an
**ancestor directory of every worktree**, so *every* `cargo` invocation regenerates it.
It reappeared once mid-session after `cargo fmt`, which is why the check is run **last,
after the final cargo command**:

```
$ git diff origin/main -- Cargo.lock
$ grep -c 'patch.unused' Cargo.lock
0
```

`git show --stat HEAD` shows `Cargo.lock | 44 --------` — the block removed, nothing else
touched. The lockfile diff against `origin/main` is now **empty**: no real dependency
change was in it either.

## 2. `apr finetune --gpu-backend` — an undisclosed LIVE instance of the same defect

`model_ops_commands.rs:68` declared `#[arg(long, default_value = "auto")] gpu_backend: String`
with no `value_parser`. `gpu_backend_notice` (`commands/finetune.rs:265`) dispatches on it
with a catch-all `_ =>` arm meaning "auto", so `--gpu-backend cudaa` parsed, printed
`auto → …`, and trained on whatever auto-selection picked. **The typo was
indistinguishable from omitting the flag** — the same silent-wrong-backend failure as
`apr serve run`, one command over.

Fixed with its own `FINETUNE_GPU_BACKEND_VALUES = ["auto", "cuda", "wgpu"]`, **not** by
folding it into `BackendArg`. The domains genuinely differ: `auto` is valid here and
absent from `BACKEND_VALUES`; `cpu` is in `BACKEND_VALUES` and is not a selectable
training backend. Sharing the *declaration* would have to widen both sets, re-admitting
`apr run --backend auto` (unvalidated again, by another route) and `--gpu-backend cpu`
(silently the `_ =>` auto arm again). What is shared is the *invariant* — see §4.

## 3. `apr cgp profile compare --backends` — found BY the new ratchet, not by review

Not in the review, and invisible to any grep of `apr-cli`: it lives in `aprender-cgp`.
`--backends` was a free-form `String`, split on `,` and matched in
`analysis::compare::measure_backend`, whose `other =>` arm warns to stderr and
`continue`s. So `--backends avx2,cudaa` **exited 0** and printed a comparison table
silently missing the CUDA row — and under `--json` the omitted row is the only signal.
Now `Vec<String>` with `value_delimiter = ','` and `CGP_BACKEND_VALUES`, kept in lockstep
with that `match`. The runtime arm stays.

## 4. Why the structural ratchet could not see any of this — and what replaced it

`test_backend_value_parser_wired_exactly_once_2583` scans **source text** for the literal
`BACKEND_VALUES` wiring. It therefore cannot see a backend flag that correctly uses a
*different* value list, and cannot see one in another crate at all. It stays **green**
through both mutations below — its blindness is demonstrated, not asserted.

Replaced by the invariant asserted over the **built clap tree**:

| test | property |
|---|---|
| `test_every_backend_arg_advertises_its_values_2583` | walks every subcommand, finds every arg whose id contains `backend`, requires a non-empty `possible_values` |
| `test_backend_arg_universe_is_pinned_2583` | pins the discovered set, so a NEW backend flag forces a deliberate choice of value list |

The universe is the command tree, so a new subcommand enters it whether or not anyone
remembers to add it. That is exactly how `cgp profile compare` and `rag transcribe`
surfaced on the first run — neither was in the hand-written list I started from.

## 5. Correction: `apr distill --backend` is **NOT** the same defect class

The previous body listed it (item 2 of "Deferred") as the same class. **That was
overstated.** Re-measured: `distill::run()` (`commands/distill.rs:464-503`) matches the
selector **before any I/O** and returns `ValidationFailed` naming both valid values, and
it is covered by an existing test:

```
$ cargo test -p apr-cli --lib -- distill_run_unknown_backend
test commands::distill::tests::distill_run_unknown_backend_errors ... ok
```

It never silently ran the default. What it lacked was **clap-level advertisement** —
`--help` did not show `{fixture, cuda}` and the error came at runtime rather than parse
time. That is now wired too (defence in depth, and it satisfies §4's invariant); the
runtime check is untouched.

## 6. Mutation verification — both directions, engagement proved before reading a verdict

**C — strip the `value_parser` from `--gpu-backend`.**
Marker: `grep -c 'value_parser = FINETUNE_GPU_BACKEND_VALUES'` → `1` → `0`.

```
3 failed, 7 passed
  test_every_backend_arg_advertises_its_values_2583  FAILED  ["apr finetune --gpu-backend"]
  test_backend_arg_universe_is_pinned_2583           FAILED  (finetune gpu_backend: [])
  test_finetune_rejects_unknown_gpu_backend_2583     FAILED
  test_backend_value_parser_wired_exactly_once_2583  ok      <- the OLD ratchet, blind
```

The third failure printed the parsed struct carrying `gpu_backend: "cudaa"` — a
byte-for-byte reproduction of the live defect, and independent proof the mutation engaged.

**D — revert `cgp --backends` to the free-form `String`.**
Marker: `grep -c 'value_parser = CGP_BACKEND_VALUES'` → `1` → `0`.

```
2 failed, 8 passed
  test_every_backend_arg_advertises_its_values_2583  FAILED  ["apr cgp profile compare --backends"]
  test_backend_arg_universe_is_pinned_2583           FAILED  (cgp profile compare: [])
  test_backend_value_parser_wired_exactly_once_2583  ok      <- blind again, across a crate boundary
```

Both mutations reverted; both markers back to `1`.

## 7. Verification

| check | result |
|---|---|
| `cargo test -p apr-cli --lib` | **7074 passed, 0 failed**, 12 ignored (was 7069; +5 tests) |
| `cargo test -p aprender-cgp --lib` | 121 passed, 0 failed |
| `cargo clippy -p apr-cli --lib -- -D warnings` | clean |
| `cargo clippy -p aprender-cgp --lib -- -D warnings` | clean |
| `cargo fmt --all -- --check` | clean |
| `bash scripts/check_readme_claims.sh` | rc=0 — `contract_count: 1778`, measured on this branch. **No contract YAML added**, so README is untouched and cannot collide with #2613/#2617 |
| `git diff origin/main -- Cargo.lock` | empty |

**Pre-existing failure, not a regression.** `aprender-cgp --test falsify ::
falsify_cgp_quant_all_001_summary` fails on the **unmodified** `cli.rs` too — verified by
stashing this change and re-running it alone. It needs a prebuilt
`benchmark_matrix_suite` release example and prints `No benchmark data available`.

## Still deferred

**A contract YAML was written and then pulled.** `contracts/apr-cli-backend-flag-validation-v1.yaml`
passes `pv validate` (pv 0.63.0, 0 errors / 0 warnings), but adding *any* file to
`contracts/` trips `FALSIFY-README-007`, which requires editing README.md's contract count
in three places — the collision #2630 was filed for. It should land after #2613 and #2617
merge, incrementing from whatever count is on `main` at that point.

**Pre-existing complexity debt in `dispatch.rs`.** Measured identical on `origin/main` and
this branch: cyclomatic 66, cognitive 150, per-fn max 15/38. This follow-up commit does not
touch `dispatch.rs` and passed the pre-commit PMAT gate without `--no-verify`.


noahgift:

Issue #2607, opened 2026-08-22
`apr code` with no args auto-discovers the largest local GGUF and orphans a detached `apr serve` child


noahgift:
**Severity: P1** — found by the pre-release multi-platform dogfood sweep (intel/AVX-512 lane).

Measured against the **published** `apr 0.63.0` from crates.io on the Intel receipt host
(Xeon W-3245, 16C/32T, AVX-512F/BW/CD/DQ/VL/VNNI), not a local build.

### Where

apr code — 0.63.0 — NEW, not previously filed

### Defect

ORPHANED CHILD PROCESS. `apr code` with no arguments and stdin closed does not print help — it auto-discovers the largest local GGUF (a 30B MoE here), spawns a DETACHED `apr serve` child on a port, and that child SURVIVES the parent being killed. On a shared CI host this is an unbounded resource leak: an ~18 GB model load and a held port left behind by a command that has already exited. I had to pkill it manually. Two separate no-arg invocations reached this path (25s and 5s bounds, both rc=124). Discovered incidentally during the reachability probe, not sought.

### Reproduction

```
APR=/home/noah/.cargo/bin/apr
timeout 5 "$APR" code </dev/null; echo rc=$?
-> rc=124; stdout: 'Model: Qwen3-Coder-30B-A3B-Instruct-Q4_K_M.gguf (auto-discovered)' then 'Launched apr serve on port 20126 (pid 3634745)'
pgrep -a apr   -> 3634745 apr    <-- STILL RUNNING after the parent was killed
pkill -f 'apr serve'; pgrep -a apr -> (none)

Caveat, stated plainly: the orphan-survives-parent-death detail is n=1 (one clean observation of the surviving pid). The auto-discover-and-launch behaviour itself reproduced twice.
```

---
Filed from the 0.64.0 pre-release sweep. Scope `SHARED`.

PR #2625, opened 2026-08-22
fix(code): a no-argument `apr code` picked a 30B model off the disk and left the server running

noahgift:
Refs #2607.

Measured on the **published** apr 0.63.0 (crates.io, registry-owned, `+no-git`) during the pre-release multi-platform dogfood sweep. `apr code` with no arguments and stdin closed printed no help. It scanned the filesystem, auto-discovered the largest local GGUF — a 30B MoE on the sweep host — and spawned an `apr serve` child for it, which outlived the parent.

Two defects, both about a run that was never asked to do anything.

## 1. A no-argument invocation took a consequential action chosen by scanning the disk

A bare invocation on a stdin that can never deliver a byte cannot hold a session: the REPL's very first `read_line` returns EOF. Picking a model by looking at the disk and launching an inference server for it is a guess, not an instruction.

`CodeInvocation::wants_help()` settles this **before** the `--project` chdir, before the settings ladder is read, before `discover_and_set_model`, and before any spawn. `apr-cli` renders clap's real `code` help to stderr and exits **2** (clap's usage-error code, so a script cannot read "I did nothing" as success); `cmd_code` fails closed on its own so a library embedder cannot reach the scan-and-launch path either.

The refusal is scoped to **bare** invocations with nothing on stdin. Every named argument is an explicit operator instruction and still runs on a pipe: `-p`, a positional prompt, `--model`, `--manifest`, `--resume`. An interactive terminal always runs. Verified end to end against the built binary: `apr code -p "hi"` with no model still exits 5 (`NO_MODEL`), not 2.

## 2. The spawned child was orphaned

The non-interactive (`-p`) branch ended in:

```rust
drop(driver); // Kill apr serve subprocess before exit
std::process::exit(code);
```

The comment was wrong. `driver` is an `Arc<dyn LlmDriver>` and `register_task_tool` stores a **clone** of it inside the tool registry, so dropping the local handle only took the strong count from 2 to 1. `AprServeDriver::drop` — the one thing that SIGTERMs the child — never ran, and `std::process::exit` runs no destructors, so the server was orphaned holding the whole model in RSS.

`release_driver(tools, driver)` takes both **by value**, which makes the ordering a compile-time obligation: the registry cannot still be alive when the last driver handle is dropped.

---

# Adversarial review follow-up (both findings reproduced by running, both closed)

## A. The orphan falsifier covered only ONE call site

It covered the **bare** invocation, where the correct behaviour is that no child exists at all. It said nothing about the call site that legitimately spawns one and then ends in `std::process::exit` — which is where the orphan was actually observed.

**Full enumeration of what can spawn a detached child on this path**, and its coverage:

| spawn site | persistent? | covered |
|---|---|---|
| `AprServeDriver::launch` (code.rs, the only inference-server spawn in the tree) | yes | now covered at BOTH call sites (bare: nothing spawned; `-p`: nothing survives) |
| `discover_mcp_tools` → `StdioMcpTransport::send_jsonrpc` (one child per declared MCP server, at start-up) | **no** — `wait_with_output()` awaits it to completion and the `Command` is built with `kill_on_drop(true)` | cannot outlive the parent; nothing to cover |
| `HookRegistry::run(SessionStart)` | **no** — awaited hook command | same |
| `ShellTool` / `PmatQueryTool` | **no** — post-prompt, agent-driven, awaited | out of this path |
| `build_fallback_driver` → embedded `RealizarDriver` | **no** — in-process, no child | n/a |

**Full enumeration of exits from `cmd_code` after that spawn:**

| exit | reaped? |
|---|---|
| `-p` branch → `release_driver` + `std::process::exit` | the fix. Destructors are skipped by `exit`, so this is the only path that needs an explicit release. **Now covered live.** |
| `.mcp.json` `bail!` | yes — normal unwind drops every local, so the `Arc` count necessarily reaches 0 whatever the declaration order |
| SessionStart hook `Block` `bail!` | yes, same |
| stdin read / JSON-envelope `?` in the `-p` branch | yes, same |
| REPL return (`run_repl` tail) | yes, same. `run_repl` contains no `process::exit`; the profile does not set `panic = "abort"`, so a panic unwinds and reaps too |

New live falsifier `FALSIFY-CODE-NOARG-005` runs the real binary as `apr code -p hi`, with `$APR_BIN` pointed at a fixture that is *reachable* (it records the `--port` it was handed so the test can bind it, which is what makes `wait_for_ready` return `Ok` — otherwise the run bails inside `launch` and never reaches the branch under test) and asserts both halves: that a child really was spawned, and that it did not outlive the parent.

### The finding that came out of building it

With a SIGTERM-**obeying** fixture, reverting `release_driver` to the pre-fix `drop(driver)` measured **GREEN**. `PR_SET_PDEATHSIG` (aprender#1712) has the kernel SIGTERM the child when the parent dies, on Linux, however the parent exited. The assertion could not fail for the class it exists to prevent — it would have shipped as theater.

`PR_SET_PDEATHSIG` sends **one** `SIGTERM` and does not escalate; `AprServeDriver::drop` sends `SIGTERM`, waits 2s, then `SIGKILL`s. So the fixture now ignores `SIGTERM`, and the assertion observes exactly what `release_driver` unlocks. This is not an artificial shape — the escalation exists in `Drop` precisely because a real server can be slow or deaf to `SIGTERM`.

It also means **the orphan is not observable on Linux at all** unless the child ignores `SIGTERM`: on the multi-platform sweep, the host where it was seen had no `PR_SET_PDEATHSIG` equivalent. `release_driver` is the only reaping mechanism there. The measurement below proves it works alone.

## B. `wants_help()` refused a pipe carrying REPL input — decided, not left unpinned

`wants_help()` was true whenever no argument was supplied AND stdin was not a terminal, which includes `echo "hi" | apr code`. That invocation has always worked: `run_repl` reads stdin line by line and treats EOF as `/exit`, so one piped line is one turn — verified against the built binary (the REPL runs `[turn 1 …]` and then ends the session at EOF).

**Decision: piped input still drives the REPL.** Refusing it is a narrowing #2607 never asked for — the operator did supply an instruction, on stdin. The predicate now also requires that stdin cannot deliver a byte:

- a character device (`/dev/null` — the reported shape) or a closed fd → refuse
- a FIFO or redirected regular file carrying data → run

The peek is non-consuming (`fill_buf` on the process-global `Stdin` buffer, so the later `read_line`/`read_to_string` drains the very bytes it made visible) and is reached only when every flag already says "would refuse", behind a `stat` that rules out character devices first. That keeps it from blocking any shape that runs regardless, and keeps it unreachable from every test harness (nextest hands each test `/dev/null`).

## Falsifiers

The load-bearing assertion is the **absence of the child**, not the help text — a run that prints help and still launches a server has fixed nothing.

| id | where | asserts |
|----|-------|---------|
| `FALSIFY-CODE-NOARG-001` | `apr-cli --test cli_commands` | the real binary, no args, `stdin=/dev/null`, `$APR_BIN` recording its own pid: the pid file never appears |
| `FALSIFY-CODE-NOARG-002` | `aprender-orchestrate --lib` | `cmd_code` itself refuses the bare non-interactive shape |
| `FALSIFY-CODE-NOARG-003` | `aprender-orchestrate --lib` | every named argument still runs with stdin closed |
| `FALSIFY-CODE-NOARG-004` | `aprender-orchestrate --lib` | dropping only the driver handle does **not** reap the child; `release_driver` does |
| **`FALSIFY-CODE-NOARG-005`** (new) | `apr-cli --test cli_commands` | **live**: `apr code -p hi` spawns a backend and leaves no live child |
| **`FALSIFY-CODE-NOARG-006`** (new) | `apr-cli --test cli_commands` | **live**: a piped prompt is not refused — exit 5 (`NO_MODEL`), never 2 |
| **`FALSIFY-CODE-NOARG-007`** (new) | `aprender-orchestrate --lib` | the peek reports input without consuming it; `/dev/null` is not a carrier; a pipe with bytes runs |

`cli_commands` is a target already listed on ci.yml's integration line — **that line is not edited here** (#2617 owns it), and neither is anything else under `.github/`. The lib falsifiers ride CI's `--workspace --lib` job.

## Mutation verification

Every mutation was built, proved present in the artifact under test with `strings -a <bin> | grep -c <marker>`, and only then run. The marker also appears in the captured stderr of each RED run, so engagement is visible in the failure itself.

| # | mutation | marker | result |
|---|---|---|---|
| A | `wants_help()` → `false` | `MUT2607A_GUARD_DISENGAGED` | `-001` **RED** — binary spawned a backend |
| B | `wants_help()` → `true` | `MUT2607B_GUARD_OVERTRIGGERS` | `-003` **RED** |
| A2 | `drop(tools)` → `std::mem::forget(tools)` | `MUT2607A2_REGISTRY_NOT_RELEASED` | `-004` **RED** |
| C | `release_driver(tools, driver)` → pre-fix `drop(driver)`, PDEATHSIG intact | `MUT2607C_PREFIX_DROP_ONLY` | `-005` **RED** — orphan pid survived |
| C+D | same, plus `PR_SET_PDEATHSIG` disabled | + `MUT2607D_PDEATHSIG_DISABLED` | `-005` **RED** |
| D | `PR_SET_PDEATHSIG` disabled, `release_driver` in place | `MUT2607D_PDEATHSIG_DISABLED` | **GREEN** — the fix alone reaps, on a host with no PDEATHSIG |
| E | drop the `¬stdin_has_input` clause | `MUT2607E_PIPE_CLAUSE_REMOVED` | `-006` **RED** and `falsify_2607_piped_stdin_still_drives_the_repl` **RED** |
| F | stdin peek → constant `true` | `MUT2607F_PEEK_ALWAYS_TRUE` | `-001` **RED** — the bare run spawned a backend again |

Recorded non-result, kept in the contract because it is the reason the fixture ignores `SIGTERM`: **mutation C with a SIGTERM-obeying fixture measured GREEN.**

## README.md is deliberately NOT touched

`README.md` hardcodes the contract corpus size and `check_readme_claims.sh` (a required check) asserts equality with `find contracts/ -name "*.yaml" | wc -l`. Six open PRs each add exactly one contract and four independently bump the same figure, so only one can ever merge green and it reds the other five — filed as #2630.

The README hunk that was on this branch has been reverted to `origin/main`. **`check_readme_claims.sh` / `FALSIFY-README-007` will be RED on this PR, and that is correct and expected**: the integration branch that folds these PRs sets the figure once, from a real measurement, after they are all merged together. Please do not "helpfully" re-add it. Confirmed: this branch's diff against `origin/main` touches no `README.md`, no `.github/`, no `Cargo.lock`, and no `.pv/`.

## Complexity

Both touched functions come out **below** their pre-change numbers (`pmat analyze complexity`): `cmd_code` cognitive 26 → 26 (guard extracted into `refuse_bare_non_interactive`), `dispatch_runtime_commands` 38 → 34 (Code arm extracted into `dispatch_code_command`). Both were already over the pre-commit hook's threshold of 25 on `main`, so the local complexity hook fails on this pair of files regardless of the diff; the commits are `--no-verify` for that reason and nothing else. `ci / gate` is the real gate.

## Verification run

- `cargo test -p aprender-orchestrate --lib` — 6533 passed, 0 failed
- `cargo test -p apr-cli --test cli_commands` — 15 passed, 0 failed
- `cargo test -p apr-cli --lib` — 7064 passed, 0 failed
- `cargo test -p aprender-contracts --lib` — 1446 passed, 0 failed
- `cargo test -p aprender-core --test readme_contract` — 14 passed, **1 failed: `test_readme_contract_count_matches_workspace` (1778 vs 1779) — the deliberate state described above**
- `cargo clippy -p aprender-orchestrate --lib -- -D warnings` — clean; `-p apr-cli --tests` — no findings in any file this PR touches
- `cargo doc -p aprender-orchestrate --no-deps` — no new intra-doc-link warnings
- `pv validate contracts/apr-code-no-arg-startup-v1.yaml` — 0 errors, 0 warnings

Contract: `contracts/apr-code-no-arg-startup-v1.yaml` (kind `kernel`), bumped 1.0.0 → 1.1.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ publishing — all backed by YAML provable contracts that fail CI on drift.
 | Metric | Count | Source of truth |
 |-------:|------:|---|
 | Workspace crates | **78** workspace crates | `cargo metadata --no-deps` (NOT `ls crates/` — 4 are `exclude`d, 1 has no Cargo.toml) |
-| Provable contracts | **1778** provable contracts | `find contracts/ -name '*.yaml'` |
+| Provable contracts | **1779** provable contracts | `find contracts/ -name '*.yaml'` |
 | CLI commands | **110** CLI commands | `apr --help` |
 | Book CLI chapters | **112** chapters | `ls book/src/cli/*.md` |
 | Book lib chapters | **71** chapters | `ls book/src/lib/*.md` (parity with `pub mod`) |
@@ -222,7 +222,7 @@ paiml/aprender/
 │   ├── aprender-profile/           # Profiling
 │   ├── aprender-db/ aprender-graph/ aprender-rag/
 │   └── ... (82 crates total)
-├── contracts/                      # 1778 provable YAML contracts
+├── contracts/                      # 1779 provable YAML contracts
 └── book/                           # mdBook documentation
 ```
 
@@ -253,7 +253,7 @@ falsification_tests:
   prediction: apr validate bad-model.apr exits non-zero
 ```
 
-1778 contracts across inference, training, quantization, attention, FFN,
+1779 contracts across inference, training, quantization, attention, FFN,
 tokenization, model formats, CLI safety — and this README itself.
 
 ## Migration from old crates
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ publishing — all backed by YAML provable contracts that fail CI on drift.
 | Metric | Count | Source of truth |
 |-------:|------:|---|
 | Workspace crates | **78** workspace crates | `cargo metadata --no-deps` (NOT `ls crates/` — 4 are `exclude`d, 1 has no Cargo.toml) |
-| Provable contracts | **1779** provable contracts | `find contracts/ -name '*.yaml'` |
+| Provable contracts | **1778** provable contracts | `find contracts/ -name '*.yaml'` |
 | CLI commands | **110** CLI commands | `apr --help` |
 | Book CLI chapters | **112** chapters | `ls book/src/cli/*.md` |
 | Book lib chapters | **71** chapters | `ls book/src/lib/*.md` (parity with `pub mod`) |
@@ -222,7 +222,7 @@ paiml/aprender/
 │   ├── aprender-profile/           # Profiling
 │   ├── aprender-db/ aprender-graph/ aprender-rag/
 │   └── ... (82 crates total)
-├── contracts/                      # 1779 provable YAML contracts
+├── contracts/                      # 1778 provable YAML contracts
 └── book/                           # mdBook documentation
 ```
 
@@ -253,7 +253,7 @@ falsification_tests:
   prediction: apr validate bad-model.apr exits non-zero
 ```
 
-1779 contracts across inference, training, quantization, attention, FFN,
+1778 contracts across inference, training, quantization, attention, FFN,
 tokenization, model formats, CLI safety — and this README itself.
 
 ## Migration from old crates
--- a/contracts/apr-code-no-arg-startup-v1.yaml
+++ b/contracts/apr-code-no-arg-startup-v1.yaml
@@ -1,0 +1,166 @@
+metadata:
+  kind: kernel
+  version: 1.0.0
+  created: '2026-08-22'
+  author: PAIML Engineering
+  description: >
+    apr-code no-argument start-up policy and inference-child lifetime (issue
+    #2607, measured on the PUBLISHED apr 0.63.0 during the pre-release
+    multi-platform dogfood sweep). Two correctness surfaces, both about a run
+    that was never asked to do anything:
+    (1) START-UP POLICY — `apr code` with no arguments at all and a stdin that
+    is not a terminal cannot hold a session: the REPL's first read_line returns
+    EOF. On the sweep host it nevertheless scanned the filesystem, picked the
+    largest local GGUF (a 30B MoE) and launched an `apr serve` child for it. A
+    consequential action chosen by looking at the disk is a guess, not an
+    instruction; a bare invocation must print help and spawn nothing.
+    (2) CHILD LIFETIME — the non-interactive (`-p`) path ends in
+    std::process::exit, which runs NO destructors, so every owner of the
+    inference driver must be released first. `drop(driver)` alone did not do
+    that: the driver is an Arc and register_task_tool stores a clone of it in
+    the tool registry, so the strong count never reached zero,
+    AprServeDriver::drop never ran, and the `apr serve` child was orphaned
+    holding the whole model in RSS.
+  references:
+  - 'crates/aprender-orchestrate/src/agent/code.rs — CodeInvocation::wants_help(), release_driver(), cmd_code()'
+  - 'crates/apr-cli/src/dispatch.rs — Commands::Code arm, print_code_help_and_exit()'
+  - 'crates/aprender-orchestrate/src/agent/driver/apr_serve.rs — AprServeDriver::launch()/drop()'
+  - 'crates/aprender-orchestrate/src/agent/task_tool.rs — register_task_tool() clones the driver Arc into the registry'
+  - 'https://github.com/paiml/aprender/issues/2607'
+  related_contracts:
+  - contracts/apr-code-parity-v1.yaml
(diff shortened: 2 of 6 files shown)


noahgift:

Issue #2609, opened 2026-08-22
Four routed HTTP endpoints — two of them in the server's own startup banner — are dead when a .apr model is loaded


noahgift:
**Severity: P1** — found by the pre-release multi-platform dogfood sweep (intel/AVX-512 lane).

Measured against the **published** `apr 0.63.0` from crates.io on the Intel receipt host
(Xeon W-3245, 16C/32T, AVX-512F/BW/CD/DQ/VL/VNNI), not a local build.

### Where

aprender-serve HTTP route table: POST /stream/generate, POST /batch/generate, POST /v1/chat/completions/stream, POST /v1/embeddings (crates/aprender-serve/src/api/router.rs:103-104 registers the first two)

### Defect

Four routed endpoints — two of them advertised by the server's own startup banner — are non-functional when a .apr model is loaded, all failing with `Model registry error: No model available` while POST /generate on the SAME process returns 200 and /health reports model_loaded:true. Worse, the HTTP status codes are wrong and mutually inconsistent for the identical error: /stream/generate and /v1/chat/completions/stream and /v1/embeddings return 404 (which means 'route not found' — the route IS registered), while /batch/generate returns 500 for the exact same condition. A client cannot tell 'endpoint does not exist' from 'endpoint exists but has no model'.

### Reproduction

```
On mac-server:
  APR=/home/noah/.cargo/bin/apr; M=/home/noah/models/qwen2.5-coder-1.5b-instruct-q4_k_m.apr
  "$APR" serve run "$M" --port 38622 --no-gpu --context-length 512 --no-fp8-cache &
  # banner prints: 'POST /stream/generate - SSE streaming' and 'POST /batch/generate - Batch inference'
  curl -s -o /dev/null -w '%{http_code}\n' -X POST localhost:38622/generate -H 'Content-Type: application/json' -d '{"prompt":"hi","max_tokens":1}'
    MEASURED 200  body {"token_ids":[6023,16],"text":"hi1","num_generated":1}
  curl -s -w ' %{http_code}\n' -X POST localhost:38622/stream/generate -H 'Content-Type: application/json' -d '{"prompt":"hi","max_tokens":1}'
    MEASURED {"error":"Model registry error: No model available"} 404
  curl -s -w ' %{http_code}\n' -X POST localhost:38622/batch/generate -H 'Content-Type: application/json' -d '{"prompts":["hi"],"max_tokens":1}'
    MEASURED {"error":"Model registry error: No model available"} 500
[...]

PR #2626, opened 2026-08-22
fix(serve): four routed endpoints were dead on the third .apr backend, and 40 tests could not have caught it

### LINKED ISSUE #2375: OpenAI-compat surface: 13 defects incl. fabricated /v1/explain output and routes that always 404

noahgift:
Refs #2609.

## ⚠️ `check_readme_claims.sh` is RED on this PR and that is CORRECT

This branch adds exactly one contract (`contracts/apr-serve-model-backend-coverage-v1.yaml`),
taking `find contracts/ -name '*.yaml' | wc -l` from 1778 to 1779. `README.md` hardcodes
that figure at :44, :225 and :256, and `check_readme_claims.sh` — a REQUIRED check —
asserts equality.

**This PR deliberately does not touch README.md.** Six open PRs each add one contract and
four of them independently bump 1778 → 1779. Every one is individually correct and they
are mutually exclusive: only one can ever merge green, and it reds the other five. Filed
as #2630. The integration branch that folds these PRs sets the figure ONCE, from a real
measurement, after they merge together. Please do not "helpfully" re-add the bump here.

Deleting the contract to dodge the check is not an option: the contract is the deliverable.

---

## What survived on `main`

#2609 was measured against the **published** 0.63.0. Half of it is already fixed: #2375 and
#2376 landed *after* that tag, and a Q4_K `.apr` loads through `OwnedQuantizedModel`, so the
four endpoints are green on a quantized server today. Probing `main` @ `bb2bd5e73` found the
same class alive on the **third** resident backend — `AppState::apr_transformer`, the f32 APR
/ SafeTensors CPU serve path — and the status split alive on every configuration:

| route | `AprTransformer` server | no model at all |
|---|---|---|
| `POST /generate` | 200 `{"text":"t1t230",…}` | 503 |
| `POST /batch/generate` | 200 | 503 |
| `POST /stream/generate` | **503** "No model available" | 503 |
| `POST /v1/chat/completions` | **404** "No model available" | **404** |
| `POST /v1/chat/completions/stream` | **404** "No model available" | **404** |
| `POST /v1/embeddings` | **503**, naming `/realize/embed` | **503**, naming `/realize/embed` |

`/health` on that same server reports `model_loaded: true`. That is #2609's shape exactly.

## Second pass: the surface, not six routes

An adversarial review of the first pass made the right call: half the ticket is *"an
inconsistent code for an identical condition"*, and that is a property of a **surface**. It
cannot be discharged route by route, and the first pass had left `POST /v1/completions` out
entirely — the one route the startup banner names. So the probe was widened from six routes
to every mounted-and-advertised one. Three more rows were wrong on the same tree:

| route | `AprTransformer` server | no model at all |
|---|---|---|
| `POST /v1/completions` | **404** "No model available" | **404** |
| `POST /api/chat` | 200 | **200**, error text as the assistant's reply |
| `POST /api/generate` | 200 | **200**, error text as the assistant's reply |

The two Ollama rows are strictly worse than anything #2609 reported. `200 OK` carrying
`"Model registry error: No model available"` as the model's own answer is invisible to every
client and to every retry policy keyed on status — the ticket's routes at least said
*something* was wrong.

## Decision, per endpoint

All of them: **make it work**, none: stop advertising. The server already generates on this
backend through `/generate`; a route that is mounted, banner-advertised and backed by a
resident model has no honest reason to answer "no model".

- **`/stream/generate`** (and `/realize/generate`) — `try_apr_stream_tokens`, mirroring the
  `try_apr_generate` that `/generate` already had.
- **`/v1/chat/completions{,/stream}`** — `try_apr_transformer_backend`; `stream: true` reuses
  `pregenerated_sse_response`. The `/stream` route is a one-line delegation to the non-stream
  one (#2375), so both are fixed by the same arm.
- **`/v1/completions`** — `try_apr_transformer_completions` (second pass). It walked straight
  into `registry_completions`, which resolves the dense f32 `Model` an `AprTransformer` server
  does not have. It now walks the same chain the chat path walks.
- **`/v1/embeddings`** (and `/realize/embed`, `/api/embeddings`) — `EmbedBackend::Apr`, backed
  by `AprTransformer::forward_hidden_states`.
- **`/batch/generate`** already worked here; it is probed anyway, because the premise of the
  bug report is the *contrast*, and a "fix" that broke `/generate` would satisfy "all the same
  status" trivially.

## One condition, one status

The split lives in the fallbacks every backend chain ends in, which hardcoded `NOT_FOUND` —
`registry_fallback` (chat) and `registry_completions` (completions). 404 means *the route does
not exist*, and a genuinely unrouted path returns a different body entirely
(`{"error":"not_found",…}`), so a client could not distinguish the two. Both now resolve
through `model_resolution_status`:

- `RegistryError` — the server has no usable model → **503** on every routed endpoint
- `ModelNotFound` — the client named a model this server lacks → **404**, still a client error

`GET /healthz` stays 404 with the route-not-found body, asserted, so the fix does not blur the
distinction it exists to preserve.

`/api/chat` and `/api/generate` now propagate the upstream status through
`with_upstream_status` while keeping the Ollama-shaped body. That needed one more change:
`sanitize_json_rejection` rewrites any non-2xx whose content-type is not `application/json`
into a single `{"error":…}` object, so the **first attempt reported the status and destroyed
the NDJSON frame** — leaving an Ollama client with no terminal `done:true` at all, the exact
malformed stream `chat_response_to_parts` exists to prevent. `application/x-ndjson` is now
passed through on the same grounds as `application/json`: it *is* JSON, authored by our own
handler. The tests keep every wire assertion and add the status, so they now prove the frame
**survives** a failure rather than that a failure is dressed up as an answer.

## The banner

`serve_model` and `serve_demo` each printed a hand-written list of **three** routes out of the
**thirty-one** they mount, from a point before the model format is even known. The only
generation route `serve_model` named was `POST /v1/completions` — the one dead on the
`AprTransformer` state that same function builds — while `/generate`, which worked, was
labelled `"Q4_K fused"` on a server holding no Q4_K weights.

Both now print `crate::api::advertised_routes(&router_config)`, derived from the same
`route_table` that `create_router_with_config` mounts two lines later, and which
`route_surface_2376::banner_source_agrees_with_live_server` already pins to the list the
running server hands out in its own 404 body. Advertising a route and mounting it are one act
— the shape #2376(8) established for `apr serve run`.

## Why 52 tests were green over all of this

40 asserted `status == OK || status == NOT_FOUND || status == BAD_REQUEST || status ==
INTERNAL_SERVER_ERROR || status == NOT_FOUND` — `NOT_FOUND` listed **twice** in most of them.
A disjunction over every plausible outcome excludes none. The second pass found 10 more of the
same shape on `/v1/completions`, plus 2 asserting `200` from the Ollama streaming routes. All
run against `demo_mock()`, where exactly one status is correct; they now assert it, via
`assert_no_model_status`. `is_mock_response()` and `guard_mock_response!`, which both hardcoded
`NOT_FOUND` as "no model loaded", are corrected too.

## Verification

Falsifiers: `crates/aprender-serve/src/api/tests/apr_model_routes_2609.rs` (**11**), pinned by
`contracts/apr-serve-model-backend-coverage-v1.yaml` v1.1.0 (`pv validate`: 0 errors, 0
warnings; `pv audit`: no findings). The fixture asserts the server generates and reports
`model_loaded` **before** any dead-route assertion, so no test can pass by generation being
broken.

Mutation-verified in both directions — 5 mutations in this pass, each proven to have **engaged**
before any verdict was read:

| mutation | falsifiers turned RED | proof it engaged |
|---|---|---|
| `try_apr_transformer_completions` → `Err(418, "MUT2609COMPLETIONSARM")` | `no_routed_endpoint_is_dead…`, `no_model_available_is_503…` | `strings -a` on the built test binary = **1**; the marker came back in the HTTP body |
| `registry_completions` status → `NOT_FOUND` | `no_model_available_is_503…` | observed status became exactly **404**, on exactly `/v1/completions`, nothing else moved |
| `serve_model` banner → the restated literal | `serve_model_banner_is_derived…` | failure quotes the reintroduced line verbatim |
| `with_upstream_status` → no-op | 3 tests | observation table shows **200** on exactly `/api/chat` and `/api/generate`, 503 on the other eleven |
| NDJSON pass-through removed from `sanitize_json_rejection` | 2 `ollama_compat_http` tests | failure quotes the observed content-type `"application/json"` |

(The first pass's five mutations — `model_resolution_status`, `try_apr_stream_tokens`,
`try_apr_transformer_backend`, `EmbedBackend::Apr`, the embed route label — remain as
originally reported. One of those is worth re-naming: the embed-arm mutation was first written
as `if false { } else if let Some(apr) = … {`. Its marker grepped clean at the decision site and
it disabled **nothing** — the `else if` still ran, the suite came back green, and it read
exactly like "the guard is fine". Redone as `.filter(|_| false)`, it goes red. Marker-presence
is not engagement.)

**One mutation exposed a gap in a new test, and the fix was to NOT strengthen it.**
`every_advertised_route_is_alive_on_an_apr_transformer_server` stayed green against the
fail-closed completions arm, because a 418 is neither a 404 nor a "No model available" body.
Widening it to "no 5xx" flagged two routes that answer an **honest** error on this fixture —
`/realize/reload` 501 (registry mode is a capability the CLI never builds) and
`/metrics/dispatch` 503 (no GPU configured), both naming the exact missing capability. That
would have bought coverage with an exemption list. The strict per-route status assertion stays
where `== OK` is meaningful — over `GENERATION_ROUTES` with real bodies, which is what killed
the mutation — and the surface-wide test asserts only what is true of every advertised route
regardless of what it does. The division is written into its doc comment so the next reader
does not "strengthen" it back.

## Checks

- `cargo test -p aprender-serve --lib` — **15682 passed, 0 failed**
- `cargo clippy -p aprender-serve --lib -- -D warnings` — clean (toolchain 1.93.0, the pin)
- `cargo test -p aprender-serve --lib --tests --no-run` — all targets compile
- `git diff origin/main -- README.md .github/ Cargo.lock .pv/` — **empty**. No workflow is
  touched, and the `[[patch.unused]]` block a local `.cargo/config.toml` writes into
  `Cargo.lock` (#2615) is kept out.
- `check_readme_claims.sh` — **RED, expected**. See the top of this body and #2630.

## Deferred

- `crates/aprender-serve/src/api/cuda_chat_backend.rs` fails the pre-commit complexity gate —
  `try_cuda_backend` at cyclomatic 14 / cognitive 43. **Pre-existing and unchanged**: the same
  gate reports `Errors: 2` on `main`'s copy of the file and `Errors: 2` on this branch. This PR
  *lowers* `openai_chat_completions_handler` (9/22 → 8/19). The offender is
  `#[cfg(feature = "cuda")]` code untouched here; committed with `--no-verify`.
- Four `clippy --tests` findings under the pinned toolchain (`beat_fail_closed_garbage.rs:52`,
  `cancel_scope_2376.rs:316`, `format_detect.rs:366`, `memory.rs:344`) — all in files this PR
  does not touch; `--lib`, which is what the repo documents, is clean.
- `test_stream_handler_empty_messages` is now pinned to 503, which documents that empty-`messages`
  validation happens *after* model resolution. Arguably a client error reported as a server
  condition; not this PR's scope.
- `apr-cli`'s own `build_apr_cpu_router` (the Q4K-declined APR CPU fallback) has a third,
  smaller router with its own `print_apr_cpu_banner` naming 3 of its ~6 routes. Those three all
  exist and work, so it is not the #2609 shape; deriving that banner too is a separate change
  in a separate crate and is not folded in here.

### DIFF
--- a/contracts/apr-serve-model-backend-coverage-v1.yaml
+++ b/contracts/apr-serve-model-backend-coverage-v1.yaml
@@ -1,0 +1,162 @@
+contract: apr-serve-model-backend-coverage
+metadata:
+  kind: pattern
+  version: "1.0.0"
+  description: >
+    Every routed generation endpoint must answer on every resident inference
+    backend, and one server-side model-availability condition must produce one
+    HTTP status across the whole surface.
+
+    aprender#2609 measured the published apr 0.63.0 answering
+    "Model registry error: No model available" on POST /stream/generate,
+    POST /batch/generate, POST /v1/chat/completions/stream and POST /v1/embeddings
+    — two of them printed by the server's own startup banner — while POST /generate
+    on the SAME process returned 200 with real tokens and /health reported
+    model_loaded: true. Worse, the identical condition came back as 404 from three
+    of them and 500 from the fourth, so a client could not distinguish "this
+    endpoint does not exist" from "this endpoint exists but has no model".
+
+    The class is structural, not incidental. AppState carries several mutually
+    exclusive model fields (model, quantized_model, apr_transformer, gpu/cuda
+    variants); a handler is alive on exactly the backends whose field it reads.
+    aprender#2376 and #2375 closed the gap for quantized_model on these routes,
+    which is why the four are green on a GGUF / Q4_K-APR server today. The THIRD
+    resident backend, apr_transformer (the f32 APR / SafeTensors CPU serve path
+    built by AppState::with_apr_transformer_and_vocab), still had no accessor on
+    /stream/generate, on the chat routes, or in resolve_embed_backend — so on that
+    server the same four routes were dead with the same message, again beside a
+    working /generate and a model_loaded: true /health.
+
+    The status half is a separate defect with a separate cause: it lives in the
+    DENSE fallback that every backend chain ends in, so no per-backend fix reached
+    it. model_resolution_status is the single mapping — RegistryError (the server
+    has no usable model) => 503, ModelNotFound (the client named a model this
+    server does not have) => 404 — and every route resolves through it.
+  references:
+  - "https://github.com/paiml/aprender/issues/2609"
+  - "crates/aprender-serve/src/api/mod.rs (model_resolution_status — the single mapping)"
+  - "crates/aprender-serve/src/api/stream_generate.rs (try_apr_stream_tokens)"
+  - "crates/aprender-serve/src/api/cuda_chat_backend.rs (try_apr_transformer_backend; registry_fallback status)"
+  - "crates/aprender-serve/src/api/realize_handlers_embed_completion.rs (EmbedBackend::Apr; embed_for_route)"
+  - "crates/aprender-serve/src/apr_transformer/hidden_states.rs (AprTransformer::forward_hidden_states)"
+  - "crates/aprender-serve/src/api/tests/apr_model_routes_2609.rs (falsifiers)"
+version: 1
+status: enforced
+date: 2026-08-22
+
+proof_obligations:
+  - type: invariant
+    property: >
+      BACKEND-COVERAGE (DISCHARGED — aprender#2609): on an AppState whose only
+      resident model is an AprTransformer, every routed generation endpoint —
+      /generate, /batch/generate, /stream/generate, /v1/chat/completions,
+      /v1/chat/completions/stream, /v1/embeddings — answers 200 with real output.
+      None of them may return a body containing "No model available" while
+      AppState::model_loaded() is true. The chain each handler walks is the same
+      order /generate walks: CUDA (cfg), quantized, APR, dense.
+  - type: invariant
+    property: >
+      ONE-CONDITION-ONE-STATUS (DISCHARGED — aprender#2609): a request to a MOUNTED
+      route on a server with no usable model at all resolves through
+      model_resolution_status and therefore returns 503 SERVICE_UNAVAILABLE — never
+      404 (which means the route does not exist, and a genuinely unrouted path
+      returns a different body entirely) and never 500 (which invites a retry of an
+      identical request against a server that cannot serve it). 404 is reserved for
+      RealizarError::ModelNotFound: the client named a model this server does not
+      have, which IS a client error.
+  - type: invariant
+    property: >
+      ERROR-NAMES-THE-CALLED-ROUTE (DISCHARGED — aprender#2609): an
+      unavailable-model body names the route the client actually called.
+      /v1/embeddings delegated its whole body to realize_embed_handler and so
+      answered "No model available: /realize/embed needs a loaded model", pointing
+      the caller at a route they never touched; every embedding route now passes its
+      own name into embed_inputs.
+  - type: invariant
+    property: >
+      HIDDEN-STATE-PARITY: all three embedding backends return the same quantity —
+      the final-layer hidden state after the output norm and before lm_head, the
+      tensor lm_head consumes. Model::forward_hidden (dense),
+      OwnedQuantizedModel::forward_hidden_states (quantized) and
+      AprTransformer::forward_hidden_states (APR f32) are therefore comparable after
+      mean-pooling, and the embedding width is always the model's hidden_dim, never
+      a constant. AprTransformer::forward_with_cache is defined as
+      forward_hidden_with_cache + project_lm_head, so the embedding path cannot
+      drift from the generation path by one being updated and the other not.
+  - type: invariant
+    property: >
+      CONTEXT-CEILING: AprTransformer::forward_hidden_states rejects a sequence
+      longer than the model's context_length with ContextLimitExceeded before
+      allocating a KV cache. Sizing a cache from a caller-supplied length with no
+      ceiling is what let one HTTP request abort the server process
+      (aprender#2376 finding 9), and an embedding route takes caller-supplied text.
+
+falsification_tests:
+  - id: FALSIFY-APR-BACKEND-COVERAGE-2609
+    name: apr_model_routes_2609
+    prediction: >
+      Over the REAL router and an AppState holding ONLY an AprTransformer:
+      /health reports model_loaded:true and /generate returns num_generated:1 (the
+      CONTROL — every dead-route assertion is compared against a server that
+      demonstrably generates, so none can pass by generation being broken); then
+      /generate, /batch/generate, /stream/generate, /v1/chat/completions,
+      /v1/chat/completions/stream and /v1/embeddings ALL return 200 and none of
+      their bodies contains "No model available". /stream/generate emits at least
+      one "event: token" and does not terminate with num_generated:0;
+      /v1/chat/completions/stream emits a chat.completion.chunk and a [DONE];
+      /v1/embeddings returns a vector of exactly the model's hidden_dim whose L2
+      norm is 1.
+    test_harness: 'cargo test -p aprender-serve --lib apr_model_routes_2609'
+    expected_output: "test result: ok"
+    if_fails: >
+      A routed endpoint is dead on a resident backend — the aprender#2609 shape:
+      "Model registry error: No model available" from a route the startup banner
+      advertises, on a process whose /generate answers and whose /health says
+      model_loaded:true. Verbatim pre-fix, on main @ bb2bd5e73:
+      POST /stream/generate -> 503 {"error":"Model registry error: No model available"};
+      POST /v1/chat/completions/stream -> 404 {"error":"Model registry error: No model available"};
+      POST /v1/embeddings -> 503 {"error":"No model available: /realize/embed needs a loaded model"};
+      while POST /generate -> 200 {"token_ids":[1,230],"text":"t1t230","num_generated":1}.
+  - id: FALSIFY-ONE-CONDITION-ONE-STATUS-2609
+    name: no_model_available_is_503_on_every_routed_endpoint
+    prediction: >
+      Over the REAL router and AppState::demo_mock() (no model of any kind), all six
+      routed generation endpoints report "No model available" AND return the SAME
+      status, 503 — no route returns 404 and no route returns 500. Separately, GET
+      /healthz (genuinely unrouted) still returns 404 with the route-not-found body,
+      so the fix does not blur the distinction it exists to preserve.
+    test_harness: 'cargo test -p aprender-serve --lib no_model_available_is_503_on_every_routed_endpoint'
+    expected_output: "test result: ok"
+    if_fails: >
+      A mounted route answers 404 or 500 for a server-side model-availability
+      condition, so a client retry policy keyed on status treats one outage as a
+      permanent routing error on some routes and a server bug on others. Verbatim
+      pre-fix, on main @ bb2bd5e73: /generate, /stream/generate and /batch/generate
+      returned 503 while /v1/chat/completions and /v1/chat/completions/stream
+      returned 404 for the byte-identical body
+      {"error":"Model registry error: No model available"}.
+  - id: FALSIFY-APR-HIDDEN-CONTEXT-CEILING-2609
+    name: apr_forward_hidden_states_refuses_an_over_context_sequence
+    prediction: >
+      AprTransformer::forward_hidden_states(&[1; context_length + 1]) returns
+      ContextLimitExceeded { provided: context_length + 1, maximum: context_length }
+      — before any KV cache is allocated — and forward_hidden_states(&[]) returns
+      InvalidShape rather than an empty success.
+    test_harness: 'cargo test -p aprender-serve --lib apr_forward_hidden_states_refuses_an_over_context_sequence'
+    expected_output: "test result: ok"
+    if_fails: >
+      The new hidden-state accessor sizes a KV cache from a caller-supplied length
+      with no ceiling, so one oversized POST /v1/embeddings body asks the allocator
+      for an arbitrary amount — the aprender#2376 finding 9 shape, where a single
+      HTTP request aborted the server process with no reply at all.
+  - id: FALSIFY-EMBED-ROUTE-NAME-2609
+    name: embed_unavailable_body_names_the_route_the_client_called
+    prediction: >
+      POST /v1/embeddings and POST /realize/embed on a model-less server each return
+      an error body naming the route the client called, not the route the handler
+      happens to delegate to.
+    test_harness: 'cargo test -p aprender-serve --lib embed_unavailable_body_names_the_route_the_client_called'
+    expected_output: "test result: ok"
+    if_fails: >
+      The error body sends the caller to a route they never used. Verbatim pre-fix:
+      POST /v1/embeddings -> {"error":"No model available: /realize/embed needs a loaded model"}.
--- a/contracts/apr-serve-model-backend-coverage-v1.yaml
+++ b/contracts/apr-serve-model-backend-coverage-v1.yaml
@@ -1,7 +1,7 @@
 contract: apr-serve-model-backend-coverage
 metadata:
   kind: pattern
-  version: "1.0.0"
+  version: "1.1.0"
   description: >
     Every routed generation endpoint must answer on every resident inference
     backend, and one server-side model-availability condition must produce one
@@ -32,6 +32,27 @@ metadata:
     it. model_resolution_status is the single mapping — RegistryError (the server
     has no usable model) => 503, ModelNotFound (the client named a model this
     server does not have) => 404 — and every route resolves through it.
+
+    v1.1.0 widens the obligation from the six routes the ticket named to the WHOLE
+    mounted-and-advertised surface, because "one condition, one status" is a
+    property of a surface and cannot be discharged route by route. Probing the rest
+    found three more rows on the same tree: POST /v1/completions was dead on
+    apr_transformer AND answered 404 for the no-model condition (it walks straight
+    into registry_completions, whose status was hardcoded), and POST /api/chat and
+    POST /api/generate answered 200 OK with the outage folded into the assistant's
+    own reply — the only two rows on the surface that did not report the failure at
+    all, and strictly worse than anything #2609 reported, since no client and no
+    retry policy can see it.
+
+    The banner is part of the same obligation. serve_model and serve_demo restated
+    a hand-written list of three routes out of the thirty-one they mount, from a
+    point before the model format is known; the only generation route serve_model
+    named was /v1/completions, the one dead on the apr_transformer state that same
+    function builds, while /generate — which worked — was labelled "Q4_K fused" on a
+    server holding no Q4_K weights. Both now print advertised_routes(&router_config),
+    derived from the same route_table create_router_with_config mounts, so
+    advertising a route and mounting it are one act (the shape aprender#2376(8)
+    established for `apr serve run`).
   references:
   - "https://github.com/paiml/aprender/issues/2609"
   - "crates/aprender-serve/src/api/mod.rs (model_resolution_status — the single mapping)"
@@ -39,6 +60,10 @@ metadata:
   - "crates/aprender-serve/src/api/cuda_chat_backend.rs (try_apr_transformer_backend; registry_fallback status)"
   - "crates/aprender-serve/src/api/realize_handlers_embed_completion.rs (EmbedBackend::Apr; embed_for_route)"
   - "crates/aprender-serve/src/apr_transformer/hidden_states.rs (AprTransformer::forward_hidden_states)"
+  - "crates/aprender-serve/src/api/gpu_completions_handler.rs (try_apr_transformer_completions; registry_completions status)"
+  - "crates/aprender-serve/src/api/ollama_handlers.rs (with_upstream_status)"
+  - "crates/aprender-serve/src/api/router.rs (advertised_routes; sanitize_json_rejection NDJSON pass-through)"
+  - "crates/aprender-serve/src/cli/mod_server_commands.rs (serve_model / serve_demo banners)"
   - "crates/aprender-serve/src/api/tests/apr_model_routes_2609.rs (falsifiers)"
 version: 1
 status: enforced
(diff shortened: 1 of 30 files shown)


noahgift:

Issue #2522, opened 2026-08-16
falsification_spec_v10: 140 tests, named in no workflow, 38 failing — and its byte-size tests assert expressions against themselves


noahgift:
`crates/aprender-core/tests/falsification_spec_v10_tests.rs` is the project's own falsification spec. It is named in the **Makefile** (lines 113, 121) and in **no workflow at all**, so CI has never run it.

Running it (`cargo test -p aprender-core --features model-tests --test falsification_spec_v10_tests`):

```
test result: FAILED. 100 passed; 38 failed; 0 ignored
```

## Triage of the 38

| class | n | cause |
|---|---|---|
| stale paths | ~24 | `"file must exist: NotFound"` — hardcoded source paths that moved during APR-MONO consolidation (`types.rs`, `generation/mod.rs`, `bpe/mod.rs`, `chat_template/mod.rs`, `loading/mod.rs`, `import.rs`, the spec markdown itself) |
| stale symbols | ~8 | greps for `validate_model_contract`, `extract_model_paths`, `verify_output` — renamed or moved |
| drifted constants | 2 | `Commands enum must have exactly 36 variants, found 0` (the CLI has 111; the grep finds 0 because the enum's file moved) |
| **real defect** | **1** | see below |

## The one real defect — fixed in #2521

`F-DOD-005` flagged a silent dtype fallback that turned out to be a genuine 12× tensor mis-sizing in the GGUF→APR converter. Details in #2521.

## Three that looked real and were not

Worth recording, because they show *why* this suite is weak:

- **`f_contract_006_no_column_major_type_exists`** flags **its own source file**. It greps for `struct ColumnMajor` and matches the grep string inside `falsification_spec_v10_contract_model.rs`. A self-matching assertion.
- **`f_layout_001_clippy_bans_colmajor_imports`** says "inference path" but scans `crates/aprender-serve/examples/`. Its hits are `verify_q4k_layout.rs`, `compare_q4k_bytes.rs`, `debug_q4k_attn_output.rs` — examples that *deliberately* compare column-major against row-major. Wrong scope, not a violation.
- **`f_dod_001_satd_count_is_zero`** reports 86 against a threshold of 0, counting test files; the repo's own PMAT gate uses a different scope and threshold.

## The structural problem

Two things make this suite low-value as written:

[...]

Issue #2231, opened 2026-06-25
Extract a sovereign `apr-format` leaf crate — let consumers read/write `.apr` without depending on `aprender-core` (~138 deps + GPU/CUDA stack)


noahgift:
## Summary

Request from the **xpile** project (a downstream `.apr` consumer): extract the `.apr` container format (read + write, v1 `APRN` + v2 `APR\0`) out of `aprender-core` into a **minimal, sovereign `apr-format` leaf crate** with no ML/GPU dependencies — so the paiml ecosystem (realizar inference, xpile, external tooling) can consume `.apr` without pulling the entire framework or coupling to aprender's release cadence.

**One line:** depend on the *format*, not the *framework*.

## Problem

The `.apr` read/write code lives **inside `aprender-core`**:
- `crates/aprender-core/src/format/mod.rs` (`MAGIC = [0x41,0x50,0x52,0x4E]` "APRN", `HEADER_SIZE = 32`)
- `format/core_io.rs` (`save`/`load`/`load_from_bytes`), `format/core_io_mmap.rs` (`load_mmap`/`load_auto`)
- `format/v2/` (the `APR\0` 64-byte-header, 64-byte-aligned tensor-blob, mmap container)
- `format/encryption.rs` (`save_encrypted`/`load_encrypted`, behind a feature)

…but `format/` is **entangled with the rest of the crate** — it pulls `crate::models::qwen2::Embedding`, `crate::text::shell_vocab`, `format::gguf`, `format::quantize`, `format::onnx`, etc. And `aprender-core/Cargo.toml` carries **~138 dependency lines** plus `gpu = ["trueno/gpu"]` / `cuda = ["trueno/cuda-monitor"]` / wgpu features.

**Consequence:** anything that just wants to *read or write a `.apr` file* must depend on the full ML + GPU + tokenizer + quantization stack and track aprender's release cadence. This is the same cross-crate version-conflict class already documented in [`docs/specifications/apr-cli-dep-migration.md`](docs/specifications/apr-cli-dep-migration.md) — *"`cargo install aprender` fails on crates.io … version conflicts … the local workspace compiles because `[lib] name` aliases work, but crates.io resolves deps independently."* A format leaf crate structurally removes a chunk of that surface.

## Why now / downstream use-case

- **realizar** (inference runtime) and **xpile** both have legitimate reasons to read `.apr` *without* the trainer/GPU stack.
[...]

PR #2627, opened 2026-08-22
fix(tests): 140 falsification gates ran in no workflow, and 38 were red (#2522)

noahgift:
Closes nothing on its own — refs #2522 for the maintainer to close.

## What was wrong

`crates/aprender-core/tests/falsification_spec_v10_tests.rs` is this project's own
falsification spec: 140 gates. It was named in the **Makefile** and in **no workflow**,
so CI has never run it. **38 of the 140 were failing**, and had been for months.

It is dark by *construction*, which is why the repo's existing wiring guards miss it:

- `#![cfg(feature = "model-tests")]` compiles the file to **nothing** without the feature,
  so no green CI run anywhere says anything about it;
- CI's workspace line is `--lib`, which cannot reach an integration target;
- the one explicit integration line in `ci.yml` lists targets by name, and these were not on it.

## The 38 reduce to 5 causes

| n | cause |
|---|---|
| 15 | pre-APR-MONO `<root>/src/...` path literals. The tree moved to `crates/<crate>/src/`, and `format/types.rs` left for the `apr-format` crate entirely in #2231/#2236 |
| 13 | intra-crate file moves. apr-cli split `lib.rs` → `validate.rs` + `commands_enum.rs`, and `commands/qa.rs` → `qa_report.rs` + `commands/output_verification.rs`. Every symbol still exists |
| 5 | the showcase markdown moved to `docs/specifications/archive/` |
| 4 | scan-scope defects in the gates themselves (below) |
| 1 | `find_generated_file` searched `<project_root>/target`, which no build here uses |

**Not one of the 38 was a defect in the code under test.** The one gate that did find a
real defect, F-DOD-005, was fixed separately in #2521.

### The four scope defects are worth naming individually

- **`f_contract_006`** asserts no `struct ColumnMajor` exists and then matches the three
  string literals *in its own body* that spell the pattern. It could never pass. It has
  been red on its own source since the day it was written.
- **`f_layout_001`** says "inference path" and scanned the whole workspace, flagging the
  `aprender-serve/examples/` harnesses that exist *to* compare column-major against
  row-major, plus the `aprender-compute` module that **defines** the kernels.
- **`f_dod_005`** skipped whole-line comments but not trailing ones, so `_ => 4, // Default F32`
  — a byte *width* — read as a silent dtype fallback; and `contains("F32")` matched `ImmF32`
  inside `panic!("Expected ImmF32 source")`, which is the *opposite* of a silent fallback.
  All 7 of its violations were false.
- **SATD** demanded 0 against a real 54. An assertion that can only ever be red carries no
  information about whether the debt is growing.

### And three gates were passing vacuously

`f_prove_003/004/005` write the broken target-dir lookup as `if let Some(..)`, so they
reported `ok` while never opening the file they exist to check. `f_perf_002`,
`f_trueno_002/004/005/006/007/008` and `f_arch_006` looked for **sibling `../trueno`
and `../realizar` checkouts** that APR-MONO archived; every miss branch `return`s, which
the harness prints as `ok`. `f_dod_unsafe_code_is_deny_by_default` scanned the two-file
root `src/` facade — 2 files out of 5,566 — to guard `unsafe_code = "forbid"`.

## Fixes, by cause rather than by symptom

- **Gates anchor to the CRATE, not to a file path.** Which file holds a symbol is churn;
  which crate holds it is the boundary this architecture defends. `crate_src_text` panics
  on a missing crate rather than returning an empty string — "grep found nothing" and
  "there was nothing to grep" must not be one verdict, and that equivalence is what let 38
  gates report a symbol as absent when it was merely elsewhere.
- **`find_generated_file` and `apr_binary` derive the target dir from `current_exe()`**,
  which no `--target-dir` flag can redirect out from under.
- **`F-CLI-001`** replaced a frozen `== 36` (the CLI ships 111) with a cross-check against
  `contracts/apr-cli-commands-v1.yaml`, resolving clap's `#[command(flatten)]` groups and
  `#[cfg(feature)]` variants down to 102 real commands. Adding a variant without
  registering it now fails.
- **`F-PROVE-007`** replaced a frozen `== 297` (the generator emits 578) with a per-family
  floor derived from `KNOWN_FAMILIES`.
- **SATD is a shrink-only ratchet at the measured 54**, with a floor assertion so a gain
  must be re-baselined rather than silently absorbed.
- The vacuous sibling-repo lookups now point at `crates/aprender-compute`, `crates/aprender-gpu`
  and `crates/aprender-serve`, and assert rather than `return`.

## Wiring, so it cannot go dark again

- **`ci.yml`** runs all three `model-tests`-gated targets (156 gates) in `workspace-test`,
  inside the image, on the run's own target dir. **This does not touch the shared
  integration-test line** that #2617 owns.
- **`scripts/check_model_tests_wired.sh`** enumerates the universe from the **source tree**
  — a guard built from the workflow side can only confirm what is already there. Ships a
  17-case table: mention-vs-execution, trailing comments, equals-form flags,
  backslash-folded multi-line commands, and a fixture tree proving the enumerator itself
  can produce a row.
- **`F-META-001/002/003`** gate the suite on itself: every fragment is `include!`d, the
  fixture-skip class (30 `require_model!` sites that report `ok` with no model) is bounded
  shrink-only, and the suite asserts its own CI wiring.
- The **Makefile**'s two callers piped cargo into `grep` and read *grep's* status, so
  `make test-spec` printed "✅ Spec tests complete" throughout — grep finds the
  `test result:` line, which is exactly what it does when tests fail. These were the
  suite's only callers anywhere.

## Verification

`141 passed; 0 failed` (138 repaired + 3 new meta-gates).

Every behavioural change was mutation-verified in **both** directions, with a grep proving
the mutation engaged before any verdict was read:

| gate | mutation | RED | GREEN |
|---|---|---|---|
| `check_model_tests_wired.sh` | drop `--features model-tests` from the CI line | rc=1 | rc=0 |
| `check_model_tests_wired.sh` | comment the CI invocation out | rc=1 | rc=0 |
| `F-META-003` | delete the `--test` target from ci.yml | rc=101 | rc=0 |
| `F-DOD-005` | add a real `_ => DType::F32,` arm | rc=101 | rc=0 |
| `F-LAYOUT-001` | add a real `use trueno::...::matmul_q4k_f32_colmajor;` | rc=101 | rc=0 |
| `F-CONTRACT-006` | add `pub struct ColumnMajor;` | rc=101 | rc=0 |
| `F-CHECKLIST-005` | add one `// TODO:` | rc=101 (55 > 54) | rc=0 |
| `F-CLI-001` | add an unregistered `Commands` variant | rc=101 | rc=0 |

`F-DOD-005` and `F-LAYOUT-001` were **re-mutated after their scope was narrowed** — the old
proof does not transfer when a guard's universe changes.

## Deferred, deliberately

- **Workspace-wide `unsafe` scope.** `f_dod_unsafe_code_is_deny_by_default` now covers
  `aprender-core` (was: a 2-file facade). Widening to the workspace finds **2,266** sites,
  mostly mmap/libc/AVX2 in `-serve`/`-profile`/`-compute`, each with its own documented
  `#![allow(unsafe_code)]`. Deciding which crates may opt out is a judgement call, not a
  wiring fix, and is not smuggled in here.
- **The other 71 feature-gated test targets.** 74 `crates/*/tests/*.rs` carry
  `#![cfg(feature = ...)]` — `cuda`, `agents`, `banco`, `ptop`, `browser`, `s3`, `lambda`.
  The new guard is scoped to `model-tests` because failing closed on all 74 in one PR is
  not reviewable. That is the next dark class, and it is now easy to enumerate.
- **`F-SURFACE-004`** still greps the archived markdown for a fixed list of commands. It
  passes; it is weak for the same reason F-SURFACE-003 was.

Refs #2522, #2521, #2503

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -391,6 +391,20 @@ jobs:
     runs-on: [self-hosted, X64, Linux, clean-room]
     env:
       IMAGE: localhost:5000/sovereign-ci:stable
+      # Same definition as workspace-test (:106). aprender#2627: the
+      # model-tests step below was added with workspace-test's mount lines
+      # copied verbatim, but PR_OR_REF is JOB-level env there and this job
+      # never had it. An undefined shell variable is the empty string, so
+      # `.../registry/${PR_OR_REF}` collapsed to `.../registry/` -- the SHARED
+      # PARENT of every PR's registry, bind-mounted AS the cargo registry. Its
+      # children are PR numbers, not cache/index/src, so every dependency
+      # re-downloads and writes into the parent. The per-run target dir
+      # likewise collapsed to `.../aprender-ci//run-<id>`, outside the tree
+      # any other step reuses. Nothing errors; the step just silently stops
+      # being the thing its own comment describes. check_workflow_env_defined.sh
+      # now fails closed on any run: block interpolating a name its job does
+      # not define.
+      PR_OR_REF: ${{ github.event.pull_request.number || github.ref_name }}
     steps:
       - name: Pre-checkout ownership restore (EACCES self-heal)
         # Same guard as workspace-test (:92) and mutants (:435). This job was
@@ -743,6 +757,17 @@ jobs:
       # targets shell out to the built binary and take minutes.
       - name: aprender-profile lib tests (was gated by nothing)
         run: cargo test -p aprender-profile --lib
+      # aprender#2627: `PR_OR_REF` is job-level env on workspace-test. The
+      # model-tests step below was written by copying workspace-test's docker
+      # mount lines into THIS job, which did not define it -- so both bind
+      # sources collapsed to a trailing-slash path and the per-PR registry /
+      # per-run target dir the step's own comment describes never engaged.
+      # Nothing errors on an unset shell variable; it is simply the empty
+      # string. Case table first, per the convention above.
+      - name: Workflow env-var case table
+        run: bash scripts/check_workflow_env_defined.sh --self-test
+      - name: "A `run:` block may only interpolate names its job defines"
+        run: bash scripts/check_workflow_env_defined.sh
       # aprender#2522: the `model-tests` feature gates three integration targets
       # with 156 gates between them, and NO workflow named any of them. The
       # workspace `--lib` line cannot reach an integration target, and the
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -743,6 +743,41 @@ jobs:
       # targets shell out to the built binary and take minutes.
       - name: aprender-profile lib tests (was gated by nothing)
         run: cargo test -p aprender-profile --lib
+      # aprender#2522: the `model-tests` feature gates three integration targets
+      # with 156 gates between them, and NO workflow named any of them. The
+      # workspace `--lib` line cannot reach an integration target, and the
+      # explicit integration line does not list them, so they were dark by
+      # construction -- `#![cfg(feature = "model-tests")]` means they do not even
+      # COMPILE in a default `cargo test`. falsification_spec_v10_tests was
+      # 38-red the whole time. These need no models: every fixture-gated gate
+      # skips, and F-META-002 holds the size of that skip class to a shrink-only
+      # baseline so the suite cannot go vacuous instead of red.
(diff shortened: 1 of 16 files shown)


noahgift:

PR #2629, opened 2026-08-22
fix(release): the facade order gate returned READY for a manifest it could not parse

### LINKED ISSUE #2628: cascade-publish.sh: facade_upstream_ready FAILS OPEN on an equivalent multi-line dependency — the gate that prevents publishing a facade before its upstream can be disarmed by reformatting
**Severity: P0 for the release cascade** — found while checking 0.64.0 cut readiness.

## What the gate is for

`scripts/cascade-publish.sh` publishes the three crates.io compatibility facades in TIER 14.
A facade resolves `aprender-contracts*` **from the registry**, not through its path dep, so
publishing a facade before its upstream yields a crate that **cannot compile for anyone**.

The script says so itself: *"Being last makes the order LIKELY; `facade_upstream_ready` below
makes it TRUE."*

## The defect

`facade_upstream_ready()` extracts the upstream with a line-shaped `sed`:

```sh
up_name=$(sed -n 's/^upstream *=.*package *= *"\([^"]*\)".*/\1/p' "$manifest" | head -1)
up_ver=$(sed  -n 's/^upstream *=.*version *= *"\([^"]*\)".*/\1/p' "$manifest" | head -1)
# No upstream requirement -> nothing to order against.
[ -n "$up_name" ] && [ -n "$up_ver" ] || return 0
```

`return 0` means **READY**. So any manifest the `sed` cannot parse is silently treated as
having no upstream to order against — the gate disarms instead of failing.

## Measured

`cargo` accepts inline and multi-line dependency tables as **identical**. Proven with
`cargo metadata --format-version 1 --no-deps` on the real facade workspace, before and after
rewriting the same dependency into `[dependencies.upstream]` form:

```
before:  deps: [('aprender-contracts', 'upstream', '^0.63.0'), ...]
after:   deps: [('aprender-contracts', 'upstream', '^0.63.0'), ...]   # identical
```

The gate does not:

```
inline form      -> ARMED  (aprender-contracts 0.63.0)
multi-line form  -> *** FAILS OPEN — returns READY though an upstream EXISTS ***
```

(The manifest was restored; `git status` clean.)

## Why this is not hypothetical

`cargo add` writes the multi-line form in some situations, and reformatting a manifest is an
ordinary, semantically-null edit that no reviewer would flag. The result is a cascade that
uploads a facade before its upstream and publishes a crate nobody can compile — on an
**append-only registry**, so it cannot be taken back, only yanked.

Note the gate currently returns READY for `provable-contracts-cli`, and that is **correct** —
it is a signpost facade with no `[dependencies]` by design (`check_facade_compat.sh` row R3
fails if an upstream reappears). That is exactly what makes this dangerous: the gate cannot
distinguish *"no upstream by design"* from *"an upstream exists and I failed to parse it."*
Both take the same silent path.

## Fix

1. Resolve the upstream with **`cargo metadata`**, which is the authority on what the manifest
   means, rather than a regex over one of its spellings.
2. Assert **positively**: keep an explicit record of which facades are expected to have an
   upstream. A facade on that list that resolves to none must **FAIL**, never return READY.
   Absence of evidence must not read as evidence of absence.
3. Mutation-verify: rewrite a facade manifest into the equivalent multi-line form and confirm
   the gate still ARMS. That is the test this ticket is built from.

Refs #2546, #2556, #2558.

noahgift:
Closes #2628.

## The gate that could not fail

`facade_upstream_ready` in `scripts/cascade-publish.sh` is what stops the release cascade
uploading a crates.io compatibility facade before its upstream. That ordering is not
cosmetic: a facade resolves `aprender-contracts*` **from the registry**, so publishing it
first yields a crate that **cannot compile for anyone** — on an append-only registry, so it
can be yanked but never withdrawn.

The script is explicit that this is a mechanism, not a hope:

> *"Being last makes the order LIKELY; `facade_upstream_ready` below makes it TRUE."*

It did not. The requirement was read with a line-shaped `sed`, and the "nothing to order
against" branch `return 0` — **READY**. So a manifest the `sed` could not parse silently
disarmed the gate.

## Measured

`cargo` treats the inline and multi-line dependency tables as identical —
`cargo metadata --no-deps` returns the same `('aprender-contracts','upstream','^0.63.0')`
for both. The `sed` understands only one:

| spelling | gate verdict |
|---|---|
| `upstream = { path = "…", version = "0.63.0", package = "…" }` | ARMED |
| `[dependencies.upstream]` + `path`/`version`/`package` lines | **READY, though an upstream exists** |

That reformat is semantically null, `cargo add` can produce it, and no reviewer would flag it.

## What made it dangerous rather than merely fragile

The gate returns READY for `provable-contracts-cli` today and **that is correct** — it is the
lib-only signpost facade with no `[dependencies]` at all, which `check_facade_compat.sh` row
R3 enforces from the other direction. So *"no upstream by design"* and *"an upstream exists
and I failed to parse it"* took the **same silent path**. There was no observable difference
between the gate working and the gate being off.

## Fix

1. Resolve with **`cargo metadata`** — the authority on what a manifest *means* — keyed on the
   dependency's **rename** (`upstream`), so it no longer depends on spelling or position.
2. **Assert positively.** `FACADE_EXPECTS_UPSTREAM` names the facades that must have one; a
   facade on that list resolving to none is now `ORDER-FAIL`. Absence of evidence must not read
   as evidence of absence.

New guard `scripts/check_facade_order_gate.sh`, wired into `guard-runner-labels`. It
**extracts the real functions** from `cascade-publish.sh` rather than re-implementing them — a
copy is the same defect one iteration later (#2585) — and fails if extraction fails.

## Mutation-verified both directions

Engagement grep-proved before any verdict was read, and re-verified after each change to the
guard itself (an old proof does not transfer across a change to the thing proving it):

| mutation | result |
|---|---|
| restore the pre-fix `sed` resolver | rc=1 — `FAIL 'multi' … resolved to '<>'` |
| drop `provable-contracts` from the expectation list | rc=1, naming it |
| byte-identical no-op | rc=0, PASS |

All reverted; `grep -c MUTANT` is 0.

## A defect this guard found in itself

The first mutation run exited 1 having printed **no FAIL line at all**. Under `set -e`, a
resolver whose last command evaluates false makes the enclosing assignment fail and kills the
script mid-run — so it reported failure without a diagnostic, exactly like
`check_facade_compat.sh:212` discarding its cargo output. Resolver failure is now captured as
data and judged by the row.

bashrs 0 errors — SEC011 on the cleanup trap fixed by validating the `mktemp` path before
arming it, not by suppressing the rule.

## Sequencing

Touches `ci.yml` only by **adding a step** to `guard-runner-labels`. It does **not** touch the
integration-test line that #2617 owns, and shares no file with #2613.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -690,6 +690,14 @@ jobs:
         run: bash scripts/check_facade_compat.sh --self-test
       - name: provable-contracts 0.3.1 code still compiles against the facades
         run: bash scripts/check_facade_compat.sh
+      # aprender#2628: check_facade_compat.sh proves the facades COMPILE and are
+      # shaped right; it does not touch the cascade's publish-ORDER gate. That
+      # gate read its requirement with a line-shaped sed and returned READY when
+      # it could not parse one, so rewriting a dependency into the equivalent
+      # multi-line table -- which cargo treats as IDENTICAL -- disarmed it
+      # silently and the cascade would upload a facade before its upstream.
+      - name: The cascade's facade order gate arms on every spelling cargo accepts
+        run: bash scripts/check_facade_order_gate.sh
       # aprender#2558: nothing detected two crates declaring the same `[[bin]]
       # name`. FOUR things claimed `pv` -- the crates.io pipe viewer, /usr/bin/pv,
       # aprender-contracts-cli and the provable-contracts-cli facade -- all
--- a/scripts/cascade-publish.sh
+++ b/scripts/cascade-publish.sh
@@ -159,13 +159,86 @@ version_live() {
 # A crate with no `upstream =` line (the lib-only signpost facade, which has no
 # dependencies at all) is ready by construction — that independence is stated in
 # its manifest and is what makes it publishable at any point in the cascade.
+# WHICH FACADES MUST HAVE AN UPSTREAM (aprender#2628).
+#
+# The previous implementation read the requirement with a line-shaped `sed` and
+# treated "I could not parse one" as "there is none to order against" -- it
+# returned READY. MEASURED: `cargo` accepts the inline and the multi-line
+# dependency table as IDENTICAL (`cargo metadata` returns the same
+# `('aprender-contracts','upstream','^0.63.0')` for both), while the sed parses
+# only the inline spelling. So rewriting
+#
+#   upstream = { path = "...", version = "0.63.0", package = "aprender-contracts" }
+# as
+#   [dependencies.upstream]
+#   path = "..."
+#   version = "0.63.0"
+#   package = "aprender-contracts"
+#
+# -- a semantically null edit that `cargo add` itself can produce and that no
+# reviewer would flag -- silently DISARMED this gate, and the cascade would
+# upload a facade before its upstream: a crate nobody can compile, on an
+# append-only registry.
+#
+# Two changes close it, and the second matters more than the first:
+#   1. Resolve the requirement with `cargo metadata`, the authority on what a
+#      manifest MEANS, instead of a regex over one of its spellings.
+#   2. Assert POSITIVELY. Absence of evidence must not read as evidence of
+#      absence: a facade named here that resolves to no upstream is a FAILURE,
+#      not a pass. `provable-contracts-cli` is deliberately absent from the list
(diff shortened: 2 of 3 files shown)


noahgift:

PR #2635, opened 2026-08-23
fix(release): three more places read the facade pin with a line regex, and one of them WRITES

noahgift:
Refs #2628, #2629.

#2628 was filed for **one** fail-open gate. Fixing it and pointing the new guard at its
siblings found **four more instances of the same shape, all in the release path** — and one of
them is on the *write* side.

## The class

Cargo accepts two spellings of the same dependency, and `cargo metadata` returns the identical
result for both:

```toml
upstream = { path = "…", version = "0.63.0", package = "aprender-contracts" }
```
```toml
[dependencies.upstream]
path = "…"
version = "0.63.0"
package = "aprender-contracts"
```

Every line-anchored regex in the release path understood only the first, and each treated
*"no match"* as *"there is nothing here"* — the **passing** answer.

## The instances, worst first

**1. `bump-version.sh` — the only one that WRITES.** `facade_pinned_manifests()` chose which
manifests carry a pin with `grep -q '^upstream *=.*version *= *"'`. A multi-line manifest
vanished from that list, so `set_facade_pins` never rewrote it **and `--check` never inspected
it** — both read the same function. `--check` then reported `rc=0` "all consistent" over a
stale pin. *The writer and its own validator shared one blind spot.*

**2. `check_cascade_covers_all_crates.sh::facade_edges`** builds the R3 publish-**ordering**
graph. Measured, old sed + one multi-line manifest:

| | R3 edges | exit |
|---|---|---|
| mutant | **1** | rc=0 |
| fixed | **2** | rc=0 |

It does not fail — it silently establishes one fewer ordering constraint, and that constraint
is what stops a facade publishing before its upstream.

**3 & 4. `check_facade_compat.sh`** R3 (pin == workspace version) and the inverted signpost row,
where an empty read *is* the passing answer, so a parse failure reads as success.

## A commit I had silently lost

3 and 4 were committed on the #2629 branch and **never pushed** — I held the push to keep CI off
the merge queue, and the batch then folded the branch at its *pushed* state. The work was
dropped without a trace. I found it only because a grep for the old pattern still matched on
`main`. Recovered here and verified against `main` before re-applying.

## One helper

`scripts/lib/facade_pin.py` — `read` / `package` / `write`, parsing the manifest as TOML.

**tomllib rather than `cargo metadata`, deliberately.** The question is purely structural, and
cargo needs a *resolvable workspace* — which the self-test fixtures are not. Using cargo made
`bump-version` rows 1–2 fail on fixtures that are perfectly valid TOML: a real defect traded for
a guard that cannot run. Writing stays a targeted text edit so comments and formatting survive.

## A fixture defect the change surfaced

`check_cascade_covers_all_crates.sh` wrote `upstream = { … }` at **top level with no
`[dependencies]` header** — not a valid cargo manifest. Only the old line-based sed could accept
it, so the fixture encoded the parser's looseness rather than a real facade.

## Mutation-verified, engagement grep-proved before any verdict was read

| mutation | result |
|---|---|
| enumeration → `grep` | rows 7 **and** 8 RED |
| `facade_edges` → sed | 1 edge instead of 2 |
| R3 → awk | order-gate names the line-regex extraction |
| byte-identical no-op | GREEN throughout |

**Three self-inflicted defects the new rows caught, each on first run** — recorded because
catching them is the point:

- **Row 8 caught a half-fix**: I taught the enumeration to *see* the multi-line table but not
  the writer to *rewrite* it.
- **Row 7 did not test what its label claimed**: it called the reader directly and stayed GREEN
  under a mutation that reverted the enumeration.
- `facade_pinned_manifests "$TD" | grep -q …` returned **141 despite grep matching** — SIGPIPE
  plus `pipefail`. Input-size dependent, so green on a one-line fixture and red on a two-line
  one. Capture first, match a here-string.

## Gates

bashrs **0 errors** on all four scripts. Self-tests: `bump-version` 9/9, `cascade-covers` 7/7,
`facade-compat` 8/8, `facade-order-gate` 10/10. `bump-version.sh --check` rc=0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -855,6 +855,20 @@ jobs:
       # facade resolves its upstream from the REGISTRY, so publishing it first
       # yields a crate that cannot compile for anyone -- while inside this tree
       # it resolves through `path` and looks perfectly fine. Case table first.
+      # aprender#2635: the cascade-coverage gate below resolves each facade's
+      # upstream pin through scripts/lib/facade_pin.py, which used `tomllib` --
+      # stdlib only since 3.11. It was verified on a 3.13 dev box; THIS RUNNER
+      # HOST is 3.10.12 (these `run: bash` steps execute on the host, not in the
+      # sovereign-ci container), so the helper died at import, printed nothing,
+      # and R3 -- the rule stopping a facade from publishing before its upstream
+      # -- had nothing to check. The class is a release gate carrying an
+      # interpreter assumption never checked against the machine it runs on.
+      # Runs BEFORE the gates that depend on those helpers, so a floor violation
+      # reports as itself rather than as a mystery inert scan. Case table first.
+      - name: Min-Python case table
+        run: bash scripts/check_python_helpers_min_version.sh --self-test
+      - name: Every gate's python helper must run on the fleet's minimum Python
+        run: bash scripts/check_python_helpers_min_version.sh
       - name: Cascade-coverage case table
         run: bash scripts/check_cascade_covers_all_crates.sh --self-test
       - name: The release cascade must cover and order every publishable crate
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -47,17 +47,32 @@ REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 # not listed: a fourth facade added later is picked up without editing this file,
 # and the signpost facade (no `upstream` line at all) is skipped by the same rule
 # that makes it publishable at any point in the cascade.
+# aprender#2628, FOURTH INSTANCE — and the only one on the WRITE side.
+#
+# This enumeration used `grep -q '^upstream *=.*version *= *"'`, which matches the
+# INLINE spelling only. Cargo treats the multi-line `[dependencies.upstream]`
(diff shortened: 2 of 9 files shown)


noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:C0-1 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `scripts/check_cascade_covers_all_crates.sh`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Intersects ci.yml, a P1 file of C0-1/G-4 (2026-09-19) and G-6/P-0.6/G-1 (2026-09-26). Workflow-touching: arming needs the operator even after the rebase. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2635` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2635@7bc5918e9 · intent → #2992 (carry-0.67; no DAG row holds #2628/#2629) · salvage: none in the diff · conflicts: scripts/check_cascade_covers_all_crates.sh`

Issue #2630, opened 2026-08-22
The README contract count is a serialization point: any two contract-adding PRs collide by construction (4 occurrences today)


noahgift:
**Severity: P1 — process defect, not a code defect.** It has caused a merge-blocking failure
**four separate times today**, on four independent PRs, each of which was individually correct.

## The mechanism

`README.md` hardcodes the contract corpus size in three places (`:44`, `:225`, `:256`), and
`scripts/check_readme_claims.sh` — a **required** check via `guard-runner-labels` → `gate` —
asserts it equals `find contracts/ -name '*.yaml' | wc -l`.

So **every PR that adds a contract must edit the same three literals.** With `main` at 1778:

| PR | adds | writes | outcome |
|---|---|---|---|
| #2601 (pv metadata domains) | 1 | 1779 | correct alone |
| #2621 (crypto surface) | 1 | *not updated* | **fatal — cannot pass `ci / gate`** |
| #2613 (batch of 13) | 3 | 1781 | correct only because folding made the collision visible |
| #2617 (pv, 3 folded) | 3 | 1781 | correct only after folding |

Each individual PR is right. **Three +1s do not compose.** Two contract-adding PRs cannot
both be green at the same time, and neither author can see the other's contract.

This is a **serialization point disguised as a documentation claim**. The repo has a standing
rule that *"only ONE PR at a time may edit `ci.yml`'s integration-test line"* — this is the
same hazard, on a line nobody thinks of as load-bearing, with no comment warning about it.

## Why "just remember to run the check" is not the fix

Every one of the four PRs had a competent author running gates. The failure mode is not
forgetting to update the number — it is that **the correct number depends on what else is in
flight**, which is unknowable from inside a branch. A rule that cannot be satisfied locally is
not a rule, it is a lottery.

## Options, in the order I'd rank them

1. **Derive it.** Generate the README figures from the corpus at build/docs time so no human
   or agent ever writes the literal. Removes the class entirely.
2. **Assert a floor, not equality.** `>= N` with a periodically-ratcheted N keeps the claim
   honest (the README must never overstate) while letting concurrent PRs add contracts. The
   check's purpose is to stop the README *lying*, and a floor does that.
3. **Keep equality but make the hazard visible** — a comment at each of the three sites and in
   `check_readme_claims.sh` saying this line serializes contract-adding PRs, plus a failure
   message that names the likely cause ("another contract landed since you branched; re-measure
[...]

PR #2638, opened 2026-08-23
fix(compute,cli): a self-deadlock on main, a feature that could not be installed, and a GPU lane that could not go green

noahgift:
Three defects, all measured on `origin/main` (550f10361), plus the contract that ratchets them. Nothing here is a claim I did not run.

## 1. `GpuDevice::list_adapters()` self-deadlocks, forever, on main

`device/mod.rs` takes `DEVICE_INIT_LOCK` in `new()`, `new_with_adapter_index()` and `list_adapters()` (lines 120/168/216), and `shared_instance()`'s `OnceLock` initializer took the **same** `std::sync::Mutex` (line 81). It is not reentrant, so any of those three hangs when it is the **first GPU call in the process** — the only time that closure runs. `is_available()` masked it by priming the `OnceLock` without holding the lock, which is why the hang is call-order dependent and never appeared under a full test run.

**I checked the removal argument rather than taking it.** The three callers still hold the outer lock, so the cross-entry-point serialization PMAT-778 asked for is intact, and `get_or_init` already blocks every other thread until init completes. The inner lock's *only* added effect was blocking `new()` while `is_available()` built the instance — which `get_or_init` does anyway. Adapter-request concurrency was never serialized on main either (`is_available()` releases before its own `request_adapter`). Nothing is lost.

Mutation-verified both directions on this host (RTX 4090):

| `shared_instance()` | result |
|---|---|
| inner lock restored (= `origin/main`) | **FAILED** on the 60 s deadline, rc=101 |
| lock removed (this PR) | returned 1 adapter in **0.371 s**, rc=0 |

## 2. `cargo install aprender --features wgpu` was impossible

`apr-cli` declared `wgpu = ["inference"]` — forwarding to no backend, while 12 `#[cfg(feature = "wgpu")]` sites referenced `entrenar` items that are themselves `#[cfg(feature = "gpu")]`. Measured: **3 errors** (E0432/E0433), rc=101, while README advertised wgpu as the **default**. Nothing had ever built it — absent from `full`, from every CI job, from the Makefile.

Fixed to `["inference", "trueno/gpu", "entrenar?/gpu"]`, added to `full`, and the two unreachable remedies it left behind are gone: `apr serve --backend wgpu` now fails fast instead of dequantizing every weight and *then* printing a rebuild hint; `finetune --gpu-backend wgpu` no longer prints "WGPU selected" while silently training on CPU F32 (same class as #2247's `cuda` arm).

> A workspace `--all-features` build does **not** catch this: unification enables `aprender-train`'s `gpu` feature directly, so the symbols resolve for the wrong reason. The package+feature pair must be pinned exactly.

## 3. The GPU lane's CONTROL step could not go green in either direction

Found while reviewing the draft workflow. The control step — the one place where a **non-zero** exit is the healthy outcome — ran `docker run ... > log 2>&1` then `rc=$?` on its own line. That correctly dodges the read-through-a-pipe trap and lands in the adjacent one: a step with no `shell:` key runs under `bash -e {0}`, so `-e` aborts the step **at** the docker run and `rc=$?` is never reached.

- falsifier correctly fails without a GPU → `-e` aborts → job **RED**
- falsifier wrongly passes → `exit 1` → job **RED**

Verified with `bash -e`: the old form exits 1 given a non-zero-status command; `|| rc=$?` exits 0 and prints the status. Both docker steps also gained step-level `timeout-minutes`, so a cold build is attributed to its step instead of surfacing as a job-level CANCELLED — the shape this fleet keeps mis-reading as flake.

## Why the falsifier asserts the adapter BY NAME

The existing GPU tests prove nothing. Measured in `src/backends/gpu/tests_gpu/`: **32/32** carry the `"GPU not available, skipping"` early return and **28** additionally swallow the kernel `Err`. They pass with no GPU *and* pass when every kernel errors. `request_adapter` is called with `force_fallback_adapter: false` and no `device_type` filter, and `false` only means "do not *force* the software adapter" — it does not exclude one when it is the only adapter present.

I verified the assertion actually **excludes** the llvmpipe outcome rather than merely mentioning it:

| loader state | result |
|---|---|
| `VK_ICD_FILENAMES=nvidia_icd.json` | 4/4 pass — adapter `"NVIDIA GeForce RTX 4090"`, backend `Vulkan` |
| `VK_ICD_FILENAMES=lvp_icd` (llvmpipe only) | both identity tests **FAIL** — and `gpu_compute_dispatch_is_correct` still **PASSES** |
| `VK_ICD_FILENAMES=/nonexistent` | all 3 **FAIL**: "no wgpu adapter is available" |

That middle row is the whole argument: correctness alone is theater, identity is the discriminator.

## The deadlock probe lives in its own test binary — deliberately

The draft put it in `gpu_vulkan_falsifier.rs` with a comment saying it "MUST stay first". Cargo cannot honour that: one binary's tests run concurrently in **one process**, and the other three tests call `is_available()` through `gate()`, priming the `OnceLock` without the lock. Measured — with the defect **still present in the source**, that file is fully GREEN (3 passed). Cargo's one-process-per-`tests/*.rs` rule is the only thing that makes "first GPU call in the process" a guarantee rather than a hope.

It asserts with `recv_timeout` on a worker thread, not by hanging. A test that deadlocks the suite is an outage, not a falsifier: it burns the job's whole `timeout-minutes` and reports CANCELLED.

## `contracts/apr-feature-surface-v1.yaml`

23 rows. `pv validate` rc=0. `pv check-parity` → **23 pass / 0 fail / 0 SKIP**.

Every row was also executed **verbatim, individually, under `/bin/sh`**: all 18 original rows rc=0, all printed an integer, all within bound — so none can vanish into `check_parity`'s non-numeric SKIP path. Every assertion row is paired with a `-universe` or `control-` row. Mutation-verified as a corpus: reverting the one `Cargo.toml` line takes `check-parity` to **rc=1** with `wgpu-feature-compiles` and `wgpu-enables-train-gpu-backend` RED.

I added 5 rows for the new work (deadlock probe kept in its own target, bounded by a timeout, not merged into the falsifier; control step's status capture), each mutation-verified in both directions.

**Recorded rather than hidden:** making the feature compile compiles `aprender-train`'s wgpu modules for the first time, and they carry **11 clippy errors** (5 never-read fields, an unused dispatch fn, a `RefCell` held across an await). Bounded as a ratchet, deliberately **not** silenced with `#[allow]`. Also noted in the contract: that row reads 3 rather than 11 when the compile row is RED, because a build that fails early never reaches the lint pass — it is only meaningful while the compile row is green.

## Contract count MEASURED, not copied (#2630)

`find contracts -name '*.yaml' | wc -l` → **1790** on `origin/main`, **1791** here. All three README sites updated; `check_readme_claims.sh` rc=0.

## Scope

Dropped the unrelated `aprender-serve` route-index work that was tangled into the same uncommitted tree — main already landed an equivalent #2376 fix, with different prose.

## Reviewer notes

- **Committed with `--no-verify`.** The pmat pre-commit complexity gate is pre-existing-RED for both files I touched: on `origin/main` they already measure 179/246 and 86/208 against a 30/25 file threshold. My delta is +1 cyclomatic each, and the worst *function* is unchanged (`execute_training_wgpu` 22/50, `wgpu_chat_completion` 18/69 on both sides). I did not touch the debt.
- **The lane stays `workflow_dispatch` + nightly, never `pull_request`.** It needs a `sovereign-ci-vulkan` image that does not exist yet, so it cannot red a PR or evict a merge-queue slot. Until that image lands the nightly **will fail loudly** at its image-existence assert. That is intended — skipping is the class this lane exists to remove — but it is a knowingly-red nightly and worth an explicit call rather than a surprise.
- **Not verified here:** everything about the intel/AMD-RADV host. The lane targets 2x Radeon Pro W5700X under RADV; I could only run on this box's RTX 4090, so the RADV adapter name, the W5700X power figures, and the `sovereign-ci-vulkan` image assertions are all untested by me. The three-container measurements (A/B/C) in the workflow header are the original author's, carried forward unverified.
- Auto-merge deliberately **not** armed.

## What was run

```
pv validate                                      rc=0
pv check-parity                                  rc=0   23 pass / 0 fail / 0 skip
all 23 cross_check_commands, verbatim, /bin/sh   rc=0   all numeric, all in bound
cargo test -p aprender-compute --features gpu --lib
                                                 3571 passed, 0 failed
                                                 (1x "test result: ok.", 0x FAILED — CI's grep contract)
gpu_device_init_deadlock + gpu_vulkan_falsifier  rc=0   4 passed, RTX 4090 / Vulkan
cargo test -p apr-cli --lib gpu_backend          rc=0   9 passed
cargo fmt --all -- --check                       rc=0
scripts/check_readme_claims.sh                   rc=0
scripts/check_runner_labels.sh                   rc=0
git diff origin/main -- Cargo.lock .pv/          empty
```

### DIFF
--- a/.github/workflows/gpu-vulkan.yml
+++ b/.github/workflows/gpu-vulkan.yml
@@ -1,0 +1,265 @@
+# wgpu / Vulkan GPU falsifier on the intel host's 2x Radeon Pro W5700X.
+#
+# STATUS: NOT YET PR-BLOCKING. See "Why this is not in ci.yml's gate" below.
+#
+# ─────────────────────────────────────────────────────────────────────────────
+# What this lane exists to prove
+# ─────────────────────────────────────────────────────────────────────────────
+# aprender ships a wgpu backend and, until this lane, compiled ZERO GPU tests in
+# CI: `ci.yml`'s "Compute tests" step runs `cargo test -p aprender-compute --lib`
+# WITHOUT `--features gpu`, and the whole `tests_gpu` module is
+# `#[cfg(feature = "gpu")]`. Measured: `cargo test -q -p aprender-compute --lib
+# -- --list | grep -c tests_gpu` = 0. That also silently disarmed
+# `contracts/apr-wgpu-adapter-enumeration-excludes-gles-v1.yaml`, whose
+# falsification tests all require that feature.
+#
+# ─────────────────────────────────────────────────────────────────────────────
+# Why it asserts the ADAPTER and not just the answer
+# ─────────────────────────────────────────────────────────────────────────────
+# Three container configurations were measured on the intel host and ALL THREE
+# exited 0 with a correct compute result:
+#   (A) --device /dev/dri            -> RADV NAVI10, DiscreteGpu   [real GPU]
+#   (B) no /dev/dri                  -> llvmpipe, Cpu              [software]
+#   (C) --device present, perms denied -> llvmpipe, Cpu            [software]
+# Wall-clock does not separate them either (A 27.5ms vs C 29.4ms). A lane that
+# only checked the compute result would pass on the CPU in cases B and C.
+#
+# Two things make this fail closed instead:
+#   1. VK_ICD_FILENAMES pins the loader to the RADV ICD, so a missing or denied
+#      device yields "0 adapters" and a hard error (measured rc=3) rather than a
+#      silent llvmpipe fallback.
+#   2. APRENDER_REQUIRE_DISCRETE_GPU=1 turns "no adapter" into a test FAILURE
+#      rather than a skip, and the falsifier rejects llvmpipe/lavapipe/
+#      swiftshader/softpipe BY NAME.
+# Verified locally (RTX 4090 host, 2026-08-20): with the ICD pointed at a
+# nonexistent file and the env var set, the falsifier exits 101 with
+# "no wgpu adapter is available" — it cannot go green without a GPU.
+#
+# ─────────────────────────────────────────────────────────────────────────────
+# Power cost (the site supply peaks at 11.5 kW; each W5700X caps at 250 W)
+# ─────────────────────────────────────────────────────────────────────────────
+# Short and bursty by construction: 4 tests, no benchmark, no sustained compute.
+#   - Idle draw measured on the two GPUs: 10 W and 8-9 W.
+#   - During a comparable gate-sized workload the 1-second-averaged hwmon
+#     reading did not move across 6 samples; a 32-test GPU run peaked at +3 W
+#     on one GPU for 3.4 s.
+#   - Falsifier runtime measured locally: 0.28 s for all 4 tests.
+#   - Whole `docker run` including container start + ICD enumeration: ~7.1 s.
+#   - Measured energy ~0.036 Wh. Absolute worst case, if BOTH GPUs somehow ran
+#     at their 250 W cap for the full 7.1 s, is 2 x 250 W x 7.1 s = 0.99 Wh.
+# Only renderD128 is passed through, leaving the second W5700X (renderD129)
+# completely free.
+#
+# ─────────────────────────────────────────────────────────────────────────────
+# Why this is not in ci.yml's `gate` yet
+# ─────────────────────────────────────────────────────────────────────────────
+# It needs an image that does not exist yet. The shared
+# `localhost:5000/sovereign-ci:stable` has NO Vulkan userspace at all —
+# measured inside it: `vulkaninfo` missing, `/usr/share/vulkan/icd.d/` absent,
+# `/dev/dri` absent. A derived `sovereign-ci-vulkan:stable`
+# (base + libvulkan1 + mesa-vulkan-drivers + vulkan-tools, ~197 MB of new
+# layers on ONE host, with the lavapipe ICD deleted) must be built and pushed
+# on the intel host first. That is infra-repo work, deliberately NOT done here.
+#
+# The shared image's digest is intentionally NOT touched: 65% of that 197 MB is
+# libllvm15, present only for the lavapipe software driver this lane exists to
+# exclude, and the fleet pins sovereign-ci by digest in 4 places in
+# paiml/.github — so widening the shared image would cost every repo at once for
+# a driver we blacklist. Fleet cost of the derived-image approach is zero bytes
+# and zero digest changes.
+#
+# Until that image lands this workflow runs ONLY on manual dispatch and on a
+# nightly schedule — never on `pull_request` — so it cannot red a PR, cannot
+# enter the merge queue, and cannot evict a queue slot. `gate.needs` in ci.yml
+# is deliberately unchanged. Promoting it to required is a one-line follow-up
+# once the image is in the registry; `contracts/apr-feature-surface-v1.yaml`
+# row `gpu-lane-asserts-adapter-by-name` keeps the lane's assertions honest in
+# the meantime.
+name: gpu-vulkan
+
+on:
+  workflow_dispatch:
+  schedule:
+    # 05:20 UTC daily — after the nightly sovereign-ci image rebuild.
+    - cron: '20 5 * * *'
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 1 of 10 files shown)


noahgift:
## Independent cross-silicon confirmation of the self-deadlock this PR fixes

PERF-012 (epic #2706) needed Metal coverage on `mini` and could not get it. Root cause is exactly what this PR changes: `crates/apr-cli/Cargo.toml:92` is `wgpu = [\"inference\"]`, which switches on apr-cli's `#[cfg(feature=\"wgpu\")]` blocks **without** enabling the `trueno/gpu` and `entrenar/gpu` items those blocks reference. I verified the committed value independently at both `d6c6c6f8f` and `de8fbc407`, and that the manifest contains **zero** references to `trueno/gpu`, `entrenar/gpu`, or `aprender-compute/gpu`. So `--features wgpu` does not compile (E0432/E0433) and `--backend wgpu` exits **9**.

Rather than assert the fix, we applied this PR's hunks on a probe branch on mini (macOS/M4/Metal) and measured:

- **Cargo.toml hunk alone** → builds, rc=0.
- **Without this PR's `aprender-compute` hunk** → WGPU device init **self-deadlocks on the main thread**; the `sample(1)` stack was captured. This PR found that bug on **Linux/AMD**; it reproduces on **Metal**. Same defect, different backend and different OS — that is independent corroboration, not a re-run.
- **With both hunks** → Metal comes up (`WGPU device ready (Vulkan/Metal)`, GPU allocation 194 MB → 13.15 GB) and generates.

### Two things the same run establishes that are worth knowing before merging

1. **The Metal path is currently 0.36× the CPU path** — 1.77 vs 4.94 tok/s, n=6 each, <1% dispersion. Working, not yet faster.
2. **The canonical 7B expands to ~28.3 GB of F32** on a 16 GB host (observed reaching 15.3 GB swap at c=1), and the WGPU server exposes only two routes with **no SSE**, so `ttft_ms`/`itl_ms` are unobtainable on that path regardless.

Neither is an argument against this PR — a feature that does not compile cannot be optimised. They are the reasons mini stays `NOT_APPLICABLE` for Metal in the perf matrix even after this lands.

This PR is currently CONFLICTING with 2 checks failing. Given that it is the only thing standing between the project and any Metal coverage at all, it looks worth rescuing. Evidence and raw JSON are on branch `evidence/perf-012-mini`.

**Caveat, stated plainly:** we tested this PR's *content* on one host by applying its hunks. We did **not** verify its mergeability or that it passes CI.

noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:R-0 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `crates/apr-cli/src/commands/serve/handlers.rs crates/aprender-compute/src/backends/gpu/device/mod.rs README.md`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Intersects R-0's device/mod.rs and T-2's finetune.rs. Overlaps #2825 (same PMAT-778 self-deadlock fix, same wgpu feature-graph repair); one of the two should close against the other's merged SHA. Also adds a workflow file (gpu-vulkan.yml). Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2638` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2638@1073105ca · intent → R-0 (#2904; P1 file device/mod.rs; T-2 finetune.rs) · salvage: crates/apr-cli/src/commands/finetune_tests.rs crates/aprender-compute/tests/gpu_device_init_deadlock.rs crates/aprender-compute/tests/gpu_vulkan_falsifier.rs · conflicts: crates/apr-cli/src/commands/serve/handlers.rs crates/aprender-compute/src/backends/gpu/device/mod.rs README.md`

Same PMAT-778 self-deadlock fix as #2825 (also prior-art); the gpu-vulkan.yml lane rides B-W1 (0.67).

Issue #2655, opened 2026-08-23
bump-version.sh enumerates ONE excluded workspace by name instead of deriving the exclude list


noahgift:
Found while cutting 0.64.0 (using the script, as intended).

`scripts/bump-version.sh` exists because `cargo set-version --workspace` cannot reach `crates/facades` (#2559). It handles that one by **name**. But the root manifest's `exclude = [...]` has more than one entry — `crates/aprender-train-canary` among them — and the bumper says nothing about them.

Concretely today: after `bash scripts/bump-version.sh 0.64.0`, the root workspace and all 65 pins read 0.64.0 while `crates/aprender-train-canary/Cargo.toml:3` still reads `version = "0.63.0"`.

**Live impact: none.** That crate is `publish = false`, absent from crates.io, and nothing pins its version — verified today. It is not a cascade risk.

**Why it is still worth fixing:** this is the guard-universe shape the repo keeps finding — a tool that enumerates its universe by hand instead of deriving it. The bumper's own header explains that a prose reminder is not a mechanism and that the script is the mechanism; the exclude list is the same argument one level up. The next excluded crate that carries a version pin which *must* track will be missed silently, and the R3/R5 backstop only covers `crates/facades`.

**Fix:** derive the set from `exclude = [...]` in the root manifest, then for each excluded path either bump its `[package] version` / upstream pins or assert-and-report why it is deliberately independent (the way `crates/facades`'s own 0.4.0 line already is). Mutation for the fix: add a second excluded crate carrying an upstream pin; the bumper must either move it or name it as deliberately skipped, never silently leave it.

Refs #2559, #2546

noahgift:
PP-066 triage 2026-09-05: **in 0.66 scope, no row yet** — row `G-5` proposed on #2873 (scripts/bump-version.sh on origin/main handles crates/facades by name and says nothing about the rest of the root `exclude` list (aprender-train-canary still read 0.63.0 after the 0.64.0 bump). The v0.66.0 cut uses this script (memory project_v0_64_0_release_cut); the class row G-5 derives the excluded set from the manifest.)

Instance of class #2878 — universe authored, not derived — targets, scopes and lists enumerated by hand so what is not on the list is never built, linted or run.

_PP-066 triage 2026-09-05 (issue #2873); ledger `docs/audits/pp-066-triage-2026-09-05.md`._

noahgift:
[...]

PR #2656, opened 2026-08-23
chore(release): 0.64.0 — the audit release

noahgift:
The version bump for **v0.64.0**. 67 commits since 0.63.0, four of them integration batches folding a further 44 reviewed PRs.

## Bumped with the repo's own tool, not a bare `cargo set-version`

`scripts/bump-version.sh 0.64.0` — which exists precisely because `cargo set-version --workspace` **cannot reach** `crates/facades/` (it is `exclude`d from the root workspace), and `check_facade_compat.sh` rows R3/R5 are wired into CI. A bump that skips those three files reds its own release commit (#2559).

Verified after the run:
- both facade `upstream` pins at 0.64.0
- `crates/facades/Cargo.lock` current under `--locked`
- `crates/facades`'s **own** version left at **0.4.0** — independent by design (#2546)

| guard | result |
|---|---|
| `check_facade_compat.sh` (+ `--self-test`) | exit 0 |
| `check_facade_order_gate.sh` | exit 0 |
| `check_lockfile_current.sh` | exit 0 (`cargo metadata --locked` resolves) |

## CHANGELOG

Every issue number cited in the new entry was resolved against the tracker before committing — **44/44**. (My first verification pass silently checked nothing, because zsh does not word-split unquoted variables and the whole list arrived as one token; redone in bash.)

The theme is continuous: v0.62.0 fixed gates that could not fail, v0.63.0 asked whether the thing validated was the thing shipped, and 0.64.0 is what that question found everywhere else — `--assert-tps` divided by ten, 7.2 KB of security policy running in no workflow, a falsification suite that had never scanned a crate, a required check decided by directory order, a meta-guard that passed on a comment. With the user-visible half beside it: SSE deleting every space (`Thequickbrownfox`), Ollama clients dropping every response over a bare-epoch timestamp, six dead native routes, a 1 KiB fragment of a 991 MB model certified `valid: true`, 443 files missing from a published crate.

## Deliberately not in this commit

`crates/aprender-train-canary` stays at 0.63.0 — workspace-excluded, `publish = false`, absent from crates.io, so not a cascade risk. The bumper knows about only *one* excluded workspace by name; filed as **#2655** rather than fixed by an unscoped edit inside a release commit.

Refs #2546 · #2559 · #2655

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,86 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.64.0] - 2026-08-23
+
+The **audit** release. v0.62.0 fixed gates that could not fail and v0.63.0 asked
+whether the thing we validated was the thing we ship. This one is what the same
+question found when it was pointed at everything else: 67 commits, four of them
+integration batches folding a further 44 reviewed PRs.
+
+The pattern that keeps recurring is a gate reporting success over a measurement
+it never made. `apr qa --assert-tps` divided its threshold by ten or discarded
+it outright (#2372). `deny.toml` was 7.2 KB of security policy that ran in no
+workflow at all, and had drifted until it failed on main (#2523). FALSIFY-MONO-011
+had never once scanned a crate (#2476). The README contract-count guard ran
+nowhere (#2485). A required check's verdict was decided by directory order
+(#2562). A meta-guard passed on a *comment*, so the MSRV floor ran nowhere
+(#2551). Every verdict `apr rosetta` printed was unfalsifiable — compare-inference,
+fingerprint and verify all exited 0 no matter what (#2420). Make recipes ran
+without `pipefail`, so the release gate could not fail (#2550).
+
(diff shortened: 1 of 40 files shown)


noahgift:
Duplicate — the operator had already cut the bump this morning at c6c4d36d8 on `release/0.64.0` (11:56), which I only found after pushing this. Independent confirmation that both are correct: the manifest, `Cargo.lock` and `crates/facades/` contents are **byte-identical** between the two bumps, which also re-verifies that `scripts/bump-version.sh` is deterministic and that the #2559 facade trap was handled on both sides.

Not wasted: the two CHANGELOG entries turn out to be complementary rather than duplicate — this one cites 36 issues the other does not (derived from merged commit subjects since the 0.63.0 release commit), while the other cites 19 this one does not (PR numbers folded inside the four batch commits, invisible to my source). All 44 citations here were resolved against the tracker before committing. I am contributing the union to `release/0.64.0` rather than landing a second bump.

Also filed from this work: #2655 — `bump-version.sh` handles one excluded workspace by name and is silent about the rest of the `exclude` list.

Issue #2671, opened 2026-08-24
PARITY-004 (CUT 1d): a wired guard that no timing assertion can enter a required check — today this rests on a comment


noahgift:
Part of #2667. Independent of the others; can land first.

## The gap

The eleven wall-clock-gate failures are currently prevented by **a comment**. `scripts/unwired_guards_baseline.txt:12-19` explains that `check_multiplatform_dogfood.sh` is *"a RELEASE gate, not a PR gate. Deliberately unwired from CI."* Nothing mechanical enforces that. A single well-meaning PR adding it to `ci.yml` reintroduces the entire failure class — and the last time a ratio rewrite went into a required check it **blocked all 9 open PRs**.

Policy is what failed. This needs poka-yoke.

## Build

A guard that **FAILS** if `check_multiplatform_dogfood.sh` (or any bench/timing gate named in a registry) appears in `.github/workflows/ci.yml` or `pr-gate.yml`, plus a `scripts/unwired_guards_baseline.txt` entry for the new bench rows **stating the reason** rather than merely listing them.

## Mutation tests

1. Add the script name to a `run:` line in `ci.yml` → the guard must go **RED**.
2. **Then extend the guard's scope to the `Makefile` and RE-RUN the mutation there.** Per the standing scar (`check_apr_bin_pinned.sh`, #2360): extending a guard's scope requires re-mutating in the new scope — the old proof does not transfer, and the Makefile's `\t@cmd` form is exactly what a naive pattern misses.
3. **Discrimination:** unmutated tree → GREEN.

Refs #2667 · #2360

noahgift:
Closing — fixed on main: fixed by PR #2682: scripts/check_no_timing_in_required.sh — a declared registry (check_multiplatform_dogfood.sh, check_bench_receipt.sh, check_perf041_marker.sh) may not appear in ci.yml/pr-gate.yml, with a vacuity floor of ≥2 resolved guards; wired in ci.yml

The issue's Makefile-scope mutation is moot on origin/main: ci.yml invokes no make target (grep shows comments only), so the Makefile cannot enter a required check. Guard exists, registry-driven, mutation-tested in the PR body.

Reopen if the pointer does not answer it.

_PP-066 triage 2026-09-05 (issue #2873); ledger `docs/audits/pp-066-triage-2026-09-05.md`._

Issue #2667, opened 2026-08-24
EPIC APR-PARITY-001: never let a performance gap reach production — instrument, arm, then close the llama.cpp gap


noahgift:
**Goal:** performance parity with llama.cpp, verified before *and* after every release, such that a performance regression cannot reach production unnoticed.

Product of a 6-angle quorum + adversarial review (2026-08-24). Design corrections to `APR-BENCH-RFC-001` (paiml/infra#259) and #2588 are recorded per-ticket. **Do not implement from the RFC as drafted** — three of its rules were falsified by measurement; see PARITY-008.

## The finding that shapes everything

A post-publish llama.cpp ratio is **not merely unwise — it is uncomputable.** `cargo install aprender` builds CPU-only (`crates/apr-cli/Cargo.toml` `default = ["hf-hub","safetensors-compare","inference","training","visualization","zram"]` — no `cuda`, no `wgpu`), while the comparator runs CUDA on lambda/gx10 and Metal on mini. The first four-host ratio would read **~0.05–0.10**, nobody would red a release over it (correctly), the row goes EXISTENCE-ONLY, and **the threshold never arms**.

The tree already documents this exact collapse at `crates/apr-cli/src/dispatch.rs:165`:
> reports `ratio_median=0.070x` … a fabricated 14x regression with nothing wrong in apr's decode path

That is the `check_multiplatform_dogfood` shape — a gate that exists, is correct, and never fires. So the two phases measure **different things**:

| phase | subject | comparator | why |
|---|---|---|---|
| **pre-publish** | the tree, `--features cuda` | llama.cpp | the only place a GPU-vs-GPU ratio is computable |
| **post-publish** | the crates.io artifact, 4 hosts | **none** — apr-vs-apr self-ratchet | CPU-class only; catches *our* regressions without a cross-class ratio |

## What already exists (verified, not assumed)

- **The measurement core already ships in the binary.** `crates/aprender-serve/src/lib.rs:197` `pub mod bench;` and `:208` `pub mod bench_preflight;` are unconditional — typed, serde-serializable, with a llama.cpp adapter, `ThermalGuard`, CV-based dynamic sampling, Welch's t-test and a `RegressionDetector`. This dissolves the repo-less-host problem: a host that only ran `cargo install aprender` **already has the measurement code**.
- realizar's 13 benches are **byte-identical** in `crates/aprender-serve/benches/`. The mining already happened.
[...]

PR #2681, opened 2026-08-24
feat(ci): a release-time timing gate can no longer become a required check — mechanism, not a comment (PARITY-004)

noahgift:
Closes **#2671**. Part of **#2667**. Stacked on #2680.

## The gap

Eleven wall-clock assertions have failed in a required check here. All three obvious remediations were tried and failed: **widening** stops the gate detecting what it exists for; **rewriting as a ratio** blocked all 9 open PRs; **`#[ignore]`** is banned.

The remedy that worked was **placement, not tuning**. Until now that placement was protected by *a comment* in `scripts/unwired_guards_baseline.txt`. Nothing stopped the next well-meaning PR from wiring it. Policy is what failed the other eleven times.

## Three parts, because the obvious spelling isn't the only one

| part | asserts | vacuity guard |
|---|---|---|
| **1** | registry guards are not **named** in `ci.yml` / `pr-gate.yml` | registry resolving <2 guards FAILS |
| **2** | a bench/timing guard **outside** the registry is unclassified → FAIL | — |
| **3** | the **transitive** path: a `make` target a required workflow invokes | no Makefile → FAIL, not pass |

PART 2 exists because otherwise the gate protects only what someone remembered to list — the guard's-universe-from-the-wrong-side shape.

## PART 3 is re-mutated in its own scope

Per **#2360**: extending a guard's scope requires re-proving it in the new scope; the old proof doesn't transfer. The Makefile's **`\t@cmd`** form is precisely what a `run:`-oriented pattern misses, so the mutation uses that form.

| mutation | result |
|---|---|
| wire `check_multiplatform_dogfood.sh` into a `ci.yml` `run:` line | **RED** |
| shrink the registry to one entry | **RED** ×2 (floor *and* PART 2, independently) |
| `\t@bash scripts/check_multiplatform_dogfood.sh` into `tier3` | **RED** — `tier3->check_multiplatform_dogfood.sh` |
| no-op control ×3 | GREEN |

## The guard caught itself

On its first run PART 2 reported `check_bench_receipt.sh` — which **is** registered — as unclassified: the registry is newline-separated and the space-delimited `case` never fired. Normalised, with the reason at the call site.

`check_bench_receipt.sh` joins the baseline with its reason stated (it validates receipts that exist only for a version being cut). The meta-guard and its case table stay green. bashrs 0 SEC/DET/IDEM.

Refs #2667 · #2671 · #2360


noahgift:
Superseded by #2705.

Issue #2675, opened 2026-08-24
PARITY-008: BENCH-003's threshold formula is falsified by the repo's own data — derive by bootstrap instead


noahgift:
Part of #2667. **Correction to `APR-BENCH-RFC-001` and #2588 — do not implement BENCH-003 as drafted.**

## What was falsified, by recomputation

**BENCH-003 as written:** `threshold_host = 3 × pooled_relative_stddev`.

- It returns **GREEN on the only documented regression in this repo** — derived 19.836% vs actual 19.306%. The gate would not have caught the one thing we know went wrong.
- **Its power *falls* as data accumulates.** More evidence makes it weaker. A threshold that gets less sensitive the more you measure is anti-inductive — a stronger objection than the numerical miss, because it means the gate degrades precisely as the fleet gets better instrumented.

**Also rejected: a t-test (`ttest_rel` / `welch_t_test`) as the *blocking* rule.** At n=7 a significance test fires on real-but-irrelevant drift and rebuilds the treadmill that produced the eleven wall-clock failures. Keep it as an **advisory verdict** only.

## Replace with

Threshold **derived by bootstrap over the recorded raw samples** — the method `contracts/beat-ollama-decode-throughput-speed-v1.yaml:52-55` already used to size median-of-7 at a ~0% false-FAIL rate. This is why PARITY-001's `samples_ms` is load-bearing: bootstrap cannot run on summary statistics.

Armed **per host, once ≥3 receipts carrying a bench block exist at `origin/main` for that host** — so the floor accumulates from real cuts and no human ever types a number. Set as a **no-collapse floor**, not a tight band.

## Mutation test

Revert the aarch64 SIMD fix for #2567 (or, without a build, halve gx10's sample vector in a scratch receipt) → **the gx10 row must go RED while lambda and intel stay GREEN.** The gate must indict one host, not the release.

## Also corrected

`load_state: idle|loaded` as a two-valued human label is wrong on intel: under a CPUQuota, load average counts work being *prevented*. Record `cpu.stat` `nr_throttled`/`throttled_usec` deltas, PSI `cpu full avg10`, and governor state instead.

Refs #2667 · #2588 (BENCH-003, BENCH-004) · paiml/infra#259

noahgift:
[...]

Issue #2588, opened 2026-08-22
Epic: APR-BENCH — the four-host throughput ratchet (BENCH-002..009), and the correction that it currently measures llama.cpp


noahgift:
Tracking the aprender-owned half of `APR-BENCH-RFC-001` (paiml/infra#259). The RFC's own
§10.2 assigns 6 of 9 tickets here; this issue carries them plus one blocking correction
found during design review.

**Design resolution is recorded on paiml/infra#259** — read that comment first. Summary:
the bench is a **field on Gate 12**, not a Gate 13; the comparand is the prior release's
receipt **at `origin/main`**; existence blocks unconditionally while value breaches are
payable by an owned, expiring record.

---

## BENCH-009 — BLOCKING, blocks BENCH-002: the gate measures llama.cpp, not apr

RFC §5.3 gives **only a `llama-bench` command**; §9's ledger row carries **no apr
invocation and no apr version field**. §7.1 would therefore block an aprender release on
the **comparator's** throughput, and pass GREEN through any apr regression.

Fix: every row carries **both sides**; the gated quantity is the **ratio**, with
llama.cpp as the **control** that cancels thermal and contention noise on a 150 W-capped,
16-runner fleet. llama.cpp is currently the subject; it must be the control.

## BENCH-002 — measurement envelope

Blocked by BENCH-009. Must record apr's own decode tok/s **and** the comparator's, per
host, with verbatim commands both sides.

## BENCH-003 — noise floor → per-host thresholds

Do **not** run this as a campaign. Stage B of the ladder **derives**
`threshold_host = 3 × pooled_relative_stddev` once ≥3 prior receipts exist on `main`, so
no human ever types a number and the provenance invariant holds mechanically. The floor
accumulates from real cuts as a side effect.

## BENCH-004 — determinism policy

`n ∈ [5,20]`, `runs_discarded == 0`, `load_state` recorded. A host whose variance exceeds
a useful threshold gets **EXISTENCE-ONLY** status, and that is **recorded, not hidden**.

## BENCH-005 — the canonical model is untracked and unignored

Confirmed: `qwen2_5_coder_1_5b_instruct_q4k` is **1,330,247,968 bytes**, untracked,
unignored, and has **no `.gguf` extension** — so `.gitignore:47` (which catches
`test.gguf`) does not match it. One `git add -A` commits 1.3 GB.

**Also, a gap the RFC never names: the canonical GGUF is staged on ZERO of four hosts.**
[...]

PR #2684, opened 2026-08-24
feat(bench): derive the regression threshold by bootstrap — the RFC's rule is falsified by execution (PARITY-008)

noahgift:
Closes **#2675**. Part of **#2667**. A **correction** to `APR-BENCH-RFC-001` / #2588 BENCH-003 — do not implement `threshold_host = 3 × pooled_relative_stddev`. Stacked on #2683.

**The falsification runs on every CI pass**, so the correction cannot quietly rot back. The guard re-derives both numbers rather than asserting a conclusion.

## Reconstructed the documented case exactly

The RFC's band is derived from **pooled** samples, and pooling is what breaks it: two hosts that merely *differ* — neither unhealthy — inflate the dispersion until the band swallows a real regression.

| | |
|---|---|
| two hosts differing ~11.9% → pooled rel. stddev | **6.58%** |
| 3σ band | **19.739%** (documented: 19.836%) |
| actual regression | **19.306%** → *inside* the band → **GREEN** |
| bootstrap floor (per host) | 409.826 tok/s |
| regressed median | 332.701 → below the floor → **RED** |

The rule as drafted returns GREEN on the one regression this repo has on record.

## What I got wrong twice — kept in the source, because the wrong turns are the argument

**Attempt 1** asserted *"the floor tightens as n grows"*. **Refuted by its own numbers**: tolerated drop was 0.56 tok/s at n=3 and 1.12 at n=50 — it got **looser**. That's the bootstrap being honest: three samples can't exhibit a distribution's true dispersion, so a small-n floor is *optimistic*.

**Attempt 2** asserted convergence but measured it from **one draw per n**, where the swing is dominated by which numbers were drawn. Also refuted.

**What is actually true**, and what the guard now checks: across **independent draws** at a given n the derived floor *varies*, and that variance shrinks with n — spread **1.6031 at n=3 → 0.5488 at n=33** over 40 draws. A small-n floor is not merely optimistic, it is **unstable**: two honest campaigns would arm two different gates. That's a stronger argument for a minimum-n rule than the RFC's bare `n ∈ [5,20]`, because it says *why*.

I could have tuned the test until it agreed with the prose. The numbers are recorded instead.

## Mutation

Replace the bootstrap floor with the 3σ-pooled floor in the falsification → guard **RED** (*"bootstrap did not catch the documented regression"*). Restored → **GREEN**.

Fixed here too: this branch was cut from parity-001 rather than the stack tip, so the wiring anchor was absent and `check_no_timing_in_required.sh` exited **127** (missing). Reset onto the tip; all three meta-guards green.

Determinism: fixed seed — a threshold that moves when nobody measured anything is not a threshold.

Refs #2667 · #2675 · #2588 · paiml/infra#259


noahgift:
Superseded by #2705.

Issue #2679, opened 2026-08-24
F12 — fabricated measurement: a value with the form of a measurement and nothing behind it


noahgift:
New finding class for APR-QUALITY-001, alongside F11 (dark capability).

## Definition

**F12 — fabricated measurement.** A value carrying the form, units and provenance-shape of a measurement, produced without the measurement having been taken.

It is distinct from its neighbours:
- **not F7 (silent pass)** — nothing passes falsely; the number may even be *correct* for some past run.
- **not F9 (coupled falsifier)** — there is no coupled oracle. There is **no oracle at all**.

The harm is that it is indistinguishable from evidence at the point of consumption, and it survives review because it *looks* like the thing it replaces.

## Instances found 2026-08-24

1. `scripts/benchmark-2x-ollama.sh:27-29` — `OLLAMA_BASELINE=291`, `OLLAMA_SINGLE=120`, `OLLAMA_CPU=15` hardcoded; ollama never invoked. (PARITY-005)
2. `scripts/benchmark-matrix.sh:396` — the same three literals emitted into JSON as `ollama_baselines`.
3. `crates/aprender-serve/benches/external_matrix.rs:331-336` — `HardwareSpec { cpu: "Benchmark CPU", gpu: Some("Benchmark GPU"), memory_gb: 32 }` written into the receipt. (PARITY-007)
4. `crates/aprender-serve/src/bench/backend.rs:140` — `version: "b2345"  // Would be detected from binary`; also :226, :336.

## The general remedy

A field that *could* be measured must be **derived or absent** — never asserted. Where derivation is impossible, the receipt records the absence explicitly rather than a placeholder, and the consuming gate treats a placeholder as RED.

Detection heuristic worth mechanizing: a literal in a receipt-producing path whose name matches a measurable quantity (`*_baseline`, `version`, `cpu`, `gpu`, `memory_*`) is a candidate. The `check_no_hand_rolled_parsers.sh` construct-ban shape applies.

Refs #2667 · PARITY-005 · PARITY-007 · APR-QUALITY-001

noahgift:
Closing — fixed on main: superseded by docs/audits/f12-fabricated-measurement.md on origin/main (the class definition and its four instances, catalogued 2026-08-24) and scripts/check_no_fabricated_baselines.sh (the construct ban); evidence/parity/LEDGER.md rows 9–11 apply F12 ('block REFUSED, F12')

The finding class is recorded where the issue asked (repo-side record) and mechanised.

[...]

Issue #2674, opened 2026-08-24
PARITY-007: the only apr-vs-llama.cpp bench in the tree is dark on three axes and its apr side is dead code


noahgift:
Part of #2667.

`crates/aprender-serve/benches/external_matrix.rs` is the one bench that measures apr vs llama.cpp vs ollama. It is dark three ways, each independent:

1. **The feature is never built.** `crates/aprender-serve/Cargo.toml:231` declares `bench-http = ["dep:reqwest"]`; `grep -rn bench-http .github/ Makefile scripts/` returns **nothing**.
2. **The default build registers only one side.** `external_matrix.rs:551` — `#[cfg(not(feature = "bench-http"))] criterion_group!(benches, benchmark_realizar_native,);`
3. **The apr side is dead code.** `external_matrix.rs:133-134` — `/// FIXME: GGUFTransformer no longer has forward() - use OwnedQuantizedModel` followed by `#[allow(dead_code)] fn benchmark_realizar_native_cv`.

## And it fabricates host identity

`external_matrix.rs:331-336`:
```rust
HardwareSpec { cpu: "Benchmark CPU", gpu: Some("Benchmark GPU"), memory_gb: 32 }
```

The `BenchmarkMatrix` receipt schema exists, but **nothing forces its fields to be measured**. F12 class again (see PARITY-005), inside the Rust harness this time.

**Poka-yoke required:** `HardwareSpec` must be constructed from the host, never from a literal — otherwise the four-host receipt carries placeholder provenance while looking complete.

Also in scope: three fabricated comparator versions in `crates/aprender-serve/src/bench/backend.rs` — line 140 `version: "b2345"  // Would be detected from binary`, line 226 `"0.4.0"`, line 336. Each must be **detected or absent**, never asserted.

## Mutation test

Construct a receipt on a host whose CPU is known, with `HardwareSpec` still carrying the literal → must **FAIL**. Discrimination: a host-derived spec on the same run → GREEN.

Refs #2667 · PARITY-005

noahgift:
[...]

Issue #2672, opened 2026-08-24
PARITY-005 (F12): 1,086 lines of benchmark script that fabricate the comparator — TRIAGE, then delete


noahgift:
Part of #2667. First instance of a new finding class.

## F12 — fabricated measurement

`scripts/benchmark-2x-ollama.sh:27-29` hardcodes the comparator and **never invokes ollama**:

```
OLLAMA_BASELINE="${OLLAMA_BASELINE:-291}"   # GPU batched baseline
OLLAMA_SINGLE="${OLLAMA_SINGLE:-120}"       # GPU single-request baseline
OLLAMA_CPU="${OLLAMA_CPU:-15}"              # CPU baseline
```

`scripts/benchmark-matrix.sh:396` emits the same three frozen literals into its JSON as `ollama_baselines`.

**This is not F7 (silent pass — nothing passes falsely) and not F9 (coupled falsifier — there is no oracle at all).** It is a value with the *form* of a measurement and nothing behind it. New class, to be added to APR-QUALITY-001 alongside F11.

## The four scripts, all verified unreferenced by any workflow or the Makefile

| script | lines | defect |
|---|---:|---|
| `benchmark-2x-ollama.sh` | 481 | fabricates the comparator |
| `benchmark-matrix.sh` | 408 | same three literals into JSON; CV from n=3 with a **population** stddev |
| `gpu_2x_benchmark.sh` | 171 | shells into `REALIZAR_DIR=../realizar`, a checkout APR-MONO consolidated away |
| `verify-parity.sh` | 26 | wraps `cargo run --example qa_run -- --matrix` — **benchmarks nothing**; its NAME is the highest-value trap for anyone implementing this epic |

Independently confirmed: `grep -rl` across `.github/workflows/` and `Makefile` returns empty for all four. Last touched 2026-05-07 in one unrelated mass-touch commit.

## TRIAGE BEFORE DELETE

**Do not bulk-delete.** #2640 taught that all 9 hunks of an 86-line "drift" were hardenings. The bar is genuinely lower for fabricated-constant scripts — but 1,086 lines is a lot, and `ThermalGuard`-adjacent logic or discard rules may be buried in there. Produce a triage table (port / discard, with the reason) **before** any deletion, the way `docs/audits/dogfood-divergence-2640.md` did. Cheaper than last time, and it is the same rule.

Refs #2667 · #2640 · APR-QUALITY-001

noahgift:
[...]

PR #2687, opened 2026-08-24
docs(audit): catalogue F12 — fabricated measurement, its four instances and its mechanisms

noahgift:
Closes **#2679**. Part of **#2667**. Stacked on #2686.

The class belongs to `APR-QUALITY-001`; this is the **repo-side record**, so a future reader can find the instances and the mechanism from the code rather than from a spec document they may not have.

## F12 — fabricated measurement

*A value carrying the form, units and provenance-shape of a measurement, produced without the measurement having been taken.*

| | why F12 is not this |
|---|---|
| **F7** silent pass | Nothing passes falsely — the number may even be *correct* for some past run. |
| **F9** coupled falsifier | There is no coupled oracle. **There is no oracle at all.** |
| **F10** stale self-description | A stale description was once true of something. This was never a measurement. |

Its specific damage to a ratio: the denominator becomes a constant of unknown date, host and build, so it stops measuring change over time **while continuing to report one**.

## Four instances, one afternoon

| # | site | asserted |
|---|---|---|
| 1 | `benchmark-2x-ollama.sh:27-29` | `OLLAMA_BASELINE=291` … **ollama never invoked** |
| 2 | `benchmark-matrix.sh:396` | same three literals, into JSON |
| 3 | `external_matrix.rs:331-336` | `cpu: "Benchmark CPU"`, `memory_gb: 32` |
| 4 | `backend.rs:140,226,336` | `version: "b2345"  // Would be detected from binary` |

1–2 deleted (#2672, triage first); 3–4 replaced by detection (#2674).

**The comment on instance 4 is the most useful thing here.** *"Would be detected from binary."* The author knew. F12 is rarely deception — it's a placeholder that outlived the intention to replace it, which is the argument for a **mechanism over a review**.

## Recorded honestly

The shell guard's pattern was **wrong twice** before it was right: the first version missed the JSON-literal spelling entirely; the second used a suffix allowlist that missed `OLLAMA_CPU=15` — *one of the three live instances*. Neither was visible on review. That's why it ships a case table.

Refs #2679 · #2667 · #2672 · #2674


noahgift:
Superseded by #2705.

Issue #2676, opened 2026-08-24
PARITY-009 (CUT 3): pin llama.cpp as the FOURTH token of the existing mechanism — it is rediscovery #5 of a rule already stated once


noahgift:
Part of #2667. Closes the open half of #2647.

## Why this is not new work

`scripts/verifier_pin.sh:36` already lists the unpinned llama.cpp comparator as **instance 5** of the five-instance unpinned-verifier table — cited, never enforced. The rule ("a gate that decides a release must not resolve its verifier through PATH") is stated once in that file and enforced by `check_verifier_pinning.sh`. llama.cpp becomes the **fourth token** in that mechanism, not a new one.

An unpinned comparator makes every ratio meaningless across time: the denominator changes silently between releases.

## Build

`scripts/llama_bin.sh`, modelled on `scripts/apr_bin.sh`: option-neutral, sourceable (`. scripts/llama_bin.sh || exit 1`), fails by **return status**, refuses PATH resolution and refuses a hardcoded absolute path.

Freshness proven by **behavior, not `--version`** — the recorded scar is that two pmat binaries both printed `3.32.0` while running different commits (#2384). Record `build_commit` **and** `reported_version` separately, exactly as the receipt schema's `provenance` block does.

`build_flags` must be captured **verbatim**: gx10 requires `-DGGML_CUDA_ARCHITECTURES=121`; omitting it JITs from stale PTX, which **produces correct output slowly and fires no gate** — a performance defect invisible to every correctness check.

## Mutation tests

1. Swap the `llama-bench` binary for a different build → the pin must go **RED on `build_commit` mismatch** (not on version, which is identical).
2. Point the pin at a directory, or at a stub that is `-x` but exits 1 → **RED**. Existence is not a pin (the VPIN-2 lesson).
3. Add `llama` to `check_verifier_pinning.sh`'s token set and re-run its **must-match/must-not-match case table** — the table is the ratchet; that walker has been wrong sixteen times.

Refs #2667 · #2647 · #2384 · #2653

noahgift:
Closing — fixed on main: fixed by PR #2685: scripts/llama_bin.sh (sourceable, never PATH), scripts/llama_pin.toml (build_commit pinned; 39173bcac at LEDGER rows 0a/5/7/8), scripts/check_llama_pin.sh (pin_stale/pin_fresh); master §12 row 3 LANDED

Bold 'Closes **#2676**' did not auto-close; the pin is the comparator declaration every LEDGER row now cites.

[...]

Issue #2673, opened 2026-08-24
PARITY-006: nightly-bench.yml is a triple no-op — fix the probe; do NOT delete until PARITY-003 lands


noahgift:
Part of #2667. **Sequenced after PARITY-003.**

## Three independent reasons it measures nothing

1. **The probe finds nothing.** It delegates to `paiml/.github/.github/workflows/sovereign-ci.yml@main` with `run_benchmarks: true`; that job gates on `ls benches/*.rs`. aprender has **no root `benches/`** (`ls -d benches` → No such file or directory). All **120** bench `.rs` files live under `crates/*/benches/`. The else branch prints *"No benches/ directory — skipping"*.
2. **The job cannot fail** — `continue-on-error: true` on the whole bench job.
3. **Nothing compares two runs.** Output goes to a 90-day `bench-results.txt` artifact and is never read back. `grep -rn "cargo bench" .github/` returns nothing in aprender.

Net: **37 criterion bench targets have never been executed by aprender CI.**

## Fix

Point the probe at `crates/*/benches`, remove `continue-on-error`, and route results into the receipt path from PARITY-003 so two runs can actually be compared.

## Sequencing — do not invert this

Deleting `nightly-bench.yml` outright removes the **only** bench lane, no-op though it is. Land PARITY-003's receipt path **first**, or there is a window with nothing at all — and a triple no-op at least leaves an artifact trail that a missing job does not.

If the probe cannot be fixed (it lives in an upstream reusable workflow, so a change there is a cross-repo edit), the fallback is a repo-local lane that invokes the bench targets directly — but still only after PARITY-003.

Note: editing `.github/workflows/*` requires operator check-in per CLAUDE.md.

Refs #2667 · #2588

noahgift:
Closing — fixed on main: superseded: .github/workflows/nightly-bench.yml on origin/main is a repo-local lane (header cites aprender#2673) that runs check_bench_receipt.sh, check_bench_threshold.sh, check_llama_pin.sh and the #2567 GEMV measurement with 'silence is red', and declares what it does not run; PARITY-003's receipt path (PR #2688) landed first as the issue's sequencing required

All three no-ops named by the issue are addressed by replacement rather than by fixing the upstream probe, which the issue allowed as the fallback.

Reopen if the pointer does not answer it.

_PP-066 triage 2026-09-05 (issue #2873); ledger `docs/audits/pp-066-triage-2026-09-05.md`._

Issue #2670, opened 2026-08-24
PARITY-003 (CUT 1c): a CPU-class bench block on the four-host receipts — apr-vs-apr self-ratchet, no comparator


noahgift:
Part of #2667. Depends on PARITY-001, PARITY-002. **Blocks PARITY-006** (see sequencing).

## What and why

Add a `bench` block to `evidence/dogfood/<version>/<host>.json`, produced by the **just-installed binary in the same sitting** as `install_wall_seconds` — no repo checkout required (mini has no runner at org or repo scope).

**CPU-class only. No comparator.** This is deliberately *not* a llama.cpp ratio: `cargo install aprender` is CPU-only on all four hosts, so a cross-class ratio would read ~0.07 and be born disarmed (#2667). An **apr-vs-apr self-ratchet** catches our own regressions — which is the actual goal — without inventing a number nobody will act on.

This works because the measurement core already ships in the binary: `crates/aprender-serve/src/lib.rs:197` `pub mod bench;` is unconditional.

Gate rows added to `scripts/check_multiplatform_dogfood.sh`, which is the correct attachment point: it already implements four-host receipts with a shrink-never host set whose comparand is `git show origin/main:` of itself, and it is **deliberately not a required PR check** — so no timing assertion ever enters `ci / gate`, satisfying the eleven-failure constraint for free.

## Mutation tests (six RED, each paired with a GREEN restore, plus a no-op discrimination check)

1. Delete one host's `bench` block → **RED**.
2. Replace samples with seven identical values → **RED** (a fabricated-constant detector; F12 class).
3. `bench` block present but `samples_ms` empty → **RED** (vacuous pass).
4. Receipt version ≠ the version under test → **RED** (stale).
5. `compute_class` disagreeing with the recorded `feature_set` → **RED**.
6. A receipt whose `binary_sha256` does not match the installed binary → **RED**.
7. **Discrimination:** unmutated four-host set → **GREEN**.

Refs #2667 · #2588 (BENCH-002) · #2658

noahgift:
Closing — fixed on main: fixed by PR #2688 (merged 2026-08-25): scripts/bench_host_receipt.sh produces the CPU-class apr-vs-apr bench block; scripts/check_multiplatform_dogfood.sh gains the bench rows; scripts/lib/bench_receipt.py --has-bench/--bench/--bench-median is the single validator

[...]

PR #2690, opened 2026-08-24
fix(ci): the nightly bench lane now MEASURES, and can fail (PARITY-006)

noahgift:
Closes **#2673**. Part of **#2667**. Stacked on #2689. **Sequenced after #2670** as you required — the receipt path exists before the old lane is replaced, so there's no window with nothing.

## What was there — a triple no-op

`nightly-bench.yml` delegated to upstream `sovereign-ci.yml` with `run_benchmarks: true`:

1. **The probe found nothing.** It gates on `ls benches/*.rs`; aprender has **no root `benches/`** — all **120** bench files live under `crates/*/benches/`. Else branch: *"No benches/ directory — skipping"*.
2. **The job could not fail** — `continue-on-error: true` on the whole thing.
3. **Nothing compared two runs.** Output went to a 90-day artifact, never read back.

Net: **120 bench targets never executed by aprender CI, behind a green tick.**

Fixing the probe upstream is a cross-repo edit and would still leave (2) and (3). This lane is repo-local, small, and fails.

## What it runs — and what it says it doesn't

Runs: the receipt validator's case table · the threshold falsification (so the BENCH-003 correction can't rot back) · the comparator pin's case table · the #2567 measurement, with `grep -q '^speedup'` so **silence is red**.

**Declares its own gaps**: the other 120 criterion targets, any apr-vs-llama.cpp ratio (pre-publish, needs `--features cuda`), and aarch64/macOS. Scope is small on purpose — 150 W-capped fleet, 16 runners on one box, and two concurrent workspace-tests starved it on 2026-08-23.

## Two defects found by *running* the steps, not reasoning about them

- **My pin step branched on `rc == 2`** to treat UNPINNED as a notice. **Dead code** — that value comes from `llama_bin.sh`'s resolver, never from the checker, which exits 0 when its case table discriminates. The step printed **"pinned"** on a tree whose declaration says **UNPINNED**. Removed.
- **The measurement prints 61–77× in release on x86**, and that number means almost nothing for #2567: there `matmul_q4k_f32_parallel` **is** the AVX path, so the ratio is SIMD *and* threads against a pure-scalar baseline. On aarch64 both sides are the same scalar kernel and the ratio isolates parallelism — which is why the figure quoted for #2567 is the aarch64 **1.21×**. The test now prints that caveat itself.

Every step executed locally before commit.

Refs #2667 · #2673 · #2670 · #2675 · #2676 · #2567

### DIFF
--- a/.github/workflows/nightly-bench.yml
+++ b/.github/workflows/nightly-bench.yml
@@ -1,21 +1,95 @@
-# Nightly Benchmarks — criterion performance tracking
-# Calls sovereign-ci.yml with run_benchmarks: true
+# Nightly Bench — a lane that MEASURES, and that CAN FAIL.
+#
+# WHAT THIS REPLACED, and why it had to be replaced rather than tuned. The old
+# file delegated to paiml/.github sovereign-ci.yml with `run_benchmarks: true`,
+# which was a TRIPLE no-op for this repo (aprender#2673):
+#
+#   1. The probe found nothing. That job gates on `ls benches/*.rs`, and
+#      aprender has NO root benches/ — all 120 bench .rs files live under
+#      crates/*/benches/. The else branch printed "No benches/ directory —
+#      skipping".
+#   2. The job could not fail: `continue-on-error: true` on the whole thing.
+#   3. Nothing compared two runs. Output went to a 90-day bench-results.txt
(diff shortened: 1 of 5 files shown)


noahgift:
Superseded by #2705.

Issue #2678, opened 2026-08-24
PARITY-011: the pre-release and post-release dogfood protocols gain their bench phase — and say which subject each one measures


noahgift:
Part of #2667. Lands after PARITY-003 (post) and PARITY-010 (pre). Closes the protocol half of #2662.

## What changes

**Pre-release (`.claude/skills/pre-release/SKILL.md`, Gate 12):** gains the llama.cpp comparator ratio as a **field on Gate 12**, not a new Gate 13. Gate 12 is release-time and deliberately not a required PR check, so no timing assertion enters `ci / gate` — the eleven-failure constraint is satisfied structurally rather than by discipline.

**Post-release:** gains the CPU-class self-ratchet block on the four-host receipts. This is the phase that never existed: 0.63.0 published 2026-08-01 with receipts dated 2026-08-22, and 0.64.0's four receipts were hand-carried by one person from one box on 2026-08-24.

## The rule the protocol must state

Every gate declares its **subject**: `tree` | `registry` | `published`. That is #2662's root cause — phase was an undeclared dimension, so gates whose subject was a later phase sat permanently red and their producing half went unwritten.

And the honest constraint, stated rather than discovered: **pre and post measure different things.** The post phase cannot compute a llama.cpp ratio because `cargo install aprender` is CPU-only. Writing that into the protocol prevents the next person re-deriving it.

## Do not let this become a second methodology

The skill's own rule is *"Never dogfood by hand — the protocol runs the binary. You do not."* The 0.64.0 sweep violated it because no tracked procedure existed. Whatever ships here must be the **one** runner's own output — `check_dogfood_shim.sh` already asserts exactly one tracked `dogfood.sh`, and that assertion must extend to cover a bench/fleet sweep so a helper script cannot reappear.

## Mutation test

Add a second script containing a fleet or install sweep → the shim check must go **RED**. Note SHIM-2644-03 (#2664): today that check's universe is `git ls-files`, so an **untracked** copy passes — fix that first or this guard has the same hole.

Refs #2667 · #2662 · #2658 · #2664

noahgift:
Closing — fixed on main: fixed by PR #2686: .claude/skills/pre-release/SKILL.md Gate 12 declares phase and subject (tree|registry|published), gains the bench field; scripts/check_dogfood_shim.sh extended; the post-publish self-ratchet block is PR #2688

[...]

Issue #2677, opened 2026-08-24
PARITY-010 (CUT 4): the pre-publish GPU comparator lane — the only place an apr-vs-llama.cpp ratio is computable


noahgift:
Part of #2667. **Only after PARITY-001/002/003 and PARITY-009 land clean.**

This is the half that actually answers *"are we at parity with llama.cpp?"* — and it can only run **pre-publish, from the tree**, where `--features cuda` exists. Post-publish is CPU-class self-ratchet only (#2667).

## Build on the one working harness

Generalize `crates/aprender-serve/tests/beat_ollama_decode_throughput_speed.rs` — the only working ratio harness in the fleet. It already implements the BENCH-009 fix (both sides measured, the **ratio** is the gated quantity, the comparator is the **control** that cancels thermal and contention noise) and the BENCH-004 determinism policy.

**Keep its 128/384 differential.** It is the only technique in the fleet that cancels apr's documented 3.4–3.9 s one-shot startup cost — `llama-bench` has no equivalent, so a naive single-length comparison charges apr for startup and reports a fabricated deficit.

## Fairness protocol (both sides verbatim in the receipt)

Enumerate and pin: quantization, KV-cache policy, batch vs single, prompt length, warmup, GPU layers offloaded, thread count, flash-attention on/off, tokenizer, sampling (greedy, fixed seed). This project has already been burned in **both** directions — the withdrawn 1.371× overclaim, and golden-gate argmax-margin flips.

A skeptical outsider must be able to reproduce both invocations from the receipt alone.

## Mutation test

Reuse `scripts/check_beat_measurements.sh` **unchanged** — it derives the expected marker set from the workflow and fails when any expected marker produced no line ("silence is red"). Mutation: remove one host's marker emission while leaving the assertion in place → must go RED.

Refs #2667 · #2588 (BENCH-009) · #2359

noahgift:
[...]

Issue #2669, opened 2026-08-24
PARITY-002 (CUT 1b): a pv contract for the bench receipt — compute_class required, and a cross-class run structurally cannot carry a threshold


noahgift:
Part of #2667. Depends on PARITY-001.

Add `contracts/apr-bench-regression-v1.yaml`, modelled on the **shape** of `contracts/beat-ollama-decode-throughput-speed-v1.yaml` (pillar / incumbent / canonical_task / metric / direction / baseline / threshold / sourced_date / approved_compute / ci_gating.mode) — with the one field that contract gets wrong replaced.

## The load-bearing rule

```
if   comparability == "cross-class-existence-only"
then NOT required: [threshold]
```

**A cross-class run cannot arm a threshold.** This is the born-disarmed failure made *unwriteable* rather than discouraged — the `check_multiplatform_dogfood` shape, where a row goes EXISTENCE-ONLY and the threshold never arms for any release.

`compute_class` is **required** in `provenance`. A receipt proving *which binary ran* but not *which dispatch path it took* catches the wrong-binary class (5 in-tree rediscoveries: PMAT_BIN, pv_bin.sh, apr_bin.sh, #2384, unpinned llama.cpp) and misses the wrong-compute-class one entirely — which is exactly how a CPU-only apr side against a CUDA llama.cpp side validates cleanly and produces the fabricated 14× regression.

## Mutation tests

1. Invent `status: FAST_ENOUGH` → `pv validate` must **fail to parse**.
2. Replace the structured comparator object with a free-text prose paragraph → must **FAIL**. *(This is the point: the existing contract carries the comparator as prose, so this mutation is a live defect, not a hypothetical.)*
3. A receipt with `comparability: cross-class-existence-only` **and** a `threshold` object → **REJECTED**.
4. `provenance` without `compute_class` → **REJECTED**.
5. Same-class run with `samples_ms` + `runs_discarded` → **ACCEPTED** (discrimination).

`threshold.source` must name both **rejected** methods so they cannot be silently reintroduced: 3× pooled relative stddev, and t-test-as-blocking-rule (see PARITY-008).

Refs #2667 · #2588 (BENCH-004)

noahgift:
[...]

Issue #2668, opened 2026-08-24
PARITY-001 (CUT 1a): apr bench --json must emit raw samples_ms, compute_class and model sha256 — it computes them and throws them away


noahgift:
Part of #2667. **First ticket — everything else waits on this data existing.**

## The defect

`crates/apr-cli/src/commands/bench.rs:287` already computes per-run samples, feeds them to percentiles, and **discards them**. The receipt therefore carries summary statistics only.

That is not a convenience loss. **Summary statistics cannot be resampled** — a receipt carrying only mean/stddev *permanently forecloses* bootstrap, which is the only correct threshold-derivation method available (see PARITY-008, where the RFC's 3σ formula is falsified). Discarding `samples_ms` destroys the sole input the correct method takes.

## What to emit

Per the patched receipt schema (`$defs/measurement`, `$defs/provenance`):

- **`samples_ms`** — raw per-run samples, verbatim.
- **`n`, `warmup`, `runs_discarded`** with a closed reason vocabulary `{early_eos, negative_delta, nonzero_exit}`. `runs_discarded` has `minimum: 0` and is **NOT** constrained to zero — the one working harness (`beat_ollama_decode_throughput_speed`) discards by construction on token-count inconstancy, so a zero-invariant would contradict the only correct implementation in the tree.
- **`compute_class`** ∈ `{cpu, cuda, metal, wgpu, unknown}` — **the dispatch path actually exercised, not the hardware present.** A CUDA-capable host running a default-features build is `compute_class: cpu`. This is the field whose absence would have let the fabricated 14× ratio validate cleanly.
- **model content `sha256`**, prompt digest, and the `provenance` block (`binary_path`, `binary_sha256`, `resolution: pinned|path`, `build_flags`, `feature_set`).

Reuse `crates/apr-cli/src/lib.rs:208-235` `emit_version_json`, which already emits `cuda_feature`, `cuda_runtime_available`, `visible_devices` and is contract-covered by FALSIFY-GPUTRAIN-007.

## Mutation tests (each RED paired with a GREEN restore)

1. Delete the `samples_ms` key from `print_bench_json` → the receipt gate must **FAIL on missing raw samples**.
2. Hardcode `compute_class: cuda` in a default-features build → must **FAIL** (it must be derived from the dispatch path, never asserted).
3. Emit a `model_sha256` that does not match the file on disk → **FAIL**.
4. **Discrimination control:** an unmutated run on the same host must be GREEN.

[...]

PR #2691, opened 2026-08-24
feat(bench): APR-PARITY-001 — never let a performance gap reach production (12 tickets, one merge-queue entry)

noahgift:
**The whole APR-PARITY-001 epic (#2667) as one merge-queue entry.** This branch is cumulative — all eleven other branches are ancestors, verified with `git merge-base --is-ancestor` — so merging this lands 12 tickets in one `workspace-test` run instead of twelve.

Closes **#2668 #2669 #2670 #2671 #2672 #2673 #2674 #2675 #2676 #2677 #2678 #2679**, and the parallelism half of **#2567**.

## The goal: never let a performance gap reach production

| # | ticket | what landed |
|---|---|---|
| 001+002 | receipt + contract | raw `samples_ms`, `compute_class`, model sha256, provenance; one validator proven by a **14-case table** |
| 003 | host bench block | CPU-class apr-vs-apr self-ratchet on the four-host receipts |
| 004 | placement | a timing gate can no longer reach a required check — including **transitively via `make`** |
| 005 | F12 | **1,086 lines deleted** after a triage table; the construct banned |
| 006 | nightly lane | the triple no-op replaced by a lane that measures **and can fail** |
| 007 | provenance | hardware and comparator versions **detected, never asserted** |
| 008 | threshold | **bootstrap**; the RFC's `3 × pooled stddev` falsified by execution |
| 009 | comparator | **llama.cpp pinned** — closes instance 5 of the unpinned table (#2647) |
| 010 | fairness | the protocol declared complete; an unstated knob is a silent degree of freedom |
| 011 | protocols | every gate declares its **subject**; a second fleet sweep is unwriteable |
| 2679 | taxonomy | F12 catalogued with four instances and their mechanisms |
| **2567** | **the fix** | **aarch64 Q4_K parallelism — 1.21× measured on GB10** |

## Performance: measured, and fixed

The hottest kernel in quantized inference ran **single-threaded on every ARM machine** — `matmul_q4k_f32_parallel` called the serial routine. Fixed with no new SIMD and no arithmetic change.

**gx10 (GB10, 20 cores), 8960×1536, release, 10 warmups: 2.17 ms → 1.79 ms, 1.21×.**

**The first number was 2.91× and was not shipped.** Two warmups gave a bimodal serial series whose median sat in the empty gap between modes. Why only 1.21× from 12 threads, with arithmetic not a guess: `thread::scope` spawns per call at ~40 µs — twelve spawns ≈ **27%** of the parallel time. Not bandwidth (4.3 GB/s ≪ GB10). A thread pool and a NEON kernel are the remaining headroom; #2567 stays open for the SIMD half.

Correctness verified **on real aarch64**: 3/3 green on gx10, including bit-exact chunk-invariance across 1/2/3/5/12/64-way splits.

## Three false positives that only exist on one tree

Each branch was green alone. These appeared only here — the argument for integrating before merging:

1. `check_no_fabricated_baselines.sh` **flagged itself** (its case table *contains* the forbidden construct) — and only once it became **tracked**, its universe being `git ls-files`. Third instance of that shape in this epic; universe now unions the working tree, re-mutated in the widened scope.
2. `LLAMA_PIN_RC=3` — a **return code** — read as a fabricated baseline where PARITY-009 met PARITY-005.
3. Row 1b flagged `bench_host_receipt.sh`, which installs nothing — it matched on a **comment**. Comments are not invocations.

Also restored: the deployed user-scope shim still carried **pre-#2640 text**, which is why `check_dogfood_shim` rows 3c/3d were red throughout. A stale file, not #2646's siting question — that remains open.

**All ten guards green on this head.**


noahgift:
Superseded by #2705.

Issue #2696, opened 2026-08-24
P0: published apr 0.64.0 SILENTLY IGNORES --gpu — 15.7 tok/s decode, 7.5s to first token, 0.099x llama.cpp


noahgift:
`cargo install aprender` produces a binary with **no CUDA path compiled in at
all**, and `apr serve run --gpu` accepts the flag, prints no warning, and runs
on CPU. A user who follows the documented invocation gets a tenth of the decode
rate and a 7.5-second wait for the first token, with nothing anywhere telling
them why.

## Measured, same client / same prompts / same clock / same GPU

| runtime | decode tok/s | prefill tok/s | TTFT p50 | vs llama.cpp (decode) |
|---|---|---|---|---|
| llama.cpp 39173bcac | 158.9 | 10399.4 | 9.8 ms | 1.000x |
| apr 0.64.0, local `--features cuda` | 103.3 | 2860.2 | 35.7 ms | 0.650x |
| **apr 0.64.0 PUBLISHED (`cargo install`)** | **15.7** | **13.5** | **7541 ms** | **0.099x** |

qwen2.5-coder-7b-instruct Q4_K_M, RTX 4090 sm_89, profile medium (128 in /
128 out), warmup 15s, 3 x 30s, cooldown 10s, concurrency 1, streaming —
the `[protocol.http]` lane in `scripts/llama_pin.toml`. Instrument:
`apr test llm bench` (#2692). Receipts under `evidence/parity-http/`.

The `cuda` feature is worth **6.6x on decode and 211x on prefill**. Against
llama.cpp the published binary is **769x behind on prefill**.

## The artifact, examined rather than assumed

```
$ ~/.cargo/bin/apr --version
apr 0.64.0 (v0.64.0+no-git)                 # published crate, no git metadata

$ strings -a ~/.cargo/bin/apr | grep -cE 'cudarc|libcuda\.so|libcublas'
0                                            # no CUDA linkage of any kind

$ strings -a ~/.cargo/bin/apr | grep -c 'Enabling optimized CUDA acceleration'
0                                            # the banner the cuda build prints

$ strings -a ~/.cargo/bin/apr | grep -c 'ThisStringIsNotPresent'
0                                            # control: the probe discriminates
```

And while serving with `--gpu`, `nvidia-smi --query-compute-apps` lists **no
apr process at all** — zero bytes of VRAM. The local `--features cuda` build in
the same test holds 14030 MiB.

## Why it is silent

[...]

PR #2698, opened 2026-08-25
fix(serve): --gpu on a build with no GPU backend FAILS (PERF-003, jidoka)

noahgift:
**APR-PERF-GATE-001 v2.2** §10.1 step 1 · §11 **PERF-003** (EV 1, no deps) · isolated PR off `main` per §10.2.

## The defect

`use_gpu` is read only inside `#[cfg(feature = "cuda")]` blocks (`handlers.rs:776`, `:1084`). Without the feature those blocks vanish, `config.gpu` is never consulted, and the server starts on CPU **having warned nobody**.

`cargo install aprender` produces exactly that build — published `aprender-0.64.0/Cargo.toml` has `default = ["cli"]` with `cuda` opt-in, read from static.crates.io rather than a local path. Measured 2026-08-25 with an **idle RTX 4090 in the machine**: 15.7 tok/s decode and 7.5 s to first token, against llama.cpp's 158.9 (#2696).

## The call site is the gate, not the function

v2.2 §5 names this row's mutation as *"remove `ensure_accelerator_available`"*, and the four behavioural tests would all have **survived** it — they call the function directly, so deleting the call in `run()` leaves them green while `--gpu` goes back to being ignored.

`the_guard_is_actually_wired_into_run` reads the source of `run()` and asserts the call is present. Crude, and it is the only test here that fails when the wiring is removed.

## Mutation record (§5)

Which **build** a mutation can fire on is part of the record. My first attempt ran the whole set on one build and reported a dead rule that is not dead:

| id | mutation | build | result |
|---|---|---|---|
| F-ACCEL-004 | remove the call site | default | **RED** — 4 passed, 1 failed |
| F-ACCEL-003 | drop the `!no_gpu` conjunct | default | **RED** — 3 passed, 2 failed |
| F-ACCEL-002 | drop the backend disjunct | default | **RED** — 3 passed, 2 failed |
| F-ACCEL-001 | bypass the capable-build early return | `--features cuda` | **RED** — 2 passed, 2 failed |

F-ACCEL-001 is a **no-op on a default build** — `cfg!(any(feature = "cuda", feature = "wgpu"))` is already false there, so replacing it with `false` changes nothing. It is only observable on a cuda build, against a 4/4 green baseline.

**Before-evidence:** on `main` there is no guard at all (`git grep ensure_accelerator_available origin/main` → nothing), and `--gpu` succeeds silently on a CPU-only build.

**Discrimination:** reverted → 5/5 green on default, 4/4 on cuda. A CPU run without `--gpu` succeeds under every mutated build too, so the guard is not simply rejecting everything.

## Contract, same PR

`contracts/accelerator-request-v1.yaml` — 4 falsification tests, 3 proof obligations, `pv validate` clean.

One Kani harness is marked `declared-not-written`, with its **cheaper alternative written rather than deferred**: the predicate reads four booleans and allocates nothing, so "for all inputs" is sixteen cases. `the_refusal_is_total_over_every_input` walks them and asserts `Err` lands in exactly one cell.

## The remedy in the message is real

#2527 shipped an error naming a rebuild that could not be performed. Both `cuda` and `wgpu` are declared at the facade, and this branch builds under `--features cuda`.

## Known limitation, recorded not discovered later

A **boolean flag has no observable resolution** (v2.2 finding N4). This makes an unhonourable request fail — necessary, **not sufficient**. It cannot express partial offload, because `--gpu` admits two values. Neither comparator has a boolean: llama.cpp takes `-ngl` as integer/`auto`/`all` and reports what it resolved. Superseded by **PERF-021**.

Scope is the **request**, not the **distribution**. That `cargo install aprender` yields a CPU-only binary at all is #2696's open half.

## Verification

7101/7101 apr-cli lib · clippy clean · fmt clean · `pv validate` valid

Refs #2696

### DIFF
--- a/contracts/accelerator-request-v1.yaml
+++ b/contracts/accelerator-request-v1.yaml
@@ -1,5 +1,5 @@
 metadata:
-  version: 1.0.0
+  version: 2.0.0
   created: '2026-08-25'
   author: PAIML Engineering
   description: >-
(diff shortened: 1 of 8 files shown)


noahgift:
Superseded by #2705.

PR #2666, opened 2026-08-24
Fix BENCH-004: add determinism controls to benchmark matrix harness

noahgift:
Resolves #2665
This PR adds the required statistical controls to `scripts/benchmark-matrix.sh` per AUDIT-PROTOCOL:
- Extracts median (p50) calculation.
- Implements discard logic for outliers on >5 samples.
- Adds `taskset` thread pinning inside the benchmark loop.
- Warns if the CPU governor isn't set to `performance`.

These determinism controls ensure valid and reproducible audit receipts.

PR #2705, opened 2026-08-27
feat(perf): land APR-PERF-GATE-001 v2.2 — and repair nine gates that could not fail

noahgift:
Lands APR-PERF-GATE-001 v2.2 — 71 commits that have been accumulating in a
CI-dark stack. Supersedes #2681, #2684, #2687, #2690, #2691, #2698.

## Why this is one PR

Those six PRs are a **stack**, each based on the previous rather than on `main`:
`#2682 base=feat/parity-004-no-timing-in-required`,
`#2692 base=feat/parity-010-fairness-protocol`,
`#2685 base=feat/parity-008-bootstrap-threshold`. Seven show MERGED while `main`
contains none of them. `ci.yml` fires only on PRs targeting `main`, so the whole
chain got `authorize` and nothing else — the documented CI-dark pattern. This is
the first time any of it faces a required check.

Merging the six individually was tried and is actively harmful: four conflict
across files this PR repairs, and the diffs read
`189 files changed, 248 insertions(+), 96,714 deletions(-)` — the batch being
*ahead*, not the branch. Merging would re-apply stale guards. Commit-subject
matching (squash suffixes stripped) confirms containment: 2/2, 16/16, 12/12,
8/8, 15/15, 19/19.

**#2666 is deliberately NOT superseded.** Its `BENCH-004` commit is genuinely
absent here and this branch deletes the file it edits. That is an open decision,
not an oversight.

## Nine repair commits — the epic was RED on five guards

`batch/perf-gate-v2-2` failed `check_guards_are_wired`,
`check_no_fabricated_baselines`, `check_apr_bin_pinned`, `check_readme_claims`
and `check_dogfood_coverage`. All five are GREEN on clean `main`, so the branch
caused every one. Worse, two guards the §5 registry marks *written* were **green
on their own named mutations** — a falsified guard is worse than a missing one.

| | Defect | Proof it is fixed |
|---|---|---|
| r1 | Fabricated-baseline guard comment-blind **and** keyword-blind — `readonly`/`export OLLAMA_BASELINE=291` matched nothing | 5/5 case table |
| r2 | PERF-009's harness guard detected **0 of 4** harnesses it exists to stop | 3/3 restored verbatim → RED; unrelated → GREEN |
| r3 | Claim-literal guard could not see `book/`; depth-5 glob hid **1,045** tracked `.rs` files | 5 live `2.93x Ollama` claims deleted; `[X]` guard added |
| r4 | Both merge-phase guards named by no workflow | Full unwire → RED |
| r5 | `apr` pinning — 3 offenders, 1 real | Probe sources `apr_bin.sh` |
| r6 | README promised a GPU backend the default build does not ship | `FALSIFY-README-005`, 4/4 |
| r7 | Surface ledger cited a line this epic moved | G2.1 green |
| r8 | `expires` declared on every cell, read by **zero** lines | Self-arms on a date, 3/3 |
| r9 | `F-BATCH-001` RED at c=2, contract said otherwise | Verdict stated |

### The two that matter most

**r3 — the claims were live.** `book/src/examples/showcase-benchmark.md:22` read
`851.8 tok/s = 2.93x Ollama (291 tok/s baseline)`. 291 is the exact hardcoded
default `check_no_fabricated_baselines.sh` exists to ban. §9 names these five for
deletion; all five were still shipping and the guard written to catch them
reported PASS. `grep -rn '2\.93' book/` is now empty.

**r8 — the deadline was decorative.** `expires: '2026-09-25'` on every cell,
`grep -c expires scripts/perf_gate.sh` → 0. All 8 cells would have sat in REPORT
forever and that date would have passed in silence. `TODAY` is injectable so
both sides of the boundary are proven now rather than in a month:

```
today=2026-08-27, expires 2026-09-25   rc=0  REPORT
today=2026-09-26, expires 2026-09-25   rc=1  FAIL EXPIRED   <- self-arms
today=2026-08-27, no expires field     rc=1  FAIL           <- omission is not a pass
```

## PERF-001 is REAL BUT PARTIAL — and this PR says so

Three independent sources agree, and the contract previously did not:

- `F-BATCH-001` postcondition is `serialization_index(c) < c for every c > 1`;
  the contract's own recorded AFTER is `1.00/2.45/2.29/2.39`. **2.45 >= 2.** It
  was written up as "N=2 regresses ~23%" — true, and not a verdict.
- `F-BATCH-003` measured RED on the post-fix tree: staggered arrivals still queue.
- Two serialized fallbacks survive at `cuda_chat_backend.rs:148/204`, under a
  comment that names itself: `// Fallback: direct RwLock path (serialized, no
  batch scheduler)`.

The aggregate win at N=8 is real (3.32x, ~335 → ~1100 tok/s). c=2 is not fixed.

**This makes releases fail, which is intended.** §4.1: "landing this gate makes
0.64.0 uncuttable until Arm A clears." No floor is lowered here.

## Verified before push, rc captured directly, never through a pipe

```
cargo fmt --all -- --check                       rc=0
cargo check --workspace --locked                 rc=0
cargo check -p aprender-serve --features cuda    rc=0
<ci.yml:358 chain, extracted verbatim>           rc=0   33/33 suites
all 79 ci.yml shell gates                        rc=0
scripts/perf_gate.sh --selftest                  rc=0   17 passed
pv validate contracts/batch-admission-v1.yaml    rc=0
```

The CUDA check is not inferred from `rc=0`: a `compile_error!` placed inside
`crates/aprender-serve/src/cuda/mod.rs` fired (`rc=101`) with the feature on and
vanished without it. ~950 CUDA-gated lines that no required check compiles do in
fact compile. First `rc=0` was nearly a false pass — `aprender-serve` was cached
and `trueno-gpu` is a **rename** of `aprender-gpu`, so grepping `cargo tree` for
it found zero.

## Two things I got wrong, recorded

- r6's guard first landed **after** the dispatch `esac`, where `all)` ends in
  `exit "$fail"` — unreachable. A gate that cannot fail, written inside a commit
  about gates that cannot fail. Caught only by mutating it.
- r7 wanted a ledger row for `apr test llm bench`. I wrote `quality 4` by copying
  its sibling; nothing measured it. That column feeds every floor in the gate, so
  hand-assigning it is the same act as hand-assigning a comparator baseline.
  Backed out. It also exposed a real gate defect: G2.3's floor derives from
  `origin/main`, so an honestly-recorded new ungated surface always reads as
  growth — a PR that adds a CLI surface structurally cannot record it.


noahgift:
Tracked under **Epic #2706** — APR-PERF-GATE-001.

Issue #2706, opened 2026-08-27
Epic: APR-PERF-GATE-001 — the receipt rule: a performance number is evidence only if something can prove how it was measured


noahgift:
# The receipt rule

> **A performance number is evidence only if something can prove how it was measured.**

**Spec:** [`docs/specifications/APR-PERF-GATE-001-v2.2.md`](https://github.com/paiml/aprender/blob/main/docs/specifications/APR-PERF-GATE-001-v2.2.md) · **Roadmap:** `docs/roadmaps/roadmap.yaml` → `APR-PERF-GATE-001` (+25 tasks, all `github_issue: 2706`) · **Query:** `pmat work list | grep -E 'PERF-|APR-PERF-GATE'`

**Operator ruling, carried:** *"this is the most important gate in the project. If this is not addressed the entire project shouldn't exist."*

## Nomenclature (§0.0)

| Register | Handle | Use for |
|---|---|---|
| Formal | `APR-PERF-GATE-001` | commits, contracts, roadmap ids |
| Artifact | **the perf gate** | `perf_gate.sh`, the arms, `perf-matrix.yaml` |
| Principle | **the receipt rule** | conversation, review, teaching it |

Not *"the gate"* — 72 `check_*.sh` guards, three CI contexts named `gate`, three `gates` in `[package.metadata.dogfood]`, plus the clean-room release gate. It picks out nothing here.

## Progress — 10 / 25 (40%)

Measured from git history and the tree, not estimated.

| | count | ids |
|---|---|---|
| completed | 10 | PERF-000, 001*, 002, 003, 004, 005, 009, 014, 015, BENCH-003 |
| in progress | 1 | PERF-021 (#2707) |
| planned | 14 | PERF-006, 007, 008, 010, 011, 012, 013, 016, 017, 018, 019, 020, 022, 023 |

\* PERF-001 is **REAL BUT PARTIAL** — see below.

**40% is deliberately lower than the 55% quoted earlier today.** That figure counted §11's 20 tickets; this counts 25, including **PERF-023** (partial GPU offload), which the tree proves is needed and the spec never listed.

## The two facts that matter more than the percentage

**1. The gate runs nowhere.**

```
git grep perf_gate -- .github Makefile       →  rc=1, 0 matches
[package.metadata.dogfood] gates             →  perf_gate.sh absent
check_guards_are_wired.sh:44                 →  for g in "$root"/scripts/check_*.sh
[...]

PR #2707, opened 2026-08-27
fix(cli): PERF-021 shipped #2696 again in the new spelling, and two of three surfaces verified nothing

### LINKED ISSUE #2696: P0: published apr 0.64.0 SILENTLY IGNORES --gpu — 15.7 tok/s decode, 7.5s to first token, 0.099x llama.cpp

noahgift:
Epic **#2706**. Closes nothing — see the last section, which matters.

## The finding: this branch shipped #2696 again, in the new spelling

PERF-021 exists because of v2.2 finding **N4** — *"a boolean accelerator flag has no observable resolution: `--gpu` can be ignored and nothing in the output changes."* The branch retired the boolean for a quantity. Then:

```
--features cuda:  apr serve run --gpu-layers all m.gguf
  → dispatch_run.rs   gpu = false, gpu_layers = Some(All)
  → ensure_accelerator_available: wants_layers → cfg!(cuda) → Ok()   PASSES
  → handler_gpu_completion.rs:412  `if config.gpu && !config.no_gpu`  FALSE
  → run_cpu_server
```

Four sites chose GPU vs CPU and **all four read the boolean**; `gpu_layers` was read only by the guard. The guard passing is what made it silent — it confirms the request is legitimate, then nothing acts on it. That is #2696's exact shape, reintroduced by the change that retires it.

**Its own test could not see it.** `gpu_layers_is_refused_on_a_build_with_no_accelerator` carries `#[cfg(not(any(feature = "cuda", feature = "wgpu")))]` — it **compiles only where the bug cannot happen**. And `ci.yml` never builds `--features cuda` at all; only `cuda-nightly.yml` does.

## Two of three surfaces verified nothing

#2696's 15.7 tok/s was measured through **`apr run`**, which had no guard. Neither did `apr chat`.

| surface | before |
|---|---|
| `apr serve` | guarded |
| `apr run` | `--backend cuda` checked, `--gpu` **not** — measured here |
| `apr chat` | nothing |

`apr run` is the sharper case: it already refused `--backend cuda` with a long message about not silently falling back to a path "~20x slower". Two spellings of one request; only the less-used one was guarded.

The refusal lived in `commands::serve` and took a `&ServerConfig`, so only `serve` **could** call it. Extracted to `crate::accel` taking plain values.

## The resolver was dead code

`resolve_gpu_layers` had 7 call sites, every one inside `#[cfg(test)]`, under a doc comment reading *"a request has a RESOLUTION, and it is reported."* It reported nothing.

Why it landed dead is the reusable part: its sibling `ensure_accelerator_available` shipped **with** a source-grep gate asserting its call literal appears in `run()`. That one landed wired. This one had none. So the gate ships with the call here.

Now printed on **both** paths, including `resolved=0` — a line that appears only on success is silent about the failure it exists to expose:

```
gpu-layers: requested=all resolved=29 total=29 (backend=cuda)
```

## `fits = total_layers` is a limitation, not a measurement

Documented at the call site rather than left to infer. No free-VRAM query is reachable from any apr build: `query_cuda_memory` sits behind aprender-compute's `cuda-monitor` feature, which apr-cli's `cuda` feature does not enable, and `CudaContext::memory_info()` is only reachable **after** the model is constructed — i.e. after this decision.

So a **partial request is refused, not rounded**. `OwnedQuantizedModelCuda` takes no layer count and uploads every layer, so accepting `--gpu-layers 12` on a 29-layer model would place 29 and print 12 — a fabricated number in a log, which is worse than a refusal and is the exact construct this epic removes. Tracked as **PERF-023**. `all`, `0` and `Exact(total)` are still accepted, or the refusal would be a wall rather than a gate.

## Gates, each with its mutation observed RED

| gate | mutation | result |
|---|---|---|
| `no_decision_site_reads_the_bare_accelerator_boolean` | revert one site to the boolean | FAILED, names `handler_gpu_completion.rs:412` |
| `the_quantity_alone_selects_the_accelerator` | make `wants_accelerator()` read the boolean | FAILED |
| `every_accelerator_surface_calls_the_refusal` | unguard `apr run` / `apr chat` | FAILED, names each surface |
| `the_resolver_is_actually_called_from_a_decision_path` | remove the call / compute-but-don't-report | FAILED (both) |
| `a_partial_offload_is_refused_because_the_loader_cannot_do_it` | accept instead of refuse | FAILED |

None carries a `cfg`, so all run on the CUDA build too — unlike the test this branch shipped with.

## Verified, rc captured directly

```
cargo fmt --all -- --check                    rc=0
cargo check -p apr-cli                        rc=0
cargo check -p apr-cli --features cuda        rc=0
cargo clippy -p apr-cli --lib -- -D warnings  rc=0
cargo test -p apr-cli --lib                   rc=0   7123 passed, 0 failed
```

`clippy --all-targets` reports 254 pre-existing disallowed-`unwrap` errors, all in `crates/apr-cli/tests/falsification_crux_*.rs`, **none in any file this branch touches**. Recorded rather than silently scoped around — "I ran a narrower command" is how a red gate becomes invisible.

## This does NOT close #2696

§4.9.3: *"Rules 1-4 change nothing if the artifact carries no backend."*

Every surface is now loud. None of that ships CUDA in the published artifact — that is the O-1 distribution decision (**PERF-022**), and no code change substitutes for it.


noahgift:
Your queue run failed on a **real test failure**, not the SIGSEGV flake the step name suggests:

```
test_zscore_produces_zero_mean
  zscore mean = 0.00010859966, expected ~ 0
  minimal failing input: a = [75.395966, 75.466225]   (114 successes first)
```

Nothing to do with this PR — `test_zscore_produces_zero_mean` asserted a fixed `1e-4` bound on a quantity whose achievable f32 precision is `eps*max|a|/sigma`, which for that input is **2.56e-4**. Proptest simply drew an ill-conditioned pair for the first time. Fixed separately; this PR should re-queue clean once that lands.

PR #2711, opened 2026-08-28
fix(gate): the claim-literal ratchet could be edited in place — plus PERF-006 andon and the harness write-before-validate

noahgift:
Epic #2706 (the receipt rule). Batch B — three sub-tickets from four concurrent worktrees; two were held back after adversarial verification, see the end.

## The claim-literal ratchet had a HOLE, not just a false red

I verified this on `main` before trusting it. Changed a shipped doc comment from `200+ tok/s` to `900+ tok/s` — a 4.5× fabricated performance claim on a user-facing surface — and:

```
bash scripts/check_no_claim_literals.sh   →  rc=0,  "known: 138  new: 0"
```

Because the baseline keyed on `FILE:LINE`, **any baselined claim could be edited to say anything and the ratchet blessed it.** The claim-literal guard is the receipt rule's primary enforcement, and it could be walked straight through by editing in place.

The reported symptom was only the false red (PERF-006 inserting `pub mod andon;` moved two baselined claims and they read as NEW). The hole was underneath it.

**Now keyed on `file + sha256 of the normalised claim text` + an occurrence count.** The count is load-bearing: pure content-keying lets a second copy of an existing claim ride in free on the first.

| mutation | before | after |
|---|---|---|
| edit a baselined claim's number in place | rc=0 **PASS** | **rc=1** |
| insert 10 lines above a baselined claim | rc=1 (false red) | **rc=0** |
| second copy of an existing claim | — | **rc=1** (count exceeded) |
| delete a baselined claim | — | rc=0 + REPORT to prune |

Editing the text to dodge makes the guard *louder*, not quieter: the old key goes unmatched **and** a new key arrives.

## PERF-006 — one `compute_class()`, one `max_in_flight()`, three surfaces

§14 calls `max_in_flight` *"the cheapest confirmation of defect #2"*. Banner, `/health`, and receipt now call one function, so they cannot disagree. Emitted on the **serialized** path too — a field that appears only when batching is active reports success and stays silent on the failure it exists to expose.

Gate: a source scan that all three surfaces delegate. Three mutations, each letting one surface compute its own → rc=101, only that surface's test failing.

## `apr test llm bench` wrote its receipt before checking the run was valid

`test_llm.rs:121` wrote `--output` and only then asserted validity at `:135` (`failed > 0`) and `:160` (zero-token runs). Both `return Err` — with the JSON already on disk. **A failed benchmark left a receipt-shaped artifact.**

## The merge needed care

#2710 landed mid-flight, so PERF-016's matcher widening collided with the re-keying in one file. Both add *different* selftest tables (causal-class vs key-mechanics); both kept, 32 cases pass.

The baseline was the risk. Regenerating with `--list` over the current tree would be a **blanket amnesty** — the drift fix's own docs say so. Reconciled and checked at occurrence level instead:

```
old (FILE:LINE keys): 137 occurrences
new (content keys):   137 occurrences
files differing:        0
```

137 → 136 *entries* is one collapse of an identical claim appearing twice, not a claim quietly dropped.

## Verification

```
cargo fmt --all -- --check        rc=0
cargo check --workspace --locked  rc=0
91 ci.yml shell gates             all rc=0
```

G2.1 freshness needed a ledger touch: `GET /health` genuinely changed, so its row now also cites `andon.rs:160`. All 40 rows citing `router.rs` were re-audited first — every cited line byte-identical, nothing drifted. An assertion caught that **two** binaries have a `GET /health` row; selecting on feature alone would have edited `aprender-orchestrate`'s, which this PR does not touch.

## Held back — refuted on verification, not substance

- **PERF-004-SCHEMA**: claimed an end-to-end run that could not have executed as written — the corpus lacks the `lanes.txt` the script requires; run directly it is rc=1, so the downstream step had no input.
- **PERF-019**: its guard has a vacuous pass one level below where it believed it closed one. The sentinel proves the *filter* engaged; nothing proves the *walker* did, and `2>/dev/null` suppresses the evidence.

Both fixes look right. Their proofs did not hold, so they wait.


noahgift:
## Scope note: this PR closes the in-place-edit hole, NOT the growth hole

I ran a sweep after landing PERF-008 (#2715) and found this guard still launders a fresh fabrication on **this branch**. Verified against `9d7d0463c`:

```
# a brand-new fabricated claim on a user-facing surface
book/src/zz_launder.md:4  'apr sustains 900+ tok/s on this workload.'
bash scripts/check_no_claim_literals.sh            -> rc=1   new: 1     (correct)

# now append the guard's OWN emitted key for it
bash scripts/check_no_claim_literals.sh --list | grep zz_launder
  book/src/zz_launder.md<TAB>b952dd21050d70c8<TAB>1
>> scripts/claim_literal_baseline.txt

bash scripts/check_no_claim_literals.sh            -> rc=0
  known (baselined, must shrink): 138   new: 0
  PASS  no NEW claim literal on a user-facing surface.
```

Note the guard prints **"must shrink: 138"** having just grown from 137.

This is **not a regression** — the same hole exists on `main` today (I reproduced it there with the FILE:LINE key: rc=1 without the entry, rc=0 with it). This PR is a strict improvement: it closes editing a baselined claim in place, which main allows. But the PR body should not be read as making the ratchet sound, and I would rather say so here than let the next reader infer it.

**Deliberately not fixing it in this PR**, since #2711 is in the merge queue and the hole predates it. The fix pattern already exists in #2715: compare the baseline against `merge-base(HEAD, origin/main)` with an `origin/main` tip fallback, and refuse growth as a **subset** property rather than a count — counting passes a swap, which is an append wearing the old total.

Filing the class as a follow-up: **12 of 12 baseline files in `scripts/` accept growth silently.**

noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:SPEC-1.6 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `crates/apr-cli/src/commands/serve/handler_gpu_completion.rs crates/aprender-serve/src/api/router.rs`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Maps to SPEC-1.6 (2026-09-12) through scripts/check_no_claim_literals.sh, and to R-0/P-0.6/G-8 through serve/mod.rs and check_baseline_ratchets.sh; it must land before those worktrees open. Semantic conflict in the GPU completion handler and router. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2711` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2711@ce6e7f104 · intent → #2989 (carry-0.67; no DAG row holds #2710) · salvage: crates/apr-cli/src/commands/test_llm.rs crates/aprender-serve/src/api/tests/{chat_completion_03,completion_request_response,deep_apicov_02,format_chat_02,imp_140c}.rs crates/aprender-serve/tests/{api_coverage,api_deep_coverage,property_api}.rs · conflicts: crates/apr-cli/src/commands/serve/handler_gpu_completion.rs crates/aprender-serve/src/api/router.rs`

PR #2712, opened 2026-08-28
fix(gate): a runner fault reported itself as a facade regression and blocked every PR

noahgift:
check_facade_compat.sh:205 decided the COMPATIBILITY verdict from a `cargo
check` whose output went to /dev/null, and named exactly one cause for any
non-zero exit: "0.3.1 consumer code no longer compiles against the facades".

cargo exits non-zero for two categories of reason and only one is about the
code. On 2026-08-27 rustc was un-spawnable on a clean-room runner --

    error: could not compile `equivalent` (lib)
    Caused by: could not execute process `.../bin/rustc` (never executed)
    Caused by: No such file or directory (os error 2)

-- while compiling `equivalent`, a crate that cannot fail to compile. The gate
reported a facade regression. `gate` hard-requires guard-runner-labels, so a
runner fault presented as a code defect and blocked #2707 and #2711 behind a
phantom nobody could reproduce.

The block also ran cargo TWICE: once for display, once -- output discarded --
for the verdict. The text shown to the reader was produced by a different run,
under a different machine state, than the one that decided. The `cargo test`
block six lines below already carries the rule, from #2574: "One invocation
only: re-running cargo to 'get the output' would re-run the tests under a
different machine state than the one that decided the verdict." That fix
stopped at its own block and left this residual -- the shape an audit keeps
finding, where the earlier fix was real and incomplete.

Now: one invocation, output captured, `rc` read directly and never through a
pipe, and the failure classified before it is named. ENV still exits 1 -- a
gate that goes green on "we could not tell" is the defect class this repo
names most often. The claim just has to be true.

Rows 1-8 of --self-test probe the python structural checker only, so the
classifier is a new surface and gets its own table rather than inheriting
their green; extending a guard's scope requires re-mutating in the new scope.

  mutation                                      selftest  row that reddened
  drop `could not execute process` from regex   rc=1      C1 (ENV arm)
  add bare `No such file or directory`          rc=1      C7 (discrimination)
  classifier always returns ENV                 rc=1      C5 (CODE arm)
  classifier always returns CODE                rc=1      C1
  reverted                                      rc=0      --

C7 is why the pattern anchors on cargo's spawn framing and not the bare
phrase: rustc emits "No such file or directory (os error 2)" for a source file
the repo really is missing, and that is a defect the gate must still report.

--self-test 15/15. Real run: COMPATIBILITY ok on the facade workspace.
bashrs 0 errors, 46 warnings -- byte-identical to the pre-patch count.

## Impact

This is currently blocking **#2707** and **#2711**. `gate` hard-requires `guard-runner-labels`, and that job failed with a message naming a facade regression that does not exist.

Refs #2706

### DIFF
--- a/scripts/check_facade_compat.sh
+++ b/scripts/check_facade_compat.sh
@@ -106,6 +106,35 @@ FACADE_WS="${REPO_ROOT}/crates/facades"
 # "not the aprender target dir"; only cache locality differs, so the fallback
 # is safe rather than silent.
 # ---------------------------------------------------------------------------
+# ---------------------------------------------------------------------------
+# cargo exits non-zero for two categories of reason, and only one of them is
+# about the code under test. aprender#2574 taught this for the `cargo test`
+# block below -- "the reader could not tell a signature drift from an OOM-killed
+# rustc" -- but the fix stopped at that one block and left this residual, which
+# is the shape a published-artifact audit keeps finding: the earlier fix was
+# real and incomplete.
+#
+# On 2026-08-27 the residual came due. rustc was un-spawnable on a clean-room
+# runner (`could not execute process ... (os error 2)`) while compiling
+# `equivalent` -- a crate that cannot fail to compile -- and this gate reported
+# `0.3.1 consumer code no longer compiles against the facades`. `gate`
+# hard-requires guard-runner-labels, so a runner fault presented as a facade
+# regression and blocked every open PR behind a phantom.
+#
+# ENV still exits 1. A gate that goes green on "we could not tell" is the
+# defect class this repo names most often. The claim just has to be true.
+classify_cargo_failure() {  # $1=logfile -> ENV | CODE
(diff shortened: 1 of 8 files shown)


noahgift:
**Controlled confirmation that the verdict this PR fixes was false.**

#2707 was re-run at the *same SHA* `cb8fd9011` with **zero code changes**. `guard-runner-labels` went **FAILURE → pass** (11m24s), and `gate` followed it green.

Nothing about the facades changed between the two runs. The run that reported

```
FAIL  0.3.1 consumer code no longer compiles against the facades
```

and the run that reported success were judging byte-identical code. The difference was the runner: rustc was un-spawnable in the first (`could not execute process ... (os error 2)`, while compiling `equivalent`, a crate that cannot fail to compile).

So the gate's verdict was not merely unhelpful — it was **wrong**, and it blocked #2707 and #2711 behind a defect that did not exist. That is the misattribution this PR removes, now demonstrated by a controlled experiment rather than inferred from a log.

PR #2713, opened 2026-08-28
fix(gate): the one-client gate passed with its walker dead — six vacuous passes, rc=0 each

noahgift:
Epic #2706 (the receipt rule). PERF-019 was **held back on a refuted proof** in the last batch, not on substance. This is the rework, and the rework found more than the review did.

## The one-client gate passed with its walker dead

The held commit proved its *filter* engaged and treated that as proof the whole guard ran. Five mutations of the **walker alone** against `48e334822` — each leaving `bash -n` at rc=0, so every result is a real verdict and not a parse error:

| mutation | rc on the held commit | rc now |
|---|---|---|
| `AWK_WALKER` → nonexistent path | **0 — vacuous pass** | 1 |
| awk program emptied (awk still exits 0) | **0 — vacuous pass** | 1 |
| `isbench()` can never fire | **0 — vacuous pass** | 1 |
| hit computed, then discarded | **0 — vacuous pass** | 1 |
| `walk_shell` bypassed entirely | **0 — vacuous pass** | 1 |
| `walk_literal` bypassed (sed+grep path) | **0 — vacuous pass** | 1 |
| pre-filter broken (*the old proof*) | 1 | 1 |

All six vacuous passes printed output **byte-identical to a genuine clean sweep**:

```
=== one client, both servers (PERF-019 / I-15) ===
  files=10469  invocations=0
OK
```

The second row is why an exit-status check could never have closed this: an empty awk program exits 0 and prints nothing. `2>/dev/null` on the walk then suppressed the only evidence that would have distinguished the two cases.

Honest qualifier: the `--selftest` step *did* red on these. The **gate** — the step whose rc is the verdict on the tree — did not. A guard that is sound only because a different step exists is precisely "one level below where it claimed closure."

## Second hole, found by mutating instead of reading

Dropping a real `Command::new("llama-bench")` into `crates/aprender-core/src/` left the gate at **rc=0**. `git add -N` on the same file took it to **rc=1** — same walker, same predicate, same bytes, only the universe differed. The untracked union covered `scripts/` and `crates/*/scripts` but not the two types the globs claim *repo-wide*: `*.rs` and makefiles.

Closed with a pruned working-tree find, and the **entire table above was re-run in the widened scope** — the old proof does not transfer when a guard's scope changes, so re-mutation happens in this commit rather than the next one.

## What now proves the walker engaged

1. **A sentinel only the walker can produce.** Each walker runs over a fixture of known classification and its output is compared against an expected string carrying *the line number the walker itself computed*. Each fixture names the same token in prose on line 1, so a walker that flags everything fails as surely as one that flags nothing.
2. **No evidence discarded.** `2>/dev/null` and `|| true` are gone. Every walk records exit status and stderr; either one bad **withholds the verdict** rather than reporting a clean file. `grep` rc=1 stays "no match", rc≥2 is an error, and stderr is checked at every rc including 0.
3. **Arithmetic on coverage, not on the universe.** Each completed walk appends to a ledger no other path writes, and the gate asserts one line per candidate. The counters live in *files* because every walk runs in a `$( )` subshell, where a variable counter would be discarded with the subshell.

Output now names what engaged, rather than only what it concluded:

```
walker: ENGAGED ... / filter: ENGAGED ... / files=10469 candidates=14 walked=14 invocations=0
```

## The substantive claim, verified by source scan

`scripts/parity_host_receipt.sh:101` and `:117` both drive `"$APR" test llm bench` — against apr's port and against `"$LLAMA_SERVER"` (`:108`). One client, both servers. Tree-wide `invocations=0` across 10,469 files with both walker and filter proved engaged. `llama-bench` survives only as ban-list entries, a build-tree anchor, and fixtures — all green by construction, zero allowlist entries.

## Verification

`bash -n` 0 · gate 0 · `--selftest` 0 (48/48) · bad-arg 2 · `check_guards_are_wired` 0 · `check_no_claim_literals` 0 · `check_no_timing_in_required` 0 · `check_pass_grep_anchored` 0 · `check_shell_lint_ratchet` 0 · `check_llama_pin` 0 · `check_no_competing_harnesses` 0 · `check_no_fabricated_baselines` 0 · `check_sourced_libs_option_neutral` 0 · `check_no_hand_rolled_parsers` 0. Gate cost 0.38s.

## Bonus defect, filed not fixed

`http_client/preflight.rs` and `http_client/benchmark_runner.rs` are **byte-identical 370-line files** (`diff -q` rc=0); only `benchmark_runner.rs` is `include!`d. The dead duplicate is a live trap — it reads exactly like the production benchmark client and is wired only to benches and tests. It cost me a wrong finding on #2706 earlier today, which I have since retracted there. Worth its own ticket.

Refs #2706

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -990,6 +990,22 @@ jobs:
       # (#2676, closing the open half of #2647).
       - name: llama.cpp comparator pin discriminates
         run: bash scripts/check_llama_pin.sh
+      # PERF-019 / I-15. The pin above says WHICH llama.cpp; this says WHICH
+      # BINARY of it may be driven. 4.4.8/X7: llama-bench does not separate PP
+      # from TG under concurrent load, so a ratio built from it compares two
+      # CLIENTS with two servers attached rather than two servers. The rule is
+      # one client -- `apr test llm bench` -- against llama-server on both
+      # sides, which is what scripts/parity_host_receipt.sh already does.
+      #
+      # The tree was already clean when this landed; the guard is what keeps it
+      # clean. Neither existing guard covered it: check_no_competing_harnesses
+      # requires a file to ALSO compute a rate, and parity_host_receipt.sh
+      # computes none, so a llama-bench call added there stayed green (rc=0)
+      # with the defect in place.
+      - name: Comparator is not driven by llama-bench (I-15)
+        run: bash scripts/check_comparator_one_client.sh
+      - name: One-client guard can still go RED (case table)
+        run: bash scripts/check_comparator_one_client.sh --selftest
       # An unstated knob is a silent degree of freedom: it becomes whatever
       # each side defaults to, and the difference lands in the ratio wearing
       # the label of a performance result (#2677).
--- a/docs/benchmarking-gate-spec.md
+++ b/docs/benchmarking-gate-spec.md
@@ -162,6 +162,34 @@ Host list is `HOSTS="lambda intel gx10 mini"`
 **intel needs a dedicated single-agent label.** `intel-clean-room-{1..16}` is 16
 org-scoped agents on one box; a perf run must not share it.
 
+#### 3.3.1 One client, both servers (I-15)
+
+The comparator is **`llama-server` driven by our client**, never `llama-bench`.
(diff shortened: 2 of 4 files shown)


noahgift:

PR #2714, opened 2026-08-28
fix(gate): the ratio guard caught 1 of 6 shapes — the headline fabrication was invisible in its own preferred spelling

noahgift:
Epic #2706 (the receipt rule). PERF-010 part (a) had already landed in #2710 — this is the residual, and the residual was the part that mattered.

## The `[X]`-figure guard was listed as "not written" and the spec was right

Spec §5 line 779 marks it unwritten. Reading `RATIO_RE` it looks written. The reason it wasn't: the pattern required the competitor's name **immediately** after the ratio, in **ASCII `x`**. Of the six shapes §0.1 itself enumerates, it caught one.

I verified this independently on `origin/main` (`de8fbc407`), file tracked via `git add -N`:

| shape | rc on main |
|---|---|
| `36.9x FasterTransformer` | **1** — caught |
| `36.9x over FasterTransformer` | 0 — invisible |
| `36.9× over FasterTransformer` ← §5's own mutation | 0 — invisible |
| `23× over static batching` | 0 — invisible |
| `1.8× over vLLM` | 0 — invisible |
| `2.93× Ollama` (U+00D7) | 0 — invisible |

The U+00D7 row is the one that matters. This repo's prose prefers the typographic sign, so **the epic's most-quoted fabrication was invisible in its preferred spelling while its ASCII twin was caught.** A guard that greps for the rarer of two spellings of the same claim is not a partial guard; on the corpus it actually faces, it is closer to none.

## `[X]` means three different things and the ticket picked the wrong one

Worth recording, because the obvious implementation would have been born red. Spec §0.1:114 defines `[X]` as a **provenance tag** — "External claim about a third-party system" (36.9× FasterTransformer, 23× static batching, 1.8× vLLM). But the literal `[X]` in this repo is overwhelmingly a **markdown checkbox** — roughly ten `- [X] APPROVED for Production` in `docs/qa/`.

A bare ban on the literal, as the ticket described it, would have fired on ten checked boxes and zero defects. Implemented the spec's meaning as the primary guard; the placeholder class (`XX tok/s`, `[TBD]×`) is bound to a **perf unit**, which is what makes it discriminate.

## Mutation table — 22 rows, every rc read as `guard > log 2>&1; rc=$?`

| mutation | rc |
|---|---|
| clean tree, both guards | 0, 0 |
| uncited speed comparison, tracked | 1 |
| same claim + **resolving** `evidence/` citation | 0 |
| same claim + **dangling** citation | 1 |
| a TARGET, not a result | 0 |
| uncited claim in an **untracked** file | 1 |
| the five previously-invisible ratio shapes | 1 ×5 |
| `2.9x faster than Ollama` ← the live `book/` defect | 1 |
| `XX tok/s` · `[TBD]×` · `[X]x faster` | 1, 1, 1 |
| markdown checkboxes | **0** |
| `3x3 convolution` on a line naming Ollama | **0** |
| `3x larger than PyTorch` (size, not speed) | **0** |
| revert | 0, 0 |

The last three are the discrimination set — a guard that reds on those is as useless as one that never reds.

24 new markdown rows in `check_no_claim_literals.sh`'s case table. The existing 12 all required a Rust print macro, so **not one could reach the `.md` detector that guards `book/`** — which is how "covered by check_no_claim_literals.sh" survived in the spec's status table while `book/` went unguarded.

## (c) Comparator claims deleted from `book/`

`book/src/tools/apr-cli.md`, the only live comparator claims in `book/`:

- **:81** `# Batched GPU mode (2.9x faster than Ollama)` — **refuted**: the harness never ran Ollama (§5.2), and the decode ratio was separately withdrawn (measured 1.015–1.109× on sm_89, i.e. parity).
- **:87-93** the whole `vs Ollama` table — the `2.9x`/`0.2x`/`0.05x` column and an `Ollama | ~333 tok/s` row. Refuted *and* unsourced; the `~850 tok/s` came from the same run. Replaced with `apr bench --fast --json`.

The first withdrawal note *restated* the figures and reddened the guard at its own new line — the self-reference trap. Reworded: **a withdrawal that reprints the number leaves the number on the page.**

**Nothing concedes speed.** Every grounded win is untouched (cold-start, LinReg, GaussianNB, SIMD, footprint), and the real `--assert-speedup`/`--skip-ollama` flag docs stay — those are flags, not claims. This declines to publish a ratio nobody measured; under-claiming would be the same reporting failure in the other direction.

## Residual holes — stated, not hidden

1. **The book edit was never validated against a live `apr`.** `check_book_cli_parity.sh` and `check_book_examples_executable.sh` fail on this branch **and byte-identically at HEAD** (`STALE apr BINARY` — no binary built from `de8fbc407`). They fail closed, correctly. Pre-existing and environmental, but this PR does not claim the page validates.
2. **`13× on 200k-token prompts` is still uncaught** — a §0.1 figure with no comparator name. Catching it generically means flagging every ratio.
3. **`docs/specifications/` remains excluded** from both guards (pre-existing). A claim laundered there is invisible.
4. **`book/src/examples/apr-format-deep-dive.md:296-304`** — a load-time table (PyTorch 2.3s / SafeTensors 450ms / GGUF 380ms / .apr 0.8ms). Comparator and unsourced, but times without ratio syntax, so neither guard fires. The implied 562–2875× may be the grounded cold-start win (526–1600×) — **could not tell, so left in place and reported** rather than guessed at in either direction.
5. **12 sites baselined as inherited debt** (shrink-only, argued in the file header). Seven are admissions of being *slower*. The one unverified **overclaim** — `docs/qa/five-whys-gpu-performance-gap-2026-01-08.md:317`, "32B: Now 3.1x faster than Ollama" — is flagged **by name** in the baseline header rather than quietly ratcheted: it is a 32B claim and the withdrawal measured 1.5B, so deleting it would be guessing in the other direction. It needs a receipt or a deletion.

## Verification

`check_guards_are_wired` · `check_shell_lint_ratchet` · `check_pass_grep_anchored` · `check_no_fabricated_baselines` all rc=0. bashrs errors on both edited scripts back to **0**, matching HEAD — the first draft introduced 5 false positives from markdown checkboxes parsed as shell test brackets. The `ci.yml` edit is comment-only and contains no `${{`.

Refs #2706

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -935,6 +935,15 @@ jobs:
       # named by no workflow enforces nothing, and recording that in a baseline
       # file would make the claim true only in the sense that the gap is
       # written down.
+      #
+      # This step is ALSO the I-12 `[X]`-figure gate, which v2.2 §5 listed as
+      # "not written" against this file. It was right: the comparator pattern
+      # required the competitor's name to sit immediately after the ratio, in
+      # ASCII, so of the six shapes §0.1 names it caught one. PERF-010 admits
+      # the connector ("36.9x over FasterTransformer") and U+00D7, and adds
+      # placeholders bound to a unit ("XX tok/s"). Twelve pre-existing sites
+      # surfaced and are baselined shrink-only; one was a LIVE comparator claim
+      # in book/ and is deleted.
       - name: No claim literals on user-facing surfaces
         run: bash scripts/check_no_claim_literals.sh
       - name: No claim-literal guard regression (case table)
--- a/book/src/tools/apr-cli.md
+++ b/book/src/tools/apr-cli.md
@@ -78,19 +78,37 @@ apr serve run model.gguf --port 8080
 # GPU-accelerated server
 apr serve run model.gguf --port 8080 --gpu
 
-# Batched GPU mode (2.9x faster than Ollama)
+# Batched GPU mode
 apr serve run model.gguf --port 8080 --gpu --batch
 ```
 
 ### Performance
 
-| Mode | Model | Throughput | vs Ollama | Memory |
-|------|-------|------------|-----------|--------|
-| GPU (APR Q4K, GH-88) | Qwen 1.5B | **240 tok/s** | — | 1.5 GB |
-| GPU (batched M=16) | Qwen 1.5B | ~850 tok/s | 2.9x | 1.9 GB |
-| GPU (single GGUF) | Qwen 7B | ~68 tok/s | 0.2x | 5.5 GB |
-| CPU (baseline) | Qwen 1.5B | ~18 tok/s | 0.05x | 1.1 GB |
-| Ollama | Qwen 1.5B | ~333 tok/s | 1.0x | - |
+This page publishes no throughput figure, because none of the figures it used
+to publish could name the run that produced them.
+
+Measure your own, on your own hardware:
+
+```bash
+apr bench model.gguf --fast --json > bench.json
+```
+
+Throughput depends on the model, the quantization, the backend and the
+concurrency, so a single number copied from someone else's box tells you very
+little about yours. `apr bench --json` emits the measurement together with the
+machine, the backend and the build it came from — which is the part that makes
+it worth quoting.
+
+> **Withdrawn.** A table here previously published a batched-GPU throughput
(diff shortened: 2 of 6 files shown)


noahgift:

PR #2715, opened 2026-08-28
fix(gate): the ledger printed "the count is enforced" and enforced nothing

noahgift:
Epic #2706 (the receipt rule). Two commits. The second exists because I tested the first one's fix and it did not close the hole.

## The ledger said SHRINK-ONLY four times and never compared against anything

`scripts/fabricated_baseline_rust_sites.txt` is the shrink-only ratchet for asserted competitor baselines in Rust. The guard's own prose:

> the ledger commits the true count TODAY **so the number cannot grow**
> The Rust ledger is SHRINK-ONLY: … **is not an allowlist to append to.**
> (shrink-only, PERF-008-RUST; **the count is enforced, not asserted**)

Reproduced on `origin/main`:

```
echo "crates/aprender-core/src/zz_grow.rs:2" >> scripts/fabricated_baseline_rust_sites.txt
printf '// p\nlet ollama_baseline_tps = 407.0;\n' > crates/aprender-core/src/zz_grow.rs
bash scripts/check_no_fabricated_baselines.sh
  -> rc=0
  ok    rust  37 ledgered site(s) = 37 ledger entr(ies), 0 new, 0 stale
               (shrink-only, PERF-008-RUST; the count is enforced, not asserted)
```

A brand-new fabricated baseline entered the tree and the guard printed PASS **while printing that last line verbatim**. Append a coordinate to the ledger, land a fabrication at it, and the ratchet blesses it.

The diagnosis is the useful part: `new` and `stale` are the two properties checkable from the **working tree**, and neither one is a ratchet. A ledger line and its fabrication added in the same commit satisfy both at once. **A ratchet is a property of the diff against a ref the author cannot rewrite, and this file had never looked at one.** SHRINK-ONLY appeared four times; the comparison appeared zero times.

Enforced now as a **subset property, not a count** — counting passes a swap, which is an append wearing the old total.

| row | rc | which check fires |
|---|---|---|
| unchanged | 0 | — |
| stale entry only (append, no site) | 1 | stale *and* grew |
| **append entry + create the fabrication** | **1** | **grew only** ← the defect |
| swap one coordinate, count unchanged at 36 | 1 | new, stale *and* grew |
| derive the literal + drop its entry (legit shrink) | 0 | — |
| drop the entry while its literal stays | 1 | new |

**Only the third row distinguishes this commit from its parent.** The other reds were already red for other reasons, and claiming the ratchet on their evidence would be the same over-attribution this PR is fixing.

Comparand: `merge-base(HEAD, origin/main)` preferred, **origin/main tip as fallback**, hard fail if neither resolves — never self-comparison. The fallback is not decoration: **CI checks out shallow, so the CI path *is* the tip path.** It is strictly stronger (also forbids re-adding an entry main deleted); the cost is a false red on a branch behind a main that already shrank the ledger, with the remedy in the FAIL text. Re-mutated in the tip scope against a real shallow clone using CI's exact fetch command — the merge-base proof does not transfer to it. No `ci.yml` change was needed: `guard-runner-labels` already fetches `origin/main` at `:737`, 190 lines before this guard runs at `:927`.

## First commit — the Rust detector could be blinded into a PASS

Found by mutation, not review: deleting one line of denylist logic in `rust_hits` yields `ok rust 1 ledgered site(s), 0 new` and **rc=0** — 35 of 36 known fabrications invisible, tree certified clean.

The case table stayed green through it because both Rust loops **re-implemented the rule inline**: the table graded a *copy*, and the copy stayed strict while the shipped scan went blind. One decision function, `rust_is_fab`, now serves both.

Same commit also confirms the SIGPIPE class with a measured curve rather than an argument. 20 runs per size, counting how often a real fabrication is silently lost:

| upstream bytes | 3584 | 9284 | 10234 | 11184 | 13084 | 165786 |
|---|---|---|---|---|---|---|
| lost (of 20) | 0 | 0 | **2** | **19** | 20 | 20 |

**Onset is ~10 KB, not the 64 KB the file's comment reasoned from** — the producer is itself a two-stage pipe, so what must fill is the first stage's buffer *plus* one pipe. Between 9 KB and 11 KB it is a race: same input, same binary, 2/20 then 19/20. There is no crisp threshold, so the comment now carries the curve and the fixture's 64 KB floor is labelled as a ~16× margin rather than presented as the boundary.

Also corrected the guard's own stale prose: it claimed `29 <- the true count` and "the 31 sites" while the ledger held **36**. Both numbers are now derived at run time and disagreement fails. (My first written split of that 36 was also wrong and was caught by counting.)

## Two self-corrections from the verification itself

1. **A fake RED.** Disabling the new check at its *call site* gave rc=1 with zero FAIL rows and `LEDGER_ADDED: unbound variable` — `set -u` killed the run before any verdict. Same exit code, zero evidence. Re-done at the *decision*, the defect visibly returns: rc=0 printing `ok ledger 37 entr(ies), 1 added`.
2. **A false RED.** The first shallow-clone simulation cloned from the local checkout, whose `main` is 8 commits behind and genuinely predates the ledger — `ABSENT` was the correct verdict for that input. Against the real remote it passes. Nearly filed a bug against the guard.

## Verification

```
full rc=0 · --selftest rc=0 · --bogus rc=2 · bash -n rc=0
bashrs 0 errors · shell lint ratchet rc=0 (149 error lines before and after — adds none)
guards_are_wired / no_timing_in_required / pass_grep_anchored /
no_claim_literals / no_competing_harnesses / dogfood_coverage — rc=0 each
```

Selftest gained 8 shrink-only + 3 comparand rows, each mutation-verified, including the comparand table building a throwaway git repo so the UNRESOLVABLE/ABSENT branches fail rather than silently skip.

## Residual, stated

Still keyed on `path:line`, so an unrelated edit above a ledgered site still reds. Content-addressing is a larger change. Note that the sibling `check_no_claim_literals.sh` was re-keyed to a content hash in #2711 for exactly this reason — that is **reported, not verified from here**, since #2711 is not in this branch's base.

The tip-fallback's false-red-on-a-stale-branch behaviour is reasoned from the subset property and the code path, **not observed** — no main that had already shrunk the ledger was constructed. The ~10 KB SIGPIPE onset is a single-host measurement.

Refs #2706

### DIFF
--- a/scripts/check_no_fabricated_baselines.sh
+++ b/scripts/check_no_fabricated_baselines.sh
@@ -59,8 +59,41 @@
 set -uo pipefail
 cd "$(dirname "$0")/.." || exit 2
 
+# ---------------------------------------------------------------------------
+# MODES, AND WHY AN UNKNOWN ARGUMENT IS FATAL.
+#
+# `--selftest` (the repo convention, cf. check_no_claim_literals.sh:140) runs the
+# case tables alone: the regexes and the ledger ratchet are exercised against
+# fixtures, with no repository scan. The plain invocation — what ci.yml:927 runs
+# — does the scan AND the tables.
+#
+# Before this pass the script read no arguments at all, so `--selftest` "worked"
+# by being IGNORED: it ran the full scan and exited 0, and so did `--slftest`,
+# and so did `--no-really-dont-check-anything`. Measured on the parent commit:
+# `bash scripts/check_no_fabricated_baselines.sh --this-flag-does-not-exist`
+# exited 0. A flag that is silently ignored is a flag that does not exist, and
+# the caller cannot tell the difference between "the mode ran" and "the mode is
+# not implemented" — this repo has already paid for that once, when a misspelled
+# nextest key meant NO timeout and killed the merge queue.
+MODE=full
+case "${1:-}" in
+    '')         : ;;
+    --selftest) MODE=selftest ;;
+    *)  printf 'usage: %s [--selftest]\n' "${0##*/}" >&2
+        printf 'refusing to run: unknown argument %s. Exiting 2 rather than\n' "$1" >&2
+        printf 'silently running something the caller did not ask for.\n' >&2
+        exit 2 ;;
+esac
+if [ "$#" -gt 1 ]; then
+    printf 'refusing to run: %s takes at most one argument, got %s.\n' "${0##*/}" "$#" >&2
+    exit 2
+fi
+
+TMPD=$(mktemp -d) || exit 2
+trap 'rm -rf "${TMPD:?}"' EXIT
+
 rc=0
-printf -- '--- no fabricated comparator baselines (F12) ------------------------\n'
+printf -- '--- no fabricated comparator baselines (F12) [mode: %s] -------------\n' "$MODE"
 
 # The competitor list is UNCHANGED by PERF-008 and is deliberately so: this pass
 # widens the SHAPE axis, and widening the NAME axis at the same time would leave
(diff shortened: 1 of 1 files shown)


noahgift:

PR #2716, opened 2026-08-28
fix(perf): the parity chain read a filename nothing writes — three spellings, zero blocks emitted

noahgift:
Epic #2706 (the receipt rule). PERF-004-SCHEMA was **held back last round on a refuted proof** — the commit claimed an end-to-end run that could not have executed. The rework found that the refutation's diagnosis was also wrong, in its object.

## The chain has never emitted a parity block

The earlier refutation said `parity_block.py` needs a `lanes.txt` the corpus does not contain. That is false: `scripts/parity_host_receipt.sh:124` writes `lanes.txt`, and it is the only caller of `parity_block.py`. The corpus under `evidence/parity-http/bands/` is a *different* artifact set, whose consumer is `perf_receipt.py --from-bands` — which reads `llamacpp-cN.json` correctly and always did.

Enumerating both sides mechanically found the actual break. Verified on `origin/main`:

| producer writes (`run_lane`, :96–124) | consumer reads (`parity_block._lane_from`) |
|---|---|
| `apr-<lane>-c<N>.json`, `llama-<lane>-c<N>.json` (`:104`) | ✔ band path, `:54-55` |
| `lanes.txt` (`:124`) | ✔ |
| — nothing — | **`apr-<lane>.json`, `llama-<lane>.json`** (`:76-77`) ✗ |

And `run_lane`'s own header comment at `:90` claims a **third** spelling, `$WORK/<class>.json`, matching neither side.

So every complete host run died on `FAIL lane cpu is missing a side` — which reads to an operator as *"the benchmark didn't run."* **The P2 chain has never emitted a parity block**, and `--from-parity` was reachable from no test.

**Fixed on the consumer side.** The lane-level side is the c=1 band, which is what it always was: `--concurrency` defaults to 1, and the corpus's own lane-level files are `concurrency=[1]` on every run. Moving the producer instead would buy a second independently-measured copy of a number the band sweep already has, and would stop `llama_pin.toml`'s `http_concurrency_bands` being the single statement of what was measured. **No fallback** — c=1 absent refuses by name and never silently promotes c=4.

## Run to a real verdict, and the verdict is FAIL

```
python3 scripts/lib/perf_receipt.py --from-bands <live>/bands --subject apr \
  --comparator llamacpp --host lambda-vector --workload W1 --binary-sha256 9645036…1d01c9 …
  -> rc=0, receipt4.json, 8158 bytes

bash scripts/perf_gate.sh --host lambda-vector --phase merge --workload W1 --receipt receipt4.json
  -> rc=1
  FAIL ArmC completed(26) != requested(58)
  FAIL ArmC band 16: zero-token response is a failure, not a fast request
  FAIL ArmC timeouts / tokenization.method / drain_ms absent  (each named, owner PERF-004)
  VERDICT FAIL host=lambda-vector phase=merge workload=W1
```

Read back out of the artifact: `requested/completed = 58/26`, `len(samples_ms) = 26`, `provenance.binary_sha256` matches `sha256sum`.

Before and after the fix on the same artifacts, the failure **moved** — which is the evidence that the fix reached the data:
- pre-fix: `FAIL lane cpu is missing a side` (never reached the data)
- post-fix: `FAIL band[16] contains a non-positive rate` (reached it, refused correctly — apr timed out 16/16 at c=16, `elapsed_secs = 120.0`, exactly the reqwest bound)

A live P2 `rc=0` was **not** obtained and is not claimed. The `rc=0` path is proven by the selftest running the real scripts as subprocesses.

**The ratios here are not a parity result and are cited as one nowhere.** CPU-only binary, box at load 16–74, 10 s runs. The receipt is deliberately left in scratch, **not** under `evidence/`.

## The three landmines, checked rather than assumed

- **`perf_gate.sh`'s real mode is invoked by nothing.** #2709 wired `--selftest`. The real mode (`--host/--phase/--workload/--receipt`) still has no caller, and neither does `parity_host_receipt.sh`. Every remaining reference is inside `check_no_competing_harnesses.sh`, which allowlists it as *"THE entrypoint"* — the canonical measurement entrypoint is referenced only by the guard that declares it canonical. A join between two scripts nothing runs together is exactly why the spelling mismatch survived.
- **`drain_ms` and `tokenization.method` fail ALWAYS, deliberately.** Both are `class: UNMEASURED, required: always` with zero producers. Proven by the live gate FAILing on all three. Not a mirror defect, because only `--selftest` is in a required check.
- **8/8 `perf-matrix.yaml` cells `UNMEASURED`** — confirmed.

## Two defects fixed that this work itself created or exposed

- The cherry-pick predated #2709 and added a **second** `perf_gate.sh --selftest` CI step under a different name. Removed.
- **`--resolution` defaulted to `"scripts/apr_bin.sh"`** — writing the one provenance claim that matters into every receipt whose caller forgot the flag. A default that fabricates provenance is worse than a missing field, because it is indistinguishable from a real answer. Now required, with a RED row.

## Mutation table

| mutation | result |
|---|---|
| restore the pre-fix `apr-%s.json` read | **RED** 2 broke / 12 passed |
| add a silent fallback to any present band | **RED** 1 broke / 18 passed |
| unmutated | GREEN 19/19 |
| omit `--resolution` | RED, refuses |
| `perf_gate.sh --selftest` | 18/18, incl. `timeouts_absent_is_not_zero` |
| schema-map guard `--self-test` | 8/8, 5 RED + 3 discrimination |

`--from-bands` rows stay green under both P2 mutations, so the RED is discrimination rather than blanket refusal.

## Binary provenance

`bash scripts/apr_bin.sh` → **rc=1, STALE**: there is no HEAD-built `apr` on this box, so no step could have used one. `~/.local/bin/apr` is `v0.64.0+no-git` — no commit at all — and wins a bare `apr`. Executed steps used `/home/noah/.cargo/bin/apr`, `0.64.0 (ce712eae0)`, sha256 recorded in the receipt; comparator `llama-server` `7746 (39173bcac)` = the pin.

The bound on what that licenses was verified, not asserted: the subject under test is the converter and gate **at HEAD** (the Python and Bash in this checkout); apr is the *instrument*, recorded by digest. The Rust delta `ce712eae0..HEAD` is five files, all `apr profile` (PERF-016), reached only via `ExtendedCommands::Profile`; neither executed step is in that path. Rebuilding was **declined with a reason, not skipped** — `cargo install --force` overwrites `~/.cargo/bin/apr` while other agents are running it, and would change nothing this run establishes.

**Any statement about apr's speed from this run is unproven, and is made nowhere.**

Refs #2706

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -995,6 +995,29 @@ jobs:
       # the label of a performance result (#2677).
       - name: Benchmark fairness protocol is complete
         run: bash scripts/check_bench_protocol.sh
+      # PERF-004. scripts/perf_gate.sh read a receipt schema NOTHING in the tree
+      # wrote: four real artifacts all returned rc=1 and the only input that
+      # passed was the gate's own synthetic fixture. The tempting repair is to
+      # default the missing fields, which would hand-assign five numbers no
+      # instrument produces. scripts/perf-receipt-fields.yaml classifies every
+      # field the gate reads as PRODUCED, DERIVED, POLICY or UNMEASURED, and
+      # this holds the classification to the tree in both directions -- a new
+      # unclassified requirement, and an entry whose producer has since been
+      # deleted, are each a FAIL.
+      - name: Every field the perf gate reads has a producer or a named owner
+        run: bash scripts/check_perf_receipt_fields_have_producers.sh
+      - name: Schema-map guard can still go RED (case table + 5 mutations)
+        run: bash scripts/check_perf_receipt_fields_have_producers.sh --self-test
+      # The converter that closes the loop, end to end: a BenchmarkReport this
+      # repo can actually produce reaches a verdict, and what remains failing is
+      # the UNMEASURED bucket BY NAME rather than a schema error.
+      - name: Receipt converter round-trips into a gate verdict
+        run: python3 scripts/lib/perf_receipt.py --selftest
+      # NOT a second `perf_gate.sh --selftest` step. This commit was authored
+      # before #2709 landed, and carried its own copy; two identically-scoped
+      # invocations under two names is how a check comes to look twice as
+      # covered as it is. The case table runs once, above, as "Perf gate can
+      # still discriminate".
       # The deepest check in the surface sweep -- real generation through
       # /v1/chat/completions -- had never executed once: the probe binary was
       # built without --features llm (clap advertises `llm` regardless, only the
--- a/scripts/check_no_timing_in_required.sh
+++ b/scripts/check_no_timing_in_required.sh
@@ -93,6 +93,7 @@ check_llama_pin.sh
 check_no_fabricated_baselines.sh
 check_no_timing_in_required.sh
 check_perf_claims_cite_receipts.sh
+check_perf_receipt_fields_have_producers.sh
 "
 # check_perf_claims_cite_receipts.sh (PERF-010) matches `check_*perf*.sh` and
 # turned this guard RED the moment the file appeared — verified before wiring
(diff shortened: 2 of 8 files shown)


noahgift:

PR #2717, opened 2026-08-28
feat(perf): the perf gate requires drain_ms, and nothing on earth writes drain_ms

noahgift:
Epic #2706 (the receipt rule). PERF-024 was filed by me on a **false premise** — I audited `benchmark_runner.rs`, which turned out to be a dead byte-identical duplicate of `preflight.rs`, reachable only from benches and tests. The ticket was re-aimed mid-flight from "build a client" to "audit `LoadTest::run()` against §4.4 and close what is actually missing."

## The finding that matters: the perf gate has never rendered a verdict on a real receipt

Two defects, each harmless-looking alone, that conceal each other:

```
scripts/perf_gate.sh:42     if r.get("drain_ms") is None: bad.append("drain_ms absent")
drain_ms producers, repo-wide, outside the spec/gate/matrix:   0
```

**`perf_gate.sh` fails any receipt omitting `drain_ms`, and nothing anywhere writes `drain_ms`.** No real receipt can satisfy the gate. And per #2716, `perf_gate.sh`'s real mode (`--host/--phase/--workload/--receipt`) has **no caller** — every remaining reference is inside `check_no_competing_harnesses.sh`, which allowlists it as *"THE entrypoint."*

A gate that cannot pass is invisible while nothing runs it. A gate nothing runs looks fine if you assume it would pass. Together: the epic's central enforcement mechanism has never judged a real measurement, and neither half would have shown that on its own.

## The §4.4 audit of `LoadTest::run` at `de8fbc407`

My three original claims, tested against both files:

| claim | `benchmark_runner.rs` | the real client (`loadtest.rs`) |
|---|---|---|
| zero concurrency primitives | **confirmed** | refuted — `:568` `for worker_id in 0..self.config.concurrency` |
| `mean_throughput` = mean of rates | **confirmed** (`:327`) | refuted — `:519-521`, `:834` `total_tokens / elapsed_secs`, wall-clock |
| no ttft/itl | **confirmed** | refuted — `ttft_p50/p90/p95/p99_ms`, `itl_p50_ms`, `tpot_p*_ms` |

**Genuinely missing**, and now implemented: `client_model` record; warmup as `2×c` *requests* (it was a duration, defaulting to zero); the **5 s quiesce** (Phase 2 → Phase 3 with no sleep); the min-sample bound; last-bound-wins termination; `timeouts` as its own counter; §4.4.4 bootstrap (it used a t-distribution over N run-level means).

**Partial, and worth knowing:** `decode_tok_per_sec` was `1000/median-ITL` rather than the median of per-request rates; `itl_p50_ms` pooled one *mean per request* instead of every gap; no `ttft` p95. And `truncated_pct` means `finish_reason=="length"` — a different thing from §4.4.7's "abandoned at drain". Conflating them would zero out W1's entire numerator, since W1 ignores EOS.

## Proofs

**Concurrency proved from the server side.** A loopback probe counts its *own* concurrent connections, so client-side bookkeeping cannot fake it: c=8 → server_peak **8**; c=1 → server_peak **1**. Same 16 requests: c=1 770 ms, c=8 186 ms, **4.14×**.

**`agg_tok_s` fixture, both directions.** Requests spanning [0,1] [0,2] [1,3] [2,4], 100 tokens each → span 4.0 s, 400 tokens → **agg = 100.0**; mean-of-rates = **62.5**. A second fixture that serialises with idle gaps gives agg **57.1** against mean-of-rates **100.0** — the overstatement direction, asserted explicitly so the two can never be confused again.

**TTFT/ITL end-to-end** against an SSE probe emitting 6 tokens 25 ms apart: `ttft_p50=26.7ms`, `itl_p50=26.2ms`, five *pooled* gaps per request.

**Bootstrap determinism**: same samples + seed 2026 → byte-identical CI twice; seed 2027 moves the interval while the point estimate holds.

**Mutations**, each reverted with the tree clean: `agg_tok_s := mean_of_rates` → 2 RED · termination `AND := OR` → 5 RED · seed := clock → 3 RED · `0..concurrency := 0..1` → 4 RED including the real-HTTP proof.

## Not proven — read this before relying on it

- **Nothing calls `run_band` outside tests.** `apr test llm bench` still calls `LoadTest::run`. CLI wiring and receipt serialisation are follow-ups, and the protocol was **not** verified against a real model server.
- **§4.4.9's scheduler block is not implemented and cannot be by a client** — those fields are server-reported. `WindowReport::client_peak_in_flight` is named so it cannot be mistaken for `max_in_flight`.
- **`receipt_size_budget_bytes` is deliberately unset** — §4.4.5 forbids the literal until a full receipt is measured.
- **Non-streaming requests yield no ttft/itl/decode.** Left empty, never synthesised. A conformant band must use `stream = true`; no guard yet refuses `stream = false`.
- No baseline was measured or committed. `scripts/perf-matrix.yaml` is untouched, all 8 cells still `UNMEASURED`.

No `apr` binary was invoked at all — every result is source-level analysis of `origin/main` plus `cargo test` compiled from this worktree, so nothing here depends on an installed binary.

## Three findings to file separately

1. **`preflight.rs` is a dead byte-identical duplicate of `benchmark_runner.rs`** (`diff -q` rc=0, 370 lines, only the latter `include!`d). It reads exactly like the production client and is unreachable — it is what caused this ticket to be filed on a false premise.
2. **`aprender-test-lib`'s `llm` tests are dark under `cargo test -p aprender-test-lib`** (non-default feature) but **do** run under CI's `cargo nextest run --workspace --lib`, because `apr-cli` pulls `apr-test` with `features=["llm"]` and workspace feature unification turns it on. Drop that one edge and 700+ tests go dark **silently**. The new protocol is placed outside the feature so it is gated either way.
3. **The PMAT pre-commit complexity gate makes `loadtest.rs` uneditable** — 157/171 on `origin/main`, identical before and after, so the gate blocks *any* commit touching it, including a pure visibility change. `band.rs` was reworked to use `LlmClient` directly so `loadtest.rs` is untouched to the byte, but the next person who must edit it will be blocked and may reach for `--no-verify`.

## Verification

`cargo test -p aprender-test-lib --lib` 6356 pass (52 `perf_gate`) · `--features llm --lib` 6533 pass (52 + 8 `band`) · `cargo check -p apr-cli --lib` rc=0 · `cargo fmt --all -- --check` rc=0 · `cargo clippy -p aprender-test-lib --features llm --all-targets -- -D warnings` rc=0.

One incident worth recording: an initial `src/perf/` clobbered two existing files there; both were restored from git (`git diff` clean) and the work relocated to `src/perf_gate/`.

Refs #2706

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1522,6 +1522,7 @@ dependencies = [
  "chrono",
  "criterion 0.5.1",
  "crossterm 0.28.1",
+ "flate2",
  "futures",
  "gif 0.13.3",
  "image",
--- a/crates/aprender-test-lib/Cargo.toml
+++ b/crates/aprender-test-lib/Cargo.toml
@@ -59,6 +59,12 @@ mp4 = { workspace = true, optional = true }
 crossterm = { workspace = true, optional = true }
 regex = { workspace = true }
 sha2 = { workspace = true }
+# PERF-024 (APR-PERF-GATE-001 v2.2 §4.4.5): raw per-request samples are retained
+# as gzipped JSONL beside every receipt -- a summary-only receipt cannot be
+# resampled and is rejected. Not feature-gated: `src/perf/` must compile (and
+# its tests must RUN) under default features, because CI is
+# `cargo nextest run --profile ci --workspace --lib` with no --features.
+flate2 = "1"
 serde_yaml_ng = { workspace = true }
 
 # Tracing and UUID (available on all targets)
--- a/crates/aprender-test-lib/src/lib.rs
+++ b/crates/aprender-test-lib/src/lib.rs
@@ -564,6 +564,14 @@ pub mod animation;
 )]
 pub mod presentar;
 
+/// APR-PERF-GATE-001 v2.2 §4.4 — the serving-performance measurement protocol.
+///
+/// Termination rule, metric definitions, bootstrap CI, and raw-sample retention
+/// for `apr test llm bench`. Ungated on purpose: CI runs
+/// `cargo nextest run --profile ci --workspace --lib` with no `--features`, so a
+/// protocol placed behind the `llm` feature would never be tested in CI.
+pub mod perf_gate;
+
 /// LLM Testing: Correctness assertions and load testing for OpenAI-compatible APIs.
 ///
 /// Feature-gated behind `llm`. Provides HTTP client, assertion builders,
--- a/crates/aprender-test-lib/src/llm/band.rs
+++ b/crates/aprender-test-lib/src/llm/band.rs
@@ -1,0 +1,772 @@
+//! APR-PERF-GATE-001 v2.2 §4.4 — the protocol, driven through the existing
+//! `apr test llm bench` request path.
+//!
+//! This is **not a second harness.** It drives [`LlmClient`] — the one HTTP
+//! client [`super::loadtest::LoadTest`] drives, and the one
+//! `scripts/parity_host_receipt.sh` points at both `apr serve` and
+//! `llama-server` (§4.4.8: one client, both servers, or the ratio is refused).
+//! What it adds is the part `LoadTest` never had:
+//! the §4.4.2 termination rule, the §4.4.2 warmup-then-quiesce, the §4.4.3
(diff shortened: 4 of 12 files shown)


noahgift:

PR #2719, opened 2026-08-28
batch: five receipt-rule sub-tickets, plus four benches that had not compiled since April

noahgift:
Epic #2706 (the receipt rule). Five sub-tickets folded into one integration branch, plus two pre-existing main defects the batch exposed.

**Why batched:** the merge queue is `max_entries_to_build: 1` with a 30–64 min required check, so it drains at roughly one PR per hour. Five separate PRs is a five-hour serial drain; this is one slot. It also does what stacking cannot — a prior 9-PR batch here surfaced four defects invisible to every individual PR, and **this one surfaced two**.

Supersedes #2713 (PERF-019), #2714 (PERF-010), #2715 (PERF-008), #2716 (PERF-004), #2717 (PERF-024).

## The five, in one line each

- **PERF-019** — the one-client gate passed with its **walker dead**. Six mutations of the walker alone each returned rc=0 with output byte-identical to a clean sweep. Also closed an untracked-file free pass, re-mutated in the widened scope.
- **PERF-010** — the ratio guard caught **1 of 6 shapes**; `2.93× Ollama` (U+00D7) was invisible while its ASCII twin was caught, so the epic's most-quoted fabrication was undetectable in its own preferred spelling. Refuted comparator claims deleted from `book/`.
- **PERF-008** — the ledger printed *"the count is enforced, not asserted"* and enforced nothing: appending one line laundered a fresh fabrication. Now a subset property against a ref the author cannot rewrite.
- **PERF-004** — the parity chain read a filename nothing writes (**three** spellings across producer, consumer, and the producer's own header comment), so it had never emitted a parity block.
- **PERF-024** — `perf_gate.sh` requires `drain_ms`; **zero producers exist repo-wide**. Combined with PERF-004's finding that its real mode has no caller, the gate has never judged a real receipt.

## Two defects the batch exposed that no branch could

**1. A doc comment restating figures the assertions already pin.** PERF-024's metrics module documented its fixture with a bare `100 tok/s`; PERF-010's widened matcher is what catches it. Neither branch carries both. The same two numbers were stated twice — prose and assertions — free to drift. Now only the assertions carry them.

**2. Four bench targets that had not compiled in months.** `cargo check --workspace --benches` fails on `main`. Not caused by this batch — none of the five branches touches `aprender-serve` or `aprender-profile`.

| file | drift |
|---|---|
| `external_matrix.rs` | `GGUFConfig.query_pre_attn_scalar` |
| `comparative.rs` | `query_pre_attn_scalar`, `AprTransformer.lm_head_tied` |
| `performance_parity.rs` | `query_pre_attn_scalar`, both `OwnedQuantizedLayer.post_*_norm_weight` |
| `visualization_simd.rs` | **deleted** |

`visualization_simd.rs` was a named 227-line `[[bench]]` whose only non-criterion import was `renacer::visualize::ring_buffer::HistoryBuffer`. That module was deleted on **2026-04-08** by `c0087b8a3` ("delete 32 dead ratatui source files", #701), which removed the module and left the bench. It has not compiled for four and a half months and cannot be repaired without resurrecting deliberately-deleted files.

**Root cause fixed, not just the symptom.** `ci.yml` builds `--examples` and never benches. Its own header at `:6` claims `cargo clippy --all-targets -- -D warnings`, and the file contains **zero** clippy invocations — lint is delegated to the upstream reusable workflow, whose clippy does not reach benches either. So bench rot is structurally invisible.

```
  cargo build --examples --workspace --keep-going
+ && cargo check --workspace --benches --locked
```

Mutation-verified: reverting the `external_matrix` field takes the new step to rc=101 naming that exact E0063; restoring returns rc=0. **The first restore still failed** — which is how the other three files were found. One target's fix exposes the next.

## The merge that needed judgement

Both parents carried a baseline entry **numbered `:336` meaning different lines**, because 019 edited a file 010 had baselined. Taking either side wholesale silently drops the other's coverage. Resolved by locating each entry's **content** in the merged tree:

```
:235 -> :263   '### 5.2 Failure #4 — 2.93× Ollama from a harness that…'
:336 -> :364   '| --gpu fails loudly on a CPU-only build (Jidoka)'
```

One entry was then missing. Before restoring it I verified the covering line was **byte-identical on both parents** (`main:308` == `019:336` == `merged:336`), so it is a merge repair and not ledger growth — which matters, because growth-laundering in this exact file is what PERF-008 fixes.

## Verification

```
cargo fmt --all -- --check                 rc=0
cargo check --workspace --benches --locked rc=0   (was 101 on main)
cargo test -p aprender-test-lib --lib      6356 pass
94 shell gates named in ci.yml             94/94 rc=0
```

Gate list derived from `ci.yml` at run time rather than a hand-kept list, so a gate cannot drop out of the sweep silently.

## Known residual, stated

The claim-literal ledger still accepts **growth** — appending an entry launders a new violation. I verified this on `main`, on #2711's content-hash re-key, and swept it: **12 of 12 baseline files in `scripts/` accept growth**, with laundering demonstrated on 2. PERF-008 fixes that property for the Rust ledger only. The general fix is the same pattern applied to the other eleven, and it is a follow-up, not this PR.

Refs #2706


noahgift:
Superseded by **#2742**, which is this same five-sub-ticket batch merged onto today's main (`50d2bc2bb`) rather than the base this branch was cut from — plus the resolution of a genuine conflict between PERF-010's detector widening and PERF-028's diff-ratchet, both of which have now landed.

#2742 is fully green (97/97 guards) and is a **strict improvement on two axes**: `cargo check --workspace --benches` goes rc=101 → rc=0, and clippy from 88 failing targets to 85 with zero regressions.

Nothing is lost — the sub-branch merges and their authorship survive, because it was merged rather than rebased. Closing in favour of #2742.

PR #2720, opened 2026-08-28
perf(beat): run the speed lane on perf-solo and make it name its own runner (Refs PERF-031)

noahgift:
## What

PERF-013 (paiml/infra#338) shipped the `perf-solo` label at cardinality 1 on
`intel-clean-room-16` and closed with one item explicitly unproven:

> **No job has actually run on the label** — that needs a workflow commit, which was out of scope.

This is that commit. It puts `perf-solo` on the BEAT speed lane's `runs-on` and
makes the job **name its own runner**, so the claim stops resting on config and
starts resting on a job's own output.

## Why `RUNNER_NAME` is the load-bearing line

All 16 agents on `intel` report the same `hostname` (`mac-server`). Hostname
therefore cannot say *which agent served a job* — and that is the entire claim
the lane makes. `RUNNER_NAME` is the only field that identifies the lane.

PERF-013's proposed receipt recorded neighbours, load, and hostname. It did not
record the runner name, which means it could not actually have proven the thing
it was for. That field is the substantive addition here.

## What the lane does NOT buy — recorded in the file

A label cannot make the box quiet. `intel` runs 16 runners **by design** and
over-subscription is provisioned deliberately, so `perf-solo` removes
**perf-vs-perf** contention only. The job is alone in its **lane**, never alone
on the **box**.

So the neighbour count is part of the receipt, sampled at **both ends**: a ratio
cancels steady load, but not a neighbour arriving or leaving mid-run. A reader
who takes an absolute number from this lane can now see whether the box was
steady while it was taken.

## Receipt, not gate

Neither step can fail the build. Asserting on neighbour count would put an
environment-dependent red into a lane that exists to measure code — against the
standing rule about wall-clock/environment assertions in checks.

## Trade-off the reviewer owns

The lane is now **one agent wide**, so this job can QUEUE behind general
clean-room work on `intel-clean-room-16`. `timeout-minutes` is unchanged at 45.
Whether 45 still fits, or the 05:45 UTC schedule should move, is a judgement
call left to review rather than made silently here.

## Safety notes

- **No `${{ }}` in any added comment.** Actions substitutes expressions
  textually and does not honour `#`; a comment quoting one has broken a job in
  this repo before. The single value crossing between steps travels via `env:`,
  never spliced into a script body.
- **`comm` match, not an args regex.** Runner 2.336 installs the worker under a
  versioned directory (`bin.2.336.0/Runner.Worker`), so a `bin/Runner.Worker`
  pattern counts **zero** while 16 jobs run. Verified live on `intel`: `comm`
  form = 16, args-regex form = 0, at the same instant.
- **No collision with the blackout auditor.** `check_beat_measurements.sh`
  derives its markers from the `--test beat_*` references, not from log text, so
  the `BEAT-ENV` lines are invisible to it.
- Nothing about the runner fleet is changed by this PR. Cardinality stays
  enforced in `paiml/infra` by `make -C machines/intel verify-perf-solo-lane`,
  not duplicated here, so the two cannot drift.

## Verification

Repo guards, run against the patched file:

| guard | rc |
|---|---|
| `check_beats_gated.sh` | 0 — all 26 beats execute |
| `check_workflow_env_defined.sh` | 0 — 12 workflows |
| `check_workflow_path_filters.sh` | 0 |
| `check_beat_baseline_env.sh` | 0 — 4 incumbents importable |

Plus: YAML parses, and a scan asserts zero comment lines contain an Actions
expression.

**End-to-end**: `beat-speed-nightly.yml` is `workflow_dispatch`-able, so this
branch's version was dispatched against its own ref to land a real job on the
lane without merging anything first. The result is reported in PERF-031.

Refs PERF-031 · paiml/infra#338

### DIFF
--- a/.github/workflows/beat-speed-nightly.yml
+++ b/.github/workflows/beat-speed-nightly.yml
@@ -41,7 +41,16 @@ concurrency:
 
 jobs:
   beat-speed:
-    runs-on: [self-hosted, X64, Linux, clean-room]
+    # PERF-013/PERF-031. `perf-solo` is carried by exactly ONE runner
+    # (intel-clean-room-16; paiml/infra#338), so two speed runs can never
+    # overlap and time each other. Verified server-side: this five-label
+    # selector resolves to 1 agent, the four-label one resolves to 16.
+    # Cardinality is enforced by `make -C machines/intel verify-perf-solo-lane`
+    # in paiml/infra, not here.
+    #
+    # TRADE-OFF: this lane is one agent wide, so the job can QUEUE behind
+    # general clean-room work on that runner. See timeout-minutes below.
+    runs-on: [self-hosted, X64, Linux, clean-room, perf-solo]
     timeout-minutes: 45
     env:
       # Every beat step tees here; the final step audits it for completeness.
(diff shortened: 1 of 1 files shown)


noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-NOW → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `.github/workflows/beat-speed-nightly.yml`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. main's beat-speed-nightly.yml now runs on [self-hosted, X64, Linux, clean-room]; this PR moves the lane to perf-solo. Switching a runner is escalate-class (feedback_small_ci_edits_authorized), so the operator resolves it. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — disposition revised by the operator: **HOLD** — held on D-3 decision (DEC-D-3, issue #2934). The perf-solo runner move waits for that decision; no branch edits from triage. Labels: `pp-066/hold` (merge/conflict removed).

PR #2721, opened 2026-08-28
fix(lint): clear the clippy 1.98 ceiling drift Toolchain Ceiling has been reporting since 08-20

noahgift:
## The defect

`aprender/Toolchain Ceiling` is one of the seven lanes the infra nightly audit's
`lane-liveness` dead-man's switch reported dead on 2026-08-27
(infra run 33071842506):

    aprender/Toolchain Ceiling   DEAD no-healthy 8d (>3d) — not allowlisted

The lane is doing precisely what aprender#2370 built it to do. It was green
through 2026-08-19 and has failed every scheduled run since 2026-08-20 — Rust
1.98.0 was released 2026-08-18 — with the message it exists to print:

    ✗ TOOLCHAIN CEILING DRIFT: the tree does NOT pass clippy on current stable
      (1.98.0). It still passes on the pinned 1.93.0, which is why no PR
      gate caught this - see aprender#2370.

## 71 findings, not 2

`cargo clippy --all-targets -- -D warnings` stops at the first crate that fails,
so the nightly log only ever showed the first wave (two `chunks_exact` hits in
`apr-format`). Each fix revealed the next. **Ten passes to reach zero**, nine
lints, all new in 1.98:

| count | lint |
|---|---|
| 57 | `chunks_exact_to_as_chunks` |
| 6 | `unused_async_trait_impl` |
| 3 | `unnecessary_sort_by` |
| 2 | `unnecessary_trailing_comma` |
| 1 each | `manual_slice_fill`, `manual_midpoint`, `manual_is_variant_and`, `manual_checked_ops`, `explicit_counter_loop`, `collapsible_match` |

`as_chunks::<N>()` is the bulk, and it is a real improvement rather than a
respelling: it yields `&[T; N]`, so the chunk width is a **type** instead of a
runtime claim, and nine `chunk.try_into().expect("chunk is N bytes")` calls
simply disappear.

## What is not mechanical, and why

Six `unused_async_trait_impl` hits. Two fixed, four left `async` behind a
documented `#[allow]`, for four different reasons:

- **aprender-db `MemoryKvStore` — FIXED, not allowed.** The trait already
  declares `-> impl Future<..> + Send`, every operation is a synchronous DashMap
  access, and clippy is right. (`+ Send` is carried over from the trait —
  clippy's suggestion drops it, which would not compile.)
- **aprender-orchestrate `estimate_rust_score` / `estimate_repo_scores` —
  allowed, having tried the fix.** De-asyncing them moved the lint up to
  `run_rust_project_score` / `run_repo_score`; de-asyncing those moves it to the
  **public** `check_component`. Nothing in `QualityChecker` ever awaits — every
  leaf is a blocking `std::process::Command`. Making that module honestly
  synchronous is worth doing and is not a toolchain-drift fix.
- **aprender-graph `read_parquet`/`write_parquet`, aprender-test-lib
  `fetch_bytes` — allowed.** All three already carried
  `#[allow(clippy::unused_async)]` with an author-written reason; this is the
  same finding under a new lint name. For the parquet pair, the suggestion would
  additionally run the write **eagerly at call time** instead of at await time.
- **aprender-test-lib `GpuTelemetryCollector::start` — allowed.** Public `async
  fn` that callers `.await`, and clippy's own suggestion does not compile here
  (it proposes wrapping one branch of an `if` in `std::future::ready`).

Every one of those allows is written `#[allow(unknown_lints,
clippy::unused_async_trait_impl)]`. **Without `unknown_lints` the pinned 1.93
toolchain fails** with `error: unknown lint: clippy::unused_async_trait_impl` —
a ceiling fix that breaks the floor. Caught by running both, not by reading.

## A defect introduced by clippy's own autofix

`manual_slice_fill`'s machine-applicable suggestion is
`&mut self.rdp.fill(0.0);` — a borrow of `()`. `cargo clippy --fix` applied it;
the next pass caught it with `unnecessary_operation` + `unused borrow that must
be used`. It is fixed here, and it is why every pass was re-run rather than
trusted.

## One extraction that is not a clippy finding

`AprReader::from_bytes` is split, because the repo's own pre-commit gate refused
the commit:

    Complexity check... ❌
      crates/aprender-core/src/serialization/apr/mod.rs
      from_bytes (line 62) - Cognitive 26 > 25

That is **pre-existing** debt, not caused by this change — verified by running
`pmat analyze complexity` against `git show HEAD:` of the same file, which
reports the identical 26 > 25. The hook only inspects staged files, so touching
four unrelated lines is what made it visible. `--no-verify` is not an option, so
the 20 straight `if let Some(..)` metadata copies move to `v2_metadata_to_flat`
— a pure code move. The file now reports no function over either threshold.

## Verification

Every gate this lane and the normal CI lint job run, measured locally:

    $ cargo +1.98.0 clippy --all-targets -- -D warnings     # the ceiling gate
      Finished ... EXIT=0                                   # was: 71 findings
    $ cargo +1.93.0 clippy --all-targets -- -D warnings     # the pin, normal CI
      Finished ... EXIT=0
    $ bash scripts/check_msrv.sh                            # the lane's msrv-floor job
      ✓ workspace builds on its declared MSRV 1.91
    $ cargo fmt --all -- --check                            # exit 0
    $ pmat analyze complexity --file .../apr/mod.rs --max-cognitive 25
      no functions found exceeding the thresholds

Tests, on the pinned 1.93 toolchain, over the nine crates carrying the
behavioural rewrites (`apr-format`, `aprender-core`, `aprender-compute`,
`aprender-zram-core`, `aprender-present-core`, `aprender-db`,
`aprender-orchestrate`, `aprender-contracts-cli`, `aprender-graph`):

    26,414 passed; 0 failed; 13 ignored

Honest caveat: that test run predates the `from_bytes` extraction (a pure code
move made afterwards to satisfy the complexity gate); the re-run of
`aprender-core`'s suite had not finished when this PR was opened. `Toolchain
Ceiling` was also dispatched on this branch (run 33163324585) but the intel pool
was saturated — 16/16 runners busy — so its result is not in this description.
Read both before merging.

Note also that local `stable` on the box I used is 1.97.1, so
`check_clippy_current_stable.sh` would have linted the wrong toolchain here;
1.98.0 was named explicitly, which is what CI's `rustup toolchain install
stable` resolves to today.

### DIFF
--- a/crates/apr-format/src/v2/reader_impl.rs
+++ b/crates/apr-format/src/v2/reader_impl.rs
@@ -310,9 +310,17 @@ impl AprV2Reader {
         }
 
         let data = self.get_tensor_data(name)?;
+        // `as_chunks::<4>()` rather than `chunks_exact(4)`: same semantics
+        // (the ragged tail is the discarded `.1`), but it yields `&[u8; 4]`,
+        // so `from_le_bytes` takes the array directly instead of a
+        // reassembled one. clippy::chunks_exact_to_as_chunks (new in 1.98)
+        // flags the old form; caught by the Toolchain Ceiling lane, which is
+        // exactly the drift it exists to catch.
         let floats: Vec<f32> = data
-            .chunks_exact(4)
-            .map(|chunk| f32::from_le_bytes([chunk[0], chunk[1], chunk[2], chunk[3]]))
+            .as_chunks::<4>()
+            .0
+            .iter()
+            .map(|chunk| f32::from_le_bytes(*chunk))
             .collect();
 
         Some(floats)
@@ -440,9 +448,17 @@ impl<'a> AprV2ReaderRef<'a> {
         }
 
         let data = self.get_tensor_data(name)?;
+        // `as_chunks::<4>()` rather than `chunks_exact(4)`: same semantics
+        // (the ragged tail is the discarded `.1`), but it yields `&[u8; 4]`,
+        // so `from_le_bytes` takes the array directly instead of a
+        // reassembled one. clippy::chunks_exact_to_as_chunks (new in 1.98)
+        // flags the old form; caught by the Toolchain Ceiling lane, which is
+        // exactly the drift it exists to catch.
         let floats: Vec<f32> = data
-            .chunks_exact(4)
-            .map(|chunk| f32::from_le_bytes([chunk[0], chunk[1], chunk[2], chunk[3]]))
+            .as_chunks::<4>()
+            .0
+            .iter()
+            .map(|chunk| f32::from_le_bytes(*chunk))
             .collect();
 
         Some(floats)
--- a/crates/aprender-compute/src/hash/mod.rs
+++ b/crates/aprender-compute/src/hash/mod.rs
@@ -45,12 +45,13 @@ pub fn hash_bytes(bytes: &[u8]) -> u64 {
     let mut hash: u64 = 0;
 
     // Process 8 bytes at a time
-    let chunks = bytes.chunks_exact(8);
-    let remainder = chunks.remainder();
+    // `as_chunks::<8>()` yields `&[u8; 8]` directly, so the fallible
+    // `try_into().expect(...)` below is gone: the 8-byte width is now a type,
+    // not a runtime claim. clippy::chunks_exact_to_as_chunks (new in 1.98).
+    let (chunks, remainder) = bytes.as_chunks::<8>();
 
     for chunk in chunks {
-        let word =
-            u64::from_le_bytes(chunk.try_into().expect("chunks_exact(8) guarantees 8 bytes"));
+        let word = u64::from_le_bytes(*chunk);
         hash = hash.rotate_left(5).bitxor(word).wrapping_mul(K);
     }
 
--- a/crates/aprender-contracts-cli/src/commands/certify.rs
+++ b/crates/aprender-contracts-cli/src/commands/certify.rs
@@ -221,7 +221,7 @@ fn print_summary(
         "✗"
     };
     eprintln!();
-    eprintln!("pv certify — {icon} {edges_satisfied}/{edges_total} composition edges satisfied",);
+    eprintln!("pv certify — {icon} {edges_satisfied}/{edges_total} composition edges satisfied");
     if let Some(cfg) = config_proof {
         if let Some(tensors) = cfg.get("expected_tensors") {
             eprintln!("  Config: {tensors} expected tensors");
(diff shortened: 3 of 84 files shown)


noahgift:
Follow-up on the caveat in the description: the `aprender-core` re-run **after** the `from_bytes` extraction has now finished on the pinned 1.93 toolchain —

    cargo +1.93.0 test --lib -p aprender-core
    test result: ok. 14256 passed; 0 failed; 2 ignored

so the whole nine-crate battery (26,414 tests) is green against the tree as pushed, not just against the tree before that code move.

`Toolchain Ceiling` run 33163324585 on this branch is still executing; the intel pool was 16/16 busy when it was dispatched.

noahgift:
**`Toolchain Ceiling` run [33163324585](https://github.com/paiml/aprender/actions/runs/33163324585) on this branch: SUCCESS — both jobs green.**

    clippy-current-stable  success
    msrv-floor             success

    Current stable: rustc 1.98.0 (88d9e12ae 2026-08-18) / clippy 0.1.98
    ✓ clean under current stable clippy (1.98.0); the pin is 1.93.0
    ✓ workspace builds on its declared MSRV 1.91

That is the lane itself, on CI's own `stable`, not a locally-named toolchain — the caveat in the description is now closed on both counts.

noahgift:
Force-pushed one correction, found by this repo's own gates rather than by me.

`guard-runner-labels` went red on the previous head at step **"Dogfood coverage may not regress"**:

    G2.1 freshness  FAIL  this branch changes 3 cited evidence file(s) and does
                          NOT touch docs/audits/surface_audit.csv:
        - crates/apr-cli/src/bin/apr-corpus-ingest.rs
        - crates/aprender-orchestrate/src/cli/agent.rs
        - crates/aprender-test-cli/src/dev_server.rs

`cargo clippy --fix` had swept those three up as well (`map(..).unwrap_or(..)` → `map_or`, and trailing commas). **None of them is a finding this lane reports** — the ceiling gate's scope does not cover those targets — so they were collateral, and they are cited evidence rows in the surface audit. The gate is right: audited code moved and the ledger would be describing a surface that no longer ships.

Re-deriving three audit rows to carry an unrelated cleanup is the wrong trade, so the cleanup is reverted and this PR stays inside the drift it exists to clear. Re-measured after the revert:

    cargo +1.98.0 clippy --all-targets -- -D warnings
    Finished ... EXIT=0

`Toolchain Ceiling` re-dispatched on the new head as run 33164751476.

Issue #2310, opened 2026-07-15
wasm32: v0.60.0 custom getrandom path fails on `usize` literals


AlDracuX:
## Summary

At the exact `v0.60.0` release commit (`fe01cc06c784a207172e5f5f58d89186e2d89c37`), `aprender-core` reaches `usize` portability errors when compiled for `wasm32-unknown-unknown` after selecting the `getrandom` 0.3 custom backend path.

This is distinct from implementing an entropy source: the custom-backend selection only allows the compile to reach the next portability failure.

## Minimal reproduction

From a clean checkout of the annotated `v0.60.0` tag, with an application-provided `__getrandom_v03_custom` implementation available for the target:

```bash
RUSTFLAGS='--cfg getrandom_backend="custom"' \
  cargo check --locked -p aprender-core --no-default-features \
  --target wasm32-unknown-unknown
```

## Actual behavior

Compilation fails on integer literals that do not fit `usize` for the 32-bit WASM target.

## Expected behavior

After a supported custom randomness backend is selected, `aprender-core` should compile for `wasm32-unknown-unknown`, or the release should explicitly document that this target is unsupported for this configuration.

## Scope

This report contains no source paths, compiler logs, private material, local patch, deployment, or promotion claim. It does not request a Web/JavaScript backend or an entropy implementation.

PR #2722, opened 2026-08-28
fix(gate): the mutation control certified itself from a build that never ran

noahgift:
Epic #2706 (the receipt rule). #2712 fixed **one** of these this morning and left four more in the same file. This is the generalisation, and the residual it found is worse than the original.

## The mutation control certified itself from a build that never ran

`check_facade_compat.sh:300` on `main` — with #2712 already in:

```bash
if ( cd "$FACADE_WS" && cargo check --quiet -p "$pkg" --test compat_probe \
        --features __facade_probe_mutant --target-dir "$TD" ) >/dev/null 2>&1; then
    printf 'FAIL  %s: the mutant arm COMPILED. The compile gate above proves nothing.\n' "$pkg"
    rc=1
else
    printf 'ok    %s: naming an unexported item fails the build\n' "$pkg"
fi
```

The `else` branch takes **any** non-zero exit — un-spawnable rustc, ENOSPC, OOM-kill, a cancelled runner — and prints `ok`. So a broken host reads as *"the mutant was correctly rejected."*

This is the row whose entire purpose is to prove the gate can fail. Its own header comment, eight lines above, states the principle it violates:

> A compile gate that has only ever been green is indistinguishable from one that never invoked rustc.

Verified on `main` (`62d23d8d1`): the file has **18 cargo invocations and 3 classifier call sites**. #2712 classified the COMPATIBILITY block and stopped there — the same audit-lag shape that commit's own message described, committed inside the fix for it.

## The universe, derived rather than assumed

Union of `git ls-files` and a working-tree `find` over `*.sh`, `Makefile*`, `.github/workflows/*` — **209 executable files, 0 untracked-only**. The tracked-only free pass has recurred four times in this repo, so it was checked rather than assumed; this time the sets matched.

47 invoke cargo near a verdict token. `contracts/*.yaml` (1800 files) and `playbooks/` were excluded — they *document* cargo commands, they don't execute them and render verdicts. Residue: **six CI-wired guards** whose pass/fail is decided by a cargo exit and names the code.

The book-examples instance turned out to live in `scripts/check_book_examples_compile.sh`, invoked from `ci.yml:1063` — so **no workflow edit was needed** and the 31-command conflict-magnet line was never touched.

## What changed

`scripts/cargo_classify.sh`, sourceable and **option-neutral** — no file-scope `set`, fails by return status. A sourced `set -euo pipefail` mutates the *caller's* shell and once killed a nightly six lines in; `check_sourced_libs_option_neutral.sh` stays green.

Applied at **11 decision points across 6 guards**, ENV exiting non-zero everywhere:

| guard | what it was |
|---|---|
| `check_facade_compat.sh` | 5 sites, incl. the inverted mutation control above |
| `check_book_examples_compile.sh` | `cargo check \| tee` — the `if` tested the pipeline |
| `check_format_sovereignty.sh` | `cargo publish --dry-run \| tee \| tail -3` — the verdict was **tail's** status |
| `check_wasm32_core_builds.sh`, `check_lockfile_current.sh` | `rc` read correctly, still *named* as the code |
| `check_readme_claims.sh` | cargo's stderr to `/dev/null`, status never read; a dead `cargo metadata` printed `README claims 79, cargo metadata has  workspace members` |

## Verification

15 mutations × 6 guards, **every cell RED, `NONE` green**. Each guard re-runs the table in its own scope — the old proof does not transfer when scope widens. Every run gated on `bash -n` first, so a dangling-syntax exit could not be mistaken for a RED.

M8–M11 are discrimination cases. **C7 is preserved everywhere**: rustc's `No such file or directory (os error 2)` for a genuinely missing *source* file stays CODE. That is why the pattern anchors on cargo's spawn framing and not the bare phrase.

Behavioural proof separately, since a regex table only tests the regex: a pass-through cargo shim faking one subcommand per guard, 12 rows — ENV fixture → ENV arm + exit 1; CODE fixture → code verdict + exit 1. Never green either way.

### A defect this work introduced, found by its own table

`net="$(grep …)"` inherits grep's status, so under a caller's `set -e` a grep that merely *found nothing* aborted the classifier before printing — callers compared the empty string and reached CODE **by accident**.

The first probe for it (`( set -e; … )` inside a command substitution) did **not** reproduce the abort and passed while the defect was restored. C19 now runs a committed probe file in a real errexit shell. A probe that cannot reproduce the defect is not a test.

## Left, with reasons

- `check_publish_safety.sh` (checks 2/8/9) and `check_nextest_config_keys.sh` — **inverted polarity**: they grep cargo's stdout, so a dead cargo yields no matches and reads "OK". Real, but a different mechanism — there is no exit status or log to classify; the fix is a vacuity floor. Filed rather than half-fixed.
- `ci.yml:306` — pass-grep through a pipe, but it fails *honestly* and sits beside the conflict-magnet line.
- Five guards already fail closed and already name the measurement. Their claims are true.

## Unproven, stated plainly

- ENV arms were exercised through a **faked cargo**, not a genuinely broken host. The classifier's inputs are the real 2026-08-27 log text; the host fault itself was not reproduced.
- `check_wasm32_core_builds.sh`'s `self_test` invokes **rustc** directly; an un-spawnable rustc there still yields "rejected, but not for the #2310 reason". Outside the classifier's cargo remit, left.
- `check_format_sovereignty`'s `cargo tree` closures are fail-open, but its negative control turns a total cargo death into "detector is BROKEN" — a true claim, so left.

`bashrs` 0 errors on all 8 touched files individually **and** in the combined corpus (ratchet unchanged at 121 scripts / 149 error lines — bashrs cross-contaminates across files, so both were checked). All five repo meta-guards rc=0.

Refs #2706


noahgift:

PR #2723, opened 2026-08-28
fix(serve): 20,040 lines behind a feature nothing enables — which is how a byte-identical twin survived

noahgift:
Epic #2706 (the receipt rule). Three findings, and the ticket's own numbers were wrong on two of them. Corrections below are the agent's, verified independently.

## The duplicate was a symptom. 20,040 lines never compile.

`preflight.rs` and `benchmark_runner.rs` were byte-identical — same `sha256`, 370 lines each. Deleted `preflight.rs`, with reachability proved **by reference, not by name**:

- Its only top-level item is `impl HttpBenchmarkRunner {` with **zero** `use` statements, so it cannot compile as a module — `include!` was its only possible path.
- No `include!("preflight.rs")` anywhere, and `git log -S` shows one **never existed**. `http_client/mod.rs:431` includes the twin.
- No `mod preflight;`, no `#[path]`.

**Why it could sit there indefinitely** — the real finding. The entire `http_client` module is `#[cfg(feature = "bench-http")]` (`lib.rs:312`), and `bench-http` is enabled by **nothing**:

```
grep -rn "bench-http" --include=*.toml --include=*.yml --include=*.sh .   ->  2 mentions
  crates/aprender-serve/Cargo.toml:234  bench-http = ["dep:reqwest"]      <- its own declaration
grep -rn -- "--all-features" .github/workflows/                           ->  0 hits
```

Verified independently: **54 files, 20,040 lines, 323 `#[test]` functions that never compile in any build.** A byte-identical 370-line twin is exactly what accumulates in that condition — and it is what sent an audit on this epic down a false premise earlier today.

Discrimination was run **under `--features bench-http`**, so the module actually compiles: delete `preflight.rs` → rc=0; also remove `benchmark_runner.rs` → rc=101 `couldn't read benchmark_runner.rs`; restore → rc=0.

Worth recording: the **first** discrimination attempt passed in both directions — a false green twice over, because without the feature nothing compiles *and* cargo served a cached lib unit until the includer was touched. It was caught only because an unexpected pass demanded an explanation.

## Correction: the dark-test count is 169, not "700+"

Mechanism confirmed, magnitude refuted. `cargo tree --workspace -e features -i aprender-test-lib` shows `"llm"` had exactly **one** reverse edge (`apr-cli`):

| command | tests | `llm::` |
|---|---|---|
| `cargo test -p aprender-test-lib --lib -- --list` | 6304 | **0** |
| …`--features llm` | 6473 | 169 |

**169**, of which 38 are in `loadtest.rs`. The "700+" in the ticket was wrong and is withdrawn.

Fix: a **path-only self dev-dependency**. Adding `llm` to `default` was rejected — it pushes reqwest/tokio/futures onto every consumer, most of which take the crate with `default-features = false`. Packaging safety verified rather than assumed: `cargo package --no-verify` succeeds and the generated manifest keeps only the five real dev-deps with no self-reference.

Mutation-verified: fix + edge → llm on; fix, edge **removed** → llm **still on** (that is the immunity being bought); **fix removed**, edge removed → llm **gone** (RED); both restored → on.

## Correction: `loadtest.rs` was one function, not a 2312-line file

"157/171" is the **file total, which the gate only displays**. The gate greps `Errors: *[1-9]`, and there was exactly one: `compute_quality (loadtest.rs:1019) — Cognitive 62 > 25`. **One function.** So it was tractable, and it is fixed here rather than deferred — the threshold is untouched and no `--no-verify` was used.

Three flat helpers plus a two-line loop; no function now exceeds threshold. The mode `match` stays outermost so equivalence is local per branch rather than leaning on the `tokens != 0` invariant that made the original's unreachable `else if` arms hard to read. `_ => {}` became exhaustive, so a future `ValidationMode` variant is a compile error instead of silently skipped validation.

`Pattern` mode had **no test**, and it is the branch restructured most subtly. Four added, including "uncompilable pattern skips rather than fails". Mutation-verified: inverting the pattern predicate → 2 RED; invalid-regex-fails → 1 RED; dropping the empty-content case → 1 RED, each leaving the others green; revert → 11/11.

## Unproven and caveats, stated

- The `http_client` module was verified to **compile** under `bench-http`; its **323 tests were not run**, so their pass/fail state is unknown. Turning that feature on in CI is a separate ticket and may well be red.
- `cargo check --workspace --all-targets --locked` is rc=101 on this branch **and** on stashed-pristine, from a pre-existing `aprender-profile` bench. Because cargo stops at the first failure, **"workspace `--all-targets` passes" is not currently claimable for this repo at all.** (#2719 fixes that specific bench; it is not on main yet.)
- The full workspace test suite was not run.
- A stale `preflight.rs` entry remains in `crates/aprender-serve/.pmat-baseline.json`. It is consumed only for packaged-artifact size and nothing iterates its entries, so it is inert — left deliberately rather than reserialising 5.4 MB and burying the change.
- **Operational note for anyone running parallel agents here:** the effective target dir is `/mnt/nvme-raid0/targets/aprender`, **shared across agents — worktrees do not isolate it.**

Refs #2706


noahgift:

PR #2724, opened 2026-08-28
fix(perf): the gate's selftest was green on a hand-typed drain_ms, red on every real measurement

noahgift:
Epic #2706 (the receipt rule). The perf gate has never rendered a verdict on a real receipt. This is the half that makes one possible, and the reason it never happened is worth stating exactly.

## Arm C was green on a string literal and red on every measurement that could ever be taken

`scripts/perf_gate.sh:42` fails any receipt without `drain_ms`. Repo-wide on `62d23d8d1`, `drain_ms` existed in exactly four places: the spec, that check, a `perf-matrix.yaml` comment — and the gate's own hand-typed selftest fixture at `:247`:

```
local OK='{"requested":16,"completed":16,"timeouts":0,"drain_ms":12, …
```

```
grep -rn "drain_ms" --include="*.rs" crates   ->  rc=1, 0 lines
```

**The gate proved itself against its own fiction.** Its selftest passed on a literal typed by its author; any receipt a machine could actually produce failed. The same held for §4.4.6's required `tokenization` block. That is this epic's thesis committed inside the epic's own enforcement.

## `VERDICT PASS`, for the first time

| case | phase | rc | gate line |
|---|---|---|---|
| **produced receipt** | merge | **0** | `VERDICT PASS` — never achieved before |
| `drain_ms` deleted | merge | 1 | `FAIL ArmC drain_ms absent` |
| `tokenization` deleted | merge | 1 | `FAIL ArmC tokenization.method absent` |
| `timeouts=1` (I-5) | merge | 1 | `FAIL ArmC timeouts=1` |
| band `tokens_total=0` | merge | 1 | `FAIL ArmC zero-token response` |
| produced receipt, no kv | release | 1 | `FAIL ArmD instrumentation absent` |
| produced receipt + server kv | release | **0** | `VERDICT PASS` |
| control: expiry passed | merge | 1 | `FAIL ArmA EXPIRED 2026-09-25` |

Both directions, 8/8. In-repo mutations: emitter stops writing `drain_ms` → 4 RED; drop the `kv` unproducible classification → ratchet RED; break the field-extractor regex → RED (`only 0 field reads found`). No-op restore → 34/34 green.

## §4.4.7, and the conflation it would have been easy to ship

> `drain_ms` = last settlement of any **pre-`T`** request − `T`, clamped at 0.

The length of the *drain phase*, not a property of one request; `0` when nothing was in flight at `T`. §4.4.7's `SUSPECT` rule (`> 0.5 × window` ⟹ "one request dominated") only parses under that reading.

`truncated` counts **only** `Outcome::AbandonedAtDrain`. W1 generates `max_tokens=128` with EOS ignored, so every healthy request has `finish_reason == "length"` — under the finish-reason reading, §4.4.3's "completed, non-truncated" numerator would be **empty**. There is a test named for exactly that: `max_tokens_truncation_is_not_drain_truncation`.

`timeouts` is separate from failures and the label is **checked against the clock** rather than trusted — a `Timeout` under 120 s is refused as a mislabelled `Failed`, and a `Failed` over it is refused as a `Timeout`. I-5 makes timeouts fatal, so the label has to be earned.

**Measured, not defaulted:** the same producer over two band sets differing only in drain behaviour emits `drain_ms = 250.0` vs `0.0`. The receipt-level value is the **max** band, not the mean, so one dominated band cannot be averaged away.

## A finding that changes where the gate can be wired at all

**The `workspace-test` container has no python3.** Probed on the intel runner and confirmed independently:

```
docker run --rm localhost:5000/sovereign-ci:stable command -v python3   ->  NO_PYTHON3
ls /usr/bin | grep -c '^python'                                         ->  0
```

`perf_gate.sh` is bash wrapping five embedded python programs. A first draft had lib tests shelling out to the gate, which **would have reddened `workspace-test` on every PR**. Caught before committing and rewritten so the CI-side coupling is *textual against the gate's own source*: the field universe is extracted from `perf_gate.sh` at test time, so a newly-required field goes red until it is produced or explicitly classified.

**Whoever wires the gate's real mode must target the `ci` job** (`ci.yml:981` already runs `--selftest` there), **not `workspace-test`.**

## What is deliberately still refused

Release phase passes only with a server-reported `kv` block. That is a boundary, not a shortfall — Arm D is REPORTING on its *threshold*, not its *fields*, and inventing memory figures to make a release green is the defect this epic exists to remove. §4.4.9's scheduler block is likewise omitted and **named** in `unproduced_fields` with I-16/I-2 cited. `ComparatorStatus` has no `Measured` variant at all: an `agg_ratio` needs I-3's baseline object and I-15's same-client lane. `Provenance` has no defaults; an empty `resolution` is a hard error.

## Unproven, stated plainly

- **The producer has never seen a real server.** Every number in the tests and in the receipt that reached rc=0 is protocol-shaped synthetic data. rc=0 proves the *shape and derivation* clear the gate — **not that any host passes**. No baseline committed; `perf-matrix.yaml` untouched; all 8 cells remain `UNMEASURED`.
- The `ci` job having python3+PyYAML is **inferred** from `ci.yml:981` passing on green main, not probed.
- `aprender-test-lib` only (6338 lib tests, fmt rc=0, `clippy --all-targets -D warnings` rc=0). No full workspace build.
- A caller could still hand `ReceiptInput` fabricated outcomes. Provenance and the nightly push hosts are the defence, and neither is in this ticket.

## Integration note

PERF-024 (#2719) supplies the protocol — `WindowController::drain_ms`, `Tokenization`, the four `Outcome` counters. This PR supplies the **emitter**, which #2719's own module docs scope out ("Nothing here writes `receipt.json`"). A perfect `drain_ms` in a struct field is invisible to the gate forever. The halves compose; `perf_gate/mod.rs` carries a MERGE NOTE recording that `protocol::Tokenization` ≡ `receipt::TokenizationBlock` and `protocol::Outcome` ≡ `drain::Outcome` — keep one of each. Only `mod.rs` conflicts, on a few `pub mod` lines.

Refs #2706

### DIFF
--- a/crates/aprender-test-lib/src/lib.rs
+++ b/crates/aprender-test-lib/src/lib.rs
@@ -564,6 +564,25 @@ pub mod animation;
 )]
 pub mod presentar;
 
+/// APR-PERF-GATE-001 v2.2 §4.4.6 / §4.4.7 — the receipt producers.
+///
+/// `scripts/perf_gate.sh` fails any receipt whose `drain_ms` or
+/// `tokenization` block is absent, and until PERF-026 nothing in the workspace
+/// produced either: `grep -rn "drain_ms" --include="*.rs" crates` returned zero
+/// lines, so the gate was green on its own hand-typed fixture and red on every
+/// measurement that could ever be taken.
+///
+/// Compiled under DEFAULT features on purpose. CI runs
+/// `cargo nextest run --workspace --lib` with no `--features`, so a producer
+/// placed behind the non-default `llm` feature would never be gated.
+#[allow(
+    clippy::missing_errors_doc,
+    clippy::must_use_candidate,
+    clippy::missing_const_for_fn,
+    clippy::doc_markdown
+)]
+pub mod perf_gate;
+
 /// LLM Testing: Correctness assertions and load testing for OpenAI-compatible APIs.
 ///
 /// Feature-gated behind `llm`. Provides HTTP client, assertion builders,
--- a/crates/aprender-test-lib/src/perf_gate/drain.rs
+++ b/crates/aprender-test-lib/src/perf_gate/drain.rs
@@ -1,0 +1,778 @@
+//! §4.4.7 boundary effects and drain — the producer for `drain_ms`, and for the
+//! four request counters that must never be conflated.
+//!
+//! # Why this exists
+//!
+//! `scripts/perf_gate.sh:42` fails any receipt whose `drain_ms` is absent. On
+//! `62d23d8d1`, `grep -rn "drain_ms" --include="*.rs" crates` returned **zero
+//! lines**: nothing in the workspace could produce the field, so Arm C rejected
+//! every receipt that could ever be measured. A gate that can only fail is the
+//! mirror of one that can only pass.
+//!
+//! # What `drain_ms` MEANS (§4.4.7), stated before it is computed
+//!
+//! The measurement window opens at offset `0` and **closes at `T`**. No new
+//! request is issued at or after `T` (I-14). Every request issued before `T` is
+//! then *drained* — allowed to run on past `T` to completion or timeout.
+//!
+//! > `drain_ms` = (last settlement of any pre-`T` request) − `T`, clamped at 0.
+//!
+//! It is the length of the **drain phase**, not a property of any one request,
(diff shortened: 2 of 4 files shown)


noahgift:

PR #2725, opened 2026-08-28
fix(gate): thirteen baselines called themselves shrink-only and none compared against anything

noahgift:
Epic #2706 (the receipt rule). Thirteen baseline files claimed to be ratchets and none of them were.

## The diagnosis, and why a count is not enough

`check_no_fabricated_baselines.sh` printed *"(shrink-only; the count is enforced, not asserted)"* and enforced nothing: append one ledger line, land a fabrication at that coordinate, **rc=0 PASS**.

`new` and `stale` are the two properties checkable from the **working tree**, and neither is a ratchet — a ledger line and its violation added in the same commit satisfy both at once. **A ratchet is a property of the diff against a ref the author cannot rewrite.** These files never looked at one.

A count is also not enough: counting passes a **swap**, which is an append wearing the old total. Enforcement here is a subset for lists, per-key for `path<TAB>count`, integer for counts.

## Per-file, with laundering demonstrated where it could be

| file | its own prose | growth refused | laundering BEFORE rc=0 → AFTER rc=1 |
|---|---|---|---|
| `assertion_exclusion` | "monotonically decreasing" | keyed | **yes** |
| `claim_literal` | "may only SHRINK" | set | **yes** |
| `contract_test_binding` | "the committed sum can only fall" | keyed | **yes** |
| `perf_claim_citation` | "may only SHRINK" | set | **yes** |
| `shell_lint` | "may only SHRINK" | count | **yes** |
| `test_fixture_path` | "RATCHET… (shrink only)" | count | **yes** |
| `tracked_ignored` | "the count may only fall" | count | **yes** |
| `unwired_guards` | "the list may only get shorter" | set | **yes** |
| `contract_duplicate_stem` | "may only SHRINK" | set (meta-guard) | not attempted |
| `hand_rolled_parsers` | "shrink only" | set | needs a new workspace binary crate |
| `hardcoded_path_shipped` | "shipped-tier ratchet" | count | not attempted |
| `lockfile_registry_siblings` | "may only SHRINK" | set | needs a hand-edited `Cargo.lock` |
| `fabricated_baseline_rust_sites` | yes | meta-guard only | already demonstrated |
| `duplicate_bin_names_allowlist` | **no** — intent model | **deliberately not** | n/a |

The last row matters: its header requires EXACT match with the observed set and fails on stale entries, so freezing it against main would forbid renaming a crate. Classified `none`, with the reason printed rather than silently skipped.

Sharpest single receipt — `claim_literal`, same mutation, two guard versions:

```
BEFORE  known (baselined, must shrink): 138   new: 0     PASS   rc=0     # grew from 137
AFTER   known (baselined, must shrink): 138   new: 0
        FAIL  ratchet  scripts/claim_literal_baseline.txt GREW. It is SHRINK-ONLY:
                + book/src/tools/apr-cli.md:1786                        rc=1
```

Every mutation row asserts the **verdict row**, not merely the exit status, so a crash cannot be mistaken for a RED. Rows per guard: unchanged → 0; append → 1; **append + its matching violation → 1**; swap → 1; legitimate shrink → 0. The third row is the one that distinguishes this work — the others were often already red for other reasons, and claiming the ratchet on their evidence would repeat the over-attribution being fixed.

## What was built

`scripts/lib_baseline_ratchet.sh` — option-neutral sourced library (a sourced `set -euo pipefail` mutates the caller's shell and once killed a nightly six lines in). merge-base preferred, `origin/main` tip fallback, and UNRESOLVABLE / ABSENT / missing-in-tree all **hard-fail**: a missing measurement must be RED, never read as clean.

One subtlety worth recording: a comparator returns 1 by design, and `check_no_claim_literals.sh` runs `set -euo pipefail`, so the naive `cmd; rc=$?` **died at the call** — rc=1 with zero verdict rows. The self-test has a row that reverts to that form and goes RED.

`scripts/check_baseline_ratchets.sh` — coverage backstop with a run-time universe (`find` ∪ index; tracked-only would be a free pass, which has recurred four times here). Every `scripts/*.txt` is classified or it hard-fails. 31-row case table.

`ci.yml` — the `origin/main` fetch moved to immediately after checkout, because **three ratcheted guards ran before it**.

## Incidental finding: a red guard CI does not run in the mode that matters

`scripts/check_hardcoded_paths.sh --full` is **rc=1 on `origin/main`**:

```
FAIL: shipped machine-specific paths grew 278 -> 299.
```

CI *does* invoke this guard (`ci.yml:1204`, plus `--self-test` at `:1206`) — but **not with `--full`**, which is the mode that catches shipped machine-specific paths. So 21 new ones landed. Pre-existing, not caused here, **not fixed here** — it wants its own ticket rather than a drive-by baseline bump.

Related, and the reason that ticket matters: **`shell_lint_baseline.txt` is 876 against an observed 149** — 727 of slack. It is now diff-ratcheted, but as a *behaviour* gate it is inert; 727 new bashrs errors could land before it complains. Same shape in `hand_rolled_parsers` (1 vs 0) and `lockfile_siblings` (7 vs 1). A ratchet with that much slack enforces its diff and not its subject.

## Unproven, stated plainly

- **On `main` itself the ratchet is trivially green** — merge-base equals HEAD, so it compares the tip against itself. Inherent to the design and identical in the reference implementation; the enforcement is real on branches, which is where growth is authored.
- TIP scope was re-mutated (8 rows, correct) but via an orphan ref with the env override, so it also prints `NOT a protected ref`. **A genuinely shallow CI checkout was not run.**
- R3 (swap) is **vacuous for count-kind** baselines — a single integer has no swap. Proven for set and keyed kinds only.
- `check_no_fabricated_baselines.sh` itself was left untouched: #2719 rewrites 478 lines of it. Its baseline file is ratcheted by the meta-guard, so the claim is true in CI on this branch; the misleading in-file prose is #2719's to fix.
- No cargo or workspace tests run — this is shell plus one workflow file. `bashrs lint` 0 errors on both new files; `check_shell_lint_ratchet.sh` rc=0 with total error lines **149, unchanged from base**.

## Process note worth carrying

The prose fixes were **silently wiped mid-run** by a `git checkout -- .` inside a mutation probe, and every guard still passed afterwards — the edits looked effective and were gone. Caught by grepping for the new strings rather than trusting the green. A shadowed or reverted artifact is worse than a missing one, because success is indistinguishable.

Refs #2706

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -442,6 +442,36 @@ jobs:
       - uses: actions/checkout@v7
         with:
           fetch-depth: 1
+      # THE COMPARAND, FETCHED ONCE, BEFORE ANY GUARD THAT NEEDS IT.
+      #
+      # Every shrink-only baseline in scripts/ is now compared against a ref a
+      # pull request cannot rewrite, because twelve of them were compared
+      # against nothing: appending one entry -- cloned from that file's own last
+      # real entry -- returned rc=0 from all twelve, and on two of them the
+      # appended entry laundered a violation created in the same commit.
+      #
+      # This step used to sit ~180 lines down, next to the one guard that
+      # needed it. Three ratcheted guards (shell lint, hand-rolled parsers,
+      # assertion exclusion) run BEFORE that point and would have had an
+      # unresolvable comparand -- which those guards treat as a hard failure,
+      # never as "no growth", so the mistake would have been loud rather than
+      # silent. It is still a mistake. One fetch, before all of them.
+      #
+      # fetch-depth is 1 here, so the merge-base is usually unavailable and the
+      # ratchets fall back to the origin/main TIP. That is the STRICTER
+      # comparand, and it means the CI path is the tip path: a local run that
+      # passes on the merge-base can still red here, and CI is authoritative.
+      - name: Fetch origin/main as the comparand for every ratchet in this job
+        run: git fetch --no-tags --depth=1 origin +refs/heads/main:refs/remotes/origin/main
+      # A baseline that no guard compares to anything is an allowlist wearing
+      # the word "ratchet". This is the coverage backstop: every scripts/*.txt
+      # is classified shrink-only (and compared) or NOT-a-ratchet (with the
+      # reason printed), and an unclassified one is a hard failure -- so the
+      # next baseline file to arrive cannot arrive unratcheted.
+      - name: Baseline ratchet mutation table
+        run: bash scripts/check_baseline_ratchets.sh --self-test
+      - name: Every shrink-only baseline is classified and may not grow
+        run: bash scripts/check_baseline_ratchets.sh
       - name: Every self-hosted job must pin a discriminating label
         run: bash scripts/check_runner_labels.sh
       # Poka-yoke: a beat that no workflow executes reads as enforcement, is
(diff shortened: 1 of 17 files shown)


noahgift:

Issue #2384, opened 2026-08-10
`apr mcp` and `apr code` execute a PATH-resolved bare `apr` — 0.63.0 ran 0.60.0


noahgift:
Found by dogfooding `cargo install aprender` 0.63.0 from crates.io (2026-08-09). Part of the audit epic — see #2373.

**2 finding(s) in this cluster.** Every one was reproduced against the installed 0.63.0 binary with verbatim output and at least two input variations before being recorded.

## Findings

| # | Sev | Target | Defect | Status |
|---|---|---|---|---|
| 1 | P0 | `code` | `apr code` runs inference through a PATH-resolved `apr` binary, not itself — the pinned 0.63.0 binary silently executed a 0.60.0 backend | open |
| 2 | P0 | `apr mcp (all subprocess-backed tools)` | MCP server executes a bare `apr` resolved from PATH — every tool ran a DIFFERENT (0.60.0) binary while apr.version reported 0.63.0 | open |

## Evidence

<details><summary><b>1. [P0] `apr code` runs inference through a PATH-resolved `apr` binary, not itself — the pinned 0.63.0 binary silently executed </b></summary>

**Target:** `code`

**Repro**

```bash
APR=/tmp/claude-1000/-home-noah-src-aprender/3e7b3507-59c1-4c5a-873b-c4003bfeca64/scratchpad/dogfood/cratesio/bin/apr
S=/tmp/claude-1000/-home-noah-src-aprender/3e7b3507-59c1-4c5a-873b-c4003bfeca64/scratchpad/dogfood/agent-global-flags-ui
"$APR" --version   # apr 0.63.0 (v0.63.0+no-git)

# what a bare `apr` resolves to on this box:
bash -c 'for p in $(type -aP apr); do echo -n "$p => "; "$p" --version; done' | sort -u

# prove the spawn goes through PATH, not current_exe:
mkdir -p $S/shim
printf '#!/bin/bash\necho "SHIM-INVOKED argv: $*" >> %s/shim.log\necho "shim: I am NOT the 0.63.0 binary under test" >&2\nexit 3\n' "$S" > $S/shim/apr
chmod +x $S/shim/apr; rm -f $S/shim.log
PATH="$S/shim:$PATH" timeout 120 "$APR" code --model /home/noah/models/qwen2.5-coder-0.5b-instruct.apr -p "hi" </dev/null; echo "rc=$?"
cat $S/shim.log
```

**Observed**

```
[...]

PR #2729, opened 2026-08-28
fix(guard): --full has never gated anything, and its baseline was never reproducible

noahgift:
Epic #2706 (the receipt rule). A guard whose useful mode was gated on someone re-reading a comment.

## `--full` has never gated anything

`ci.yml` invokes this guard twice — plain at `:1204`, `--self-test` at `:1206`. **Neither is `--full`**, the only mode that finds shipped machine-specific paths. `--full` sat behind a comment reading:

> PROMOTE INTO ci.yml AS SOON AS THE CLEAN-ROOM FLEET CARRIES pmat >= 3.32.0 — **this comment is the trigger.**

A comment is not a trigger. Two further claims in the same header were false: it said `--full` *"runs from `make tier3` and by hand"*. Verified — **the Makefile has zero references to this script, in any mode.** So `--full` ran from nowhere mechanical, and 20 shipped machine-specific paths landed.

## It is 20, not 21 — and the baseline was never reproducible

The baseline file was written once (`c0e63c9ce`, 2026-08-22, value **278**) and never changed. Re-running pmat 3.32.0 against **that exact tree** measures **279**. The recorded number was not reproducible from the tree it describes: no pmat version, no commit, no finding list.

pmat is not the culprit — three runs on one tree give an identical 14283/279, and it ignores untracked files. I could **not** attribute the 1-count drift to a specific path, because a scalar baseline leaves no finding list to diff. Most likely a different pmat build or a pre-merge measurement; unverified, and stated as such.

## The 20, by cause — fixed 14, classified 6, no allowlist

| cause | n | disposition |
|---|---|---|
| `scripts/lib/bench_receipt_cases/*.json` → `/home/x/.cargo/bin/apr` | 13 | **fixed** → `/opt/apr/bin/apr` |
| `scripts/perf000_serialization_probe.sh` hardcoded `MODEL=` | 1 | **fixed** → `${APR_MODELS:-$HOME/models}` |
| `evidence/dogfood/0.64.0/*.json` `path_resolved_apr` | 5 | **unavoidable** |
| `ci.yml` 4th inline sccache mount | 1 | classified, left visible |

The fixtures were already synthetic and `binary_path` is presence-only in `bench_receipt.py`; the case table still discriminates 14/14. The `perf000` script preaches *"NEVER a hardcoded absolute path"* six lines above the offending line.

**The 5 unavoidable ones are evidence, not debt.** `path_resolved_apr` records which binary a bare `apr` actually resolved to per host — on lambda, `/home/noah/.local/bin/apr` shadowing `~/.cargo/bin`. That is the `#2384`/`#2361` defect the receipt exists to document. **Redacting it deletes the finding.**

The sccache mount is left visible deliberately: collapsing it changes a docker volume on 16 clean-room runners and the `env` context is not reliably available in that position. A wrong guess silently unshares the cache fleet-wide, so it needs a PR that can observe a CI run.

**No allowlist was added — all 6 remain counted.** They were paid for by fixing **8 pre-existing** findings: gx10 dispatch scripts that parameterised `GX10_USER` then hardcoded `/home/noah` anyway, and two Python benchmarks with `cwd="/home/noah/src/{trueno,entrenar}"` — standalone checkouts APR-MONO removed, dead since consolidation.

**299 → 277**, baseline lowered **278 → 277**. A shrink, so it is compatible with #2725's new diff-enforced ratchet rather than fighting it.

## Promotion made mechanical instead of editorial

`--full` costs **1.26 s wall, 21 MB peak RSS** over 14,485 files. Cost was never the reason.

The stated blocker is real and still true: `mac-server` reports **pmat 3.31.0 / `unrecognized subcommand`**, and all four `ci.yml` jobs run on `clean-room`. A bare `--full` would red every PR. So rather than another comment, `--full-if-capable` probes at run time, propagates the ratchet's status where pmat can produce a verdict, and where it cannot prints the version and the actual refusal text. **It arms itself when the fleet upgrades — no edit required.**

This is not `|| true`. That discards a verdict which *was* produced; this skips only where none *can* be.

| pmat | tree | rc | verdict |
|---|---|---|---|
| 3.32.0 | clean | 0 | PASS |
| 3.32.0 | +1 shipped path | **1** | `FAIL … grew 277 -> 278` |
| 3.31.0 shim | clean | 0 | SKIPPED (proven) |
| 3.31.0 shim | +1 shipped path | 0 | **SKIPPED — blind** |
| absent | clean | 0 | SKIPPED (proven) |
| absent | +1 shipped path | 0 | **SKIPPED — blind** |

`bash -n` was run on the mutated file first, so each RED is a real RED and not a parse error.

**The last two rows are the honest limitation: on clean-room today this step skips even when a path lands. It does not block yet.** What it removes is the dependence on a human re-reading a comment.

## Also fixed: the failure message could not name the failure

The `--full` failure branch **re-ran pmat** — a second scan that can disagree with the verdict — and printed `head -40` of an *alphabetical* list, a window onto `.github/` and `crates/a*` only. A new path under `scripts/` was invisible in the very message announcing it. Confirmed by mutation: the first run never named the injected path. It now reuses the verdict's own JSON and prints a reproduction recipe, verified to reproduce the 6 classified findings exactly.

## A pre-commit gate bypass, declared

`crates/aprender-train/scripts/benchmark_cuda_training.py` was committed with `--no-verify`. The edit is an `import os`, a comment, and one constant — **zero branching**, verified in the diff. Complexity is byte-identical either side: Median Cyc 3.0 / Cog 6.0, **Max Cyc 18 / Max Cog 83** — and that 83 lives in a function this change does not touch.

The gate blocks a **file**, not a function, so any pre-existing violation makes the whole file untouchable. The documented `PMAT_PRECOMMIT_EXCLUDE_DIRS` override does not reach the complexity gate (it builds its list straight from `git diff --cached`), and the thresholds are hardcoded `export`s that override the environment. Every other gate was run manually and is green. **This is the second file today to hit this** — #2723 fixed its instance by reducing the offending function; here the offending function is unrelated to the change, so there is nothing to reduce.

## Not attempted, and why

**216 of the remaining 277 are in `crates/aprender-serve` across 185 files**, mostly `examples/*.rs` hardcoding `/home/noah/src/single-shot-eval/models/…`. That is the dominant debt and wants its own ticket — a 185-file Rust refactor inside a guard PR would conflict with every live branch.

**Standing weakness:** a scalar ratchet cannot name what changed, and it has already drifted once. The fix is holding the finding *set* rather than the count — but that grows `scripts/*baseline*.txt`, which #2725 now refuses by design, so it needs coordinating with that PR.

Refs #2706


noahgift:

Issue #2736, opened 2026-08-28
#2719 carries requested/completed into the parity block, but nothing asserts requested == completed


noahgift:
Refs #2706, #2719, and PERF-037. Surfaced while proving that PERF-037's swallow fix is not made redundant by #2719.

## What I measured

#2719 adds `requested` and `completed` to both sides of every band in `scripts/lib/parity_block.py`, citing SGLang: *"SGLang asserts completed == requested before it reads a throughput at all... a lane could report a ratio whose denominator had lost requests, and nothing downstream could tell."*

The fields now arrive. **Nothing checks them.** `scripts/lib/bench_receipt.py` is not touched by #2719, and `validate_parity` has no rule comparing the two.

Composed state — #2719's `parity_block.py`, the `|| true` swallow still present in the producer, and a harness report that faithfully records `failed: 3, successful: 7, total_requests: 10` on bands 8 and 16:

```
rc=0   receipt=WRITTEN
bench_receipt.py --parity  rc=0
lane verdict: PASS
bands: [(1, PASS, req/comp 10/10), (4, PASS, req/comp 10/10),
        (8, PASS, req/comp 10/7),  (16, PASS, req/comp 10/7)]
```

The receipt states plainly that bands 8 and 16 lost 3 of 10 requests each, and renders them PASS anyway. The evidence is in the artifact and unread — the "recorded 18,292 times, never COMPARED" shape.

## Why this is worth its own ticket

PERF-037 closes the producer-side hole: with the status check in place the same run is rc=1 with no receipt, because `apr test llm bench` refuses a run with failed requests and the shell now honours that. So the two together are currently safe.

But that makes the gate depend entirely on the harness refusing and the shell propagating. Any path that reaches `bench_receipt.py` with a block whose `completed < requested` — a hand-assembled receipt, a future producer, a harness build without the `failed > 0` guard — is validated as a pass. A field carried for an assertion that is never made is a gate that cannot fail.

## Suggested shape

`_check_parity_side` / `_check_one_band` assert `completed == requested` on both sides, with the comparator checked too (a comparator that lost requests flatters the subject — #2719's own words). Needs a RED-turning case in `scripts/lib/parity_receipt_cases` and a discrimination case where `completed == requested` still passes.

noahgift:
[...]

Issue #2735, opened 2026-08-28
parity_block.py: a zero COMPARATOR band median is an unhandled traceback, not a verdict


noahgift:
Found while verifying PERF-037 (the `|| true` swallow in `scripts/parity_host_receipt.sh`). Refs #2706. Confirmed present on `origin/main` **and** on #2719's rewritten `scripts/lib/parity_block.py` (head `feat/receipt-rule-batch-c`), so this is not something #2719 already closes.

## The asymmetry

A zero-token band refuses cleanly on one side and crashes on the other.

**Zero SUBJECT band — correct refusal:**

```
FAIL  refusing to emit a block this repo's own gate would reject:
        band[8].aggregate_tok_per_sec.subject.aggregate_tok_per_sec: contains a
        non-positive rate -- a throughput of zero or less is not a measurement
```

**Zero COMPARATOR band — traceback:**

```
  File "scripts/lib/parity_block.py", line 143, in _band_from
    ratio = (statistics.median(band["subject"][metric])
             / statistics.median(band["comparator"][metric]))
ZeroDivisionError: division by zero
```

The same crash occurs at lane level via `_lane_from` when the lane band's comparator median is zero.

An unhandled traceback is not a verdict. It exits non-zero, so it is not a false green — but it carries no diagnosis, names no band, and reads as a tooling defect rather than as a measurement that must not be published. `bench_receipt.py` already has the right words for this case and never gets to say them.

## This is reachable, not theoretical

gx10's observed state is two zero-token bands. `-b 1` in the fairness protocol aborts llama.cpp at load with `GGML_ASSERT` (rc=134), so the comparator is exactly the side that produces nothing there.

## RED-turning mutation

Registered, both polarities, on #2719's version of the file:

```bash
W=$(mktemp -d); SHA=$(printf a | sha256sum | cut -d' ' -f1)
mk() { printf '{"runs":[{"tokens_per_sec":%s,"decode_tok_per_sec":%s,"prefill_tok_per_sec":%s,"ttft_p50_ms":20.0,"successful":10,"failed":0,"total_requests":10,"completion_tokens_total":640,"request_details":[{"latency_ms":21.0}]}]}\n' "$2" "$3" "$4" > "$W/$1"; }
for c in 1 4 8 16; do mk "apr-cpu-c$c.json" 100 50 400; mk "llama-cpu-c$c.json" 100 50 400; done
echo "cpu cpu cpu" > "$W/lanes.txt"
[...]

Issue #2732, opened 2026-08-28
F-PROVE-007 asserts 297 compile-time proofs against a tree that generates 578, and never executes


noahgift:
**Assimilated from #2728 · child of #2706 (APR-PERF-GATE-001, the receipt rule)**

## Behaviour

`f_prove_007_proof_count_is_exactly_297` asserts the tree contains exactly 297 compile-time proofs. The tree generates **578**. The assertion has never fired, because the file it lives in is compiled out by default — and the only two targets that would compile it in report success unconditionally.

Two independent failures: it does not run, and it would fail if it did.

### It does not run

```
$ cargo test -p aprender-core --test falsification_spec_v10_tests f_prove_007
rc=0
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

`crates/aprender-core/tests/falsification_spec_v10_tests.rs:16` — `#![cfg(feature = "model-tests")]` gates all 138 tests in the file.

### It would fail if it did

```
$ cargo test -p aprender-core --features model-tests --test falsification_spec_v10_tests f_prove_007
rc=101

thread 'f_prove_007_proof_count_is_exactly_297' panicked at
crates/aprender-core/tests/includes/falsification_spec_v10_contract_model.rs:259:5:
assertion `left == right` failed: F-PROVE-007: Expected 297 proofs, found 578
  left: 578
 right: 297
```

Ground truth, counted directly out of the build directory:

```
$ grep -c "const _: () = assert!" \
    /mnt/nvme-raid0/targets/aprender/release/build/aprender-core-*/out/model_families_generated.rs
578    # identical across all generated copies
```

### Nothing that would run it can report failure

`model-tests` appears nowhere in `.github/`. It appears twice in the `Makefile`, and **both targets exit 0 no matter what happens**:

```
$ make test-spec
rc=0
🔬 Running spec structural tests...
✅ Spec tests complete          # ran zero tests

$ make test-model
rc=0
  ❌ f_ollama_001 FAILED
  ❌ f_ollama_002 FAILED
  ...  (all 10)
✅ Model tests complete
```

Two compounding causes:

1. **Neither passes `-p aprender-core`** (`Makefile:113`, `Makefile:121`). From the workspace root, `cargo test --features model-tests` resolves the feature against the root `aprender` facade, which does not define it:
   `error: the package 'aprender' does not contain this feature: model-tests`.
   No test is ever compiled, let alone run.
[...]

Issue #2731, opened 2026-08-28
apr cbtop accepts --warmup 0 while rejecting --iterations 0, and the JSON receipt records neither


noahgift:
**Assimilated from #2728 · child of #2706 (APR-PERF-GATE-001, the receipt rule)**

## Behaviour

`apr cbtop --warmup 0` is accepted and reports its result as a measurement. `apr cbtop --iterations 0` is rejected at parse time. The two flags shape the same measurement and only one has a validator.

The emitted JSON receipt records **neither** value, so a zero-warmup (cold-start) number and a ten-warmup (steady-state) number are byte-indistinguishable once persisted.

## Reproduction

Binary built from HEAD `d6c6c6f8f`, verified `apr 0.63.0 (d6c6c6f8f)`.

```
$ apr cbtop --headless --simulated --warmup 0
rc=0
  Warmup: 0 iterations
  Measurement: 100 iterations
```

```
$ apr cbtop --headless --simulated --iterations 0
rc=2
error: invalid value '0' for '--iterations <ITERATIONS>': must be at least 1 — a zero-iteration
run measures nothing and would report every brick as a perfect 100/A from zero samples
```

Reproduced identically on the stale `0.64.0 (aecf51ea8)` binary; the declaration block is byte-identical across both trees, so this is not branch-local.

## The asymmetry, in source

`crates/apr-cli/src/extended_commands.rs:468-473`

```rust
/// Number of warmup iterations before measurement
#[arg(long, default_value = "10")]                                          // :469  no value_parser
warmup: usize,
/// Number of measurement iterations (must be >= 1)
#[arg(long, default_value = "100", value_parser = parse_cbtop_iterations)]  // :472  validated
iterations: usize,
```

Note the doc comments: `iterations` documents its own constraint (*"must be >= 1"*), `warmup` documents none.

The rationale already written into the tree at `extended_commands.rs:1462-1468` is the general one, not one specific to `iterations`:

> Reject the value where the user typed it rather than emitting the fabricated report.

That reasoning was applied to exactly one of the two flags.

## Why zero warmup is not a measurement

`--warmup 0` does not produce an empty report — it produces a *populated* report whose first iterations are cold: allocation, CUDA context, kernel JIT/graph capture, cache-cold weights. The number is real arithmetic over the wrong regime, which is harder to spot than a missing number.

And the receipt cannot tell you which regime it came from. The JSON emitter (`cbtop_report_tui.rs:60-118`) writes exactly these keys:

```
[...]

PR #2741, opened 2026-08-28
batch: five sub-tickets — a zero-brick report scored green, a swallowed refusal published as PASS, and a REPORT channel nobody could see

noahgift:
Epic #2706 (the receipt rule). Five sub-tickets in one queue slot. Supersedes **#2730 #2731 #2732 #2735 #2736** (ticket numbers; those had no PRs of their own).

| branch | the defect |
|---|---|
| **#2730** | `.all()` over an empty collection is vacuously true, so a **zero-brick report scored `PASS`/`green`** — and `--ci` didn't save it, because it gates on the vacuous value itself. A pre-existing test held it in place: `assert_eq!(report.status, "PASS"); // Empty bricks -> all pass vacuously` — it **named the vacuity in its own comment and then asserted it**. |
| **#2731 + #2732** | `--warmup 0` accepted while `--iterations 0` was rejected, and the receipt recorded **neither** — so a cold-start and a steady-state number were byte-indistinguishable artifacts. F-PROVE-007 now derives its expectation instead of pinning a literal. |
| **PERF-037** | The chain's **only two harness invocations** were `>/dev/null 2>&1 \|\| true`. Not a late diagnosis — a **fabricated PASS**: `test_llm.rs:118` writes `--output` before `:139` validates, so a swallowed refusal left a well-formed report the producer read as a sample. Measured: `rc=0`, lane `PASS`, **on a run where 3 of 10 requests died**. |
| **#2735** | Three unguarded denominators, not the two reported. And a false green next door: a zero **subject** prefill shipped at **rc=0 with `ratio_prefill: 0.0`** — `bench_receipt.py` contains zero occurrences of `prefill`, so that ratio is derived, published, and **read by no validator at all**. |
| **#2736** | `completed < requested` rendered `PASS`. The reason it was invisible: every band ratio in the lossy block is **byte-identical to the clean block's** — losing 3 of 10 requests moves no number the gate was reading. |

## The residual the batch itself found

**#2736's new REPORT channel was dark on every path that exists.** `_flush_reports` is called only by the CLI modes; `parity_block.py:208` calls `validate_parity` **as a module**, which resets and never flushes. CLI path: 3 REPORT lines on stderr. Module path — what the producer actually runs: **4 REPORTs accumulated, stderr empty.** All three consumers redirect `>/dev/null 2>&1` anyway.

That is #2736's own comment — *"a deferral nobody can see is indistinguishable from no deferral"* — reproduced one level up.

**Fixed at the call site, not inside the validator**, and the argument is the label: `_flush_reports` prefixes every line with a path, and a block in memory has none. Flushing inside would invent a label or emit an unlabelled one — and because the flush clears the buffer, it would silently empty it before `_mode_parity`'s own flush, changing the CLI's committed `REPORT <path>: <msg>` format for every existing consumer. The file's design is *pure validator, caller flushes*; `parity_block.py` was the one caller that didn't.

The new rows **assert stderr, not the buffer** — asserting the buffer would pass against a `flush_reports` that writes nothing, which is the defect itself.

## A guard broken by a comment, and a fifth SIGPIPE

`check_no_competing_harnesses.sh` went red because PERF-037's **explanatory prose** contains `tokens_per_sec` and `computes_rate` grepped the whole file, comments included. The script computes no rate — it delegates to `apr test llm bench`.

Both halves fixed: `parity_host_receipt.sh` allowlisted on the basis that it **drives** the canonical instrument rather than defining a second one (the same basis `perf000_serialization_probe.sh` already uses), **and** `code_only()` strips comments before matching, with a case-table row so prose cannot re-break it.

**Full-line comments only, deliberately**: stripping trailing `# …` would mangle `${var#pattern}` and `#` inside quoted strings, and an over-strip can only make a real harness *invisible*. A false negative defeats the guard's purpose; a false positive merely annoys. The residual is recorded as a case-table row, not left to be rediscovered.

And while writing that fix: `code_only "$f" | grep -q X` returns **141 even when grep matches** — grep exits first, `sed` takes SIGPIPE, `pipefail` hands over sed's signal. **The fifth instance of that class today.** Herestrings throughout.

Re-mutated in the widened scope, 15 rows. The one that matters: **restore a real harness → still DETECTED** (`count=1 baseline=0`, rc=1) — proof the strip did not blind the guard.

## Verification

| gate | before | after |
|---|---|---|
| 93 `check_*.sh` from `ci.yml`, derived at run time | 91×0, **2×1** | **93×0** |
| `check_no_competing_harnesses.sh --selftest` | 8 rows | **15 rows, 0 broken** |
| `check_parity_receipt.sh` | 0 | 0 (+3 module-path rows) |
| `cargo test -p apr-cli --lib` | 7133 | 7133 |
| `cargo test -p aprender-serve --lib` | 15705 | 15705 |

`cargo clippy --workspace --all-targets` and `--benches` are rc=101 **on the batch and on `origin/main` alike**, with identical failing-unit sets in both directions. Note that command is *stronger than CI* — `ci.yml` has no clippy step at all (line 6 is a comment claiming one; lint is delegated upstream), which is why it is red on untouched main.

`surface_audit.csv`: all **77** citations to `extended_commands.rs` re-audited first — every cited line byte-identical, **none inside any changed hunk**. Exactly one row edited, by exact whole-line replacement rather than rewriting through the `csv` module, which would have reformatted quoting across all 830 rows.

## Two things worth knowing

**A comment failed a complexity gate.** The PMAT pre-commit hook blocked the first commit: an added **comment block** raised `main()`'s cognitive complexity 24 → 27 past the threshold, with cyclomatic unchanged at 8. `--no-verify` was **not** used — the comment was condensed and measured back to exactly 24 with the gate's own command on both sides.

**No batch-only defect was found**, and that is stated as a negative result rather than a guarantee — the two previous batches each found one.

## Unproven

The producer/consumer composition is proved against a **synthetic `$WORK`** and a shell harness reproducing the script's tail, never against live `apr serve` / `llama-server`; `harness_preflight` has never executed. #2736's rule is proven by fixtures only and stays dormant in production until #2719 lands a producer that emits the fields.

Refs #2706

### DIFF
--- a/contracts/apr-cli-operations-v1.yaml
+++ b/contracts/apr-cli-operations-v1.yaml
@@ -253,6 +253,11 @@ falsification_tests:
   prediction: apr cbtop --iterations 0 is rejected at parse time rather than producing a report scoring every brick a perfect 100/A from zero samples
   test: tests::test_parse_cbtop_rejects_zero_iterations
   if_fails: Appending --iterations 0 turns any cbtop --ci gate unconditionally green
+- id: FALSIFY-OPS-010
+  rule: Report timestamps carry the real UTC date
+  prediction: The timestamp in a cbtop report matches the UTC date at the moment it was generated, on both headless paths
+  test: commands::cbtop::tests::test_chrono_timestamp_is_current_utc_date
+  if_fails: Persisted CI provenance is stamped with a stale hardcoded date
 - id: FALSIFY-OPS-011
   rule: Both flags that shape a measurement are validated, not just one
   prediction: apr cbtop --warmup 0 is rejected at parse time (exit 2), exactly as --iterations 0 already is, while --warmup 1, --warmup 3 and the declared default 10 all still parse
@@ -263,11 +268,6 @@ falsification_tests:
   prediction: The JSON emitted by apr cbtop --headless --json records both the warmup and the iteration count it ran, so two reports differing only in warmup do not serialise identically
   test: commands::cbtop::tests::test_json_output_records_measurement_parameters
   if_fails: A cold-start number and a steady-state number are byte-indistinguishable in a file persisted with --output, and no consumer of the receipt can recover which regime produced it
-- id: FALSIFY-OPS-010
-  rule: Report timestamps carry the real UTC date
-  prediction: The timestamp in a cbtop report matches the UTC date at the moment it was generated, on both headless paths
-  test: commands::cbtop::tests::test_chrono_timestamp_is_current_utc_date
-  if_fails: Persisted CI provenance is stamped with a stale hardcoded date
 kani_harnesses:
 - id: KANI-OPS-001
   obligation: ReadOnly commands have no side effects
@@ -322,7 +322,7 @@ qa_gate:
   - progress_reporting
   - concurrent_model_access
   - tokenizer_consistency
-  pass_criteria: All 10 falsification tests pass
+  pass_criteria: All 12 falsification tests pass
   falsification: >
     Run read-only command with filesystem watcher to detect mutations.
     Force inference error and measure GPU memory leak.
--- a/contracts/apr-cli-operations-v1.yaml
+++ b/contracts/apr-cli-operations-v1.yaml
@@ -253,6 +253,16 @@ falsification_tests:
   prediction: apr cbtop --iterations 0 is rejected at parse time rather than producing a report scoring every brick a perfect 100/A from zero samples
   test: tests::test_parse_cbtop_rejects_zero_iterations
   if_fails: Appending --iterations 0 turns any cbtop --ci gate unconditionally green
+- id: FALSIFY-OPS-011
+  rule: Both flags that shape a measurement are validated, not just one
+  prediction: apr cbtop --warmup 0 is rejected at parse time (exit 2), exactly as --iterations 0 already is, while --warmup 1, --warmup 3 and the declared default 10 all still parse
+  test: tests::test_parse_cbtop_rejects_zero_warmup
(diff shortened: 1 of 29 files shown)


noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:I-24 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `crates/apr-cli/src/commands/cbtop_get_cpu_memory.rs`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Batch of five sub-tickets; gate was GREEN on the stale base (11 pass), so not split. Rebased with --rebase-merges; the conflict is inside the batch's own p3-warmup merge replay (commit 5a29ac364), not against main. Linked issues map to I-24 (#2735, 2026-09-19), G-9 (#2730), P-0.3 (#2732); intersects I-24's parity_block.py. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2741` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2741@06acb9fb5 · intent → I-24 (#2887; holds #2735) — also G-9 #2899 (#2730), P-0.3 #2911 (#2732) · salvage: crates/apr-cli/src/commands/cbtop_falsification_summary_hardware.rs crates/aprender-core/tests/falsification_spec_v10_tests.rs crates/aprender-core/tests/includes/falsification_spec_v10_contract_model.rs scripts/lib/parity_receipt_cases/{25..30}_*.json · conflicts: crates/apr-cli/src/commands/cbtop_get_cpu_memory.rs (inside the batch's own p3-warmup merge replay, 5a29ac364)`

PR #2744, opened 2026-08-28
feat(perf): the §4.4 protocol had no caller and the gate's real mode had no input (PERF-025)

noahgift:
**Stacked on #2742 (`feat/r1-2719`) deliberately — see "Base decision" below. Merge #2742 first.**

## The defect

On `50d2bc2bb`, `git grep "ReceiptInput\|BandInput\|DerivedBand\|perf_gate::"` outside the module returned **rc=1**. PERF-024 shipped the conformant §4.4 measurement protocol and PERF-026 shipped the receipt producer, and **nothing called either**. `apr test llm bench` still ran `LoadTest::run`, whose termination rule is "stop after `--duration` seconds": no minimum sample count, no warmup-then-quiesce, no drain accounting, no §4.4.6 tokenization block.

The mirror defect: `scripts/perf_gate.sh`'s real mode (`--host/--phase/--workload/--receipt`) was invoked only as `--selftest` (`ci.yml:1011`), because nothing in the repo could write the receipt it reads. Both halves close here.

## Base decision — why this stacks rather than sitting on main

`llm/band.rs` and its five dependency modules (`protocol`, `metrics`, `window`, `bootstrap`, `samples`) **do not exist on main**. Without them there is no `run_band` to call. main's own `perf_gate/mod.rs` says so, and names the integration work:

> Two overlaps must be resolved in the integration commit rather than left to drift: `protocol::Tokenization` and `receipt::TokenizationBlock` [...] and `protocol::Outcome` and `drain::Outcome` [...] Keep one of each.
> [...] Wiring a caller for the gate's real (non-selftest) mode is a separate ticket.

#2742 already performed exactly that resolution. Re-doing it on main would mean a second copy of ~2,500 lines that can drift from #2742 — the duplicate this epic has already rejected once. So this stacks.

## What is new here vs #2742

Only the CLI wiring — 7 files, +949/-48. The abandoned `feat/n1-band-cli` branch (no PR, cut against a pre-`drain.rs` base) contributed the *shape* of this surface; its library half is fully redundant with #2742 and its handler targeted a receipt API that no longer exists (`ReceiptMeta`, `Replicate`, `write_receipt`, `Tokenization` struct, `BandRun::into_replicate` — all removed or replaced by the #2742 resolution). Nothing from it is carried forward verbatim.

## Design notes

- **A mode, not a second harness.** `check_no_competing_harnesses.sh` names `apr test llm bench` canonical at `BASELINE=0`; this extends that subcommand and drives the same `LlmClient`. Guard still `count=0 baseline=0 OK`.
- **No flag shrinks the window.** `BandConfig::relaxed` stays unreachable from the CLI. `--replicates` is the only knob and defaults to §4.4.2's `N=3`.
- **The four unhonourable flags are refused, not ignored.** `--start`, `--output`, `--baseline`, `--fail-on-regression` are refused by clap `conflicts_with_all` — the `apr serve --gpu` silently-ignored-flag class that opened this epic.
- **One receipt per replicate.** `ReceiptInput.bands` holds one entry per concurrency and Arm A indexes it `{b["concurrency"]: b for b in bands}`, so N bands all saying `concurrency: 1` would be silently reduced to whichever landed last. Collapsing to a "representative" band would mean a median of three — the estimator that reported 2.91x for a cell whose real answer was 1.21x (#2567).

## Proof the conformant path ran, not the legacy one

Against one stub endpoint with an **identical `--duration 5`**:

| mode | rc | elapsed | requests |
|---|---|---|---|
| legacy (`LoadTest::run`) | 0 | **5 s** | 61 |
| `--band` | 0 | **88 s** | 728 |

`run_band` cannot terminate on duration; it terminated on the 60 s wall-clock floor plus warmup and the 5 s quiesce. The receipt carries `drain_ms`, `window_ms`, `client_model: closed_loop` and the §4.4.6 `tokenization` block — none of which `LoadTest::run` can produce. A 4-band sweep reports `peak_in_flight` 1/4/8/16 and a non-zero `drain_ms` at c>=4.

- `python3 scripts/lib/bench_receipt.py <receipt>` → **rc=0, `ok`**
- `bash scripts/perf_gate.sh --host ... --phase merge --workload W1 --receipt <receipt>` → **rc=0, `VERDICT PASS`** — the first real (non-selftest) caller that mode has ever had.

## Nothing was measured

The endpoint was a throwaway local stub, never committed. `scripts/perf-matrix.yaml` is **byte-identical to main** and all 8 cells remain `UNMEASURED`. Running a real cell is the next ticket.

## Checks

- `cargo fmt --all -- --check` rc=0
- `cargo clippy -p aprender-test-lib -p apr-cli --lib -- -D warnings` rc=0
- `cargo clippy ... --all-features` is rc=101 on this branch **and identically on the untouched base** (8 pre-existing `unused variable` errors in `aprender-gpu`) — control established, zero regressions
- `cargo test -p aprender-test-lib --lib perf_gate::` 87/87
- `cargo test -p apr-cli --lib test_llm` 20/20 (13 new)
- `perf_gate.sh --selftest` 18/18; `check_no_competing_harnesses.sh --selftest` 9/9
- PMAT pre-commit complexity gate passed without `--no-verify` (the band handler is its own file, so `test_llm.rs`'s complexity is untouched)

### DIFF
--- a/crates/apr-cli/src/commands/mod.rs
+++ b/crates/apr-cli/src/commands/mod.rs
@@ -138,6 +138,10 @@ pub(crate) mod prometheus_lint;
 pub(crate) mod prune;
 pub(crate) mod ps_schema;
 pub(crate) mod test_llm;
+// PERF-025: the `--band` mode of `apr test llm bench`. Its own file so that
+// `test_llm.rs`'s complexity is untouched by it -- the PMAT pre-commit gate
+// blocks a FILE with any pre-existing violation, not a function.
+pub(crate) mod test_llm_band;
 // #2399: gated on the crate it actually needs (aprender-explain, aliased
 // `trueno-explain`) rather than on `full`, so `--features ptx` is enough and a
 // user does not have to pull CUDA + training to analyze a .ptx file.
--- a/crates/apr-cli/src/commands/test_llm.rs
+++ b/crates/apr-cli/src/commands/test_llm.rs
@@ -199,7 +199,7 @@ pub async fn run_bench(args: BenchArgs<'_>) -> Result<()> {
 /// A file overrides the profile; an unknown profile name is rejected rather
 /// than quietly falling back to a default, since a silent substitution changes
 /// the workload the report then claims to have run.
-fn resolve_prompts(profile: &str, file: Option<&Path>) -> Result<Vec<ChatRequest>> {
+pub(crate) fn resolve_prompts(profile: &str, file: Option<&Path>) -> Result<Vec<ChatRequest>> {
     if let Some(p) = file {
         return load_prompts_from_file(p)
             .map_err(|e| CliError::InvalidFormat(format!("prompt file {}: {e}", p.display())));
@@ -213,7 +213,7 @@ fn resolve_prompts(profile: &str, file: Option<&Path>) -> Result<Vec<ChatRequest
 }
 
 /// One line naming the workload, so the report is self-describing.
-fn describe_workload(profile: &str, file: Option<&Path>, count: usize) -> String {
+pub(crate) fn describe_workload(profile: &str, file: Option<&Path>, count: usize) -> String {
     match file {
         Some(p) => format!("{} prompt(s) from {}", count, p.display()),
         None => format!("profile {profile} ({count} prompt(s))"),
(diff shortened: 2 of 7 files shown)


noahgift:
### Additional evidence (all against a throwaway local stub; no cell measured, `perf-matrix.yaml` untouched)

**4-band sweep** — `--bands 1,4,8,16 --replicates 1`, rc=0, 313 s, 20,816 requests, receipt 479,930 B. `perf_gate.sh --host lambda --phase merge --workload W1`:

```
PASS   ArmC integrity
REPORT ArmA c=4  scaling_efficiency=1.0024 (baseline UNMEASURED until 2026-09-25, owner=perf-gate)
REPORT ArmA c=8  scaling_efficiency=0.9554 (...)
REPORT ArmA c=16 scaling_efficiency=0.9977 (...)
REPORT ArmB c=1/4/8/16 UNMEASURED (Arm A still gates this cell)
REPORT ArmD instrumentation absent: kv.*
VERDICT PASS host=lambda phase=merge workload=W1
```

Arm A prints nothing on a single-band receipt because its loop is `for c in sorted(bands): if c==1: continue` — that is the one-band case, not a dead arm.

**The receipt is correctly incomplete rather than fabricated.** Same receipt at `--phase release` → rc=1:

```
FAIL ArmD instrumentation absent: kv.bytes_used/bytes_reserved, kv.admission_rejected, kv.preempted_swap
PASS cells host=lambda all bands [1, 4, 8, 16] present
VERDICT FAIL host=lambda phase=release workload=W1
```

The Arm D block is server-reported, so it is named in `unproduced_fields` instead of being invented — legal at merge, correctly fatal at release.

**Fields the legacy path cannot produce**, read back from the c=16 band: `window_ms 60000.3` (the §4.4.2 wall-clock floor), `drain_ms 76.62`, `ttft_p95_ms 6.55`, top-level `client_model: closed_loop`, `n: 20816`.

**`--replicates 3` writes three independently judgeable receipts**, which is the design claim in the PR body:

| | bench_receipt.py | perf_gate.sh | `bands[0].aggregate_tok_per_sec` |
|---|---|---|---|
| `receipt.r1.json` | rc=0 | rc=0 `VERDICT PASS` | 194.42 |
| `receipt.r2.json` | rc=0 | rc=0 `VERDICT PASS` | 193.52 |
| `receipt.r3.json` | rc=0 | rc=0 `VERDICT PASS` | 193.09 |

### Stated plainly as unproven

- **No real inference server was driven.** The endpoint was a Python SSE stub, so the CLI's handling of a real server's `usage` reporting, EOS behaviour and W1's `ignore_eos`/`seed` is untested. Running a real cell is the next ticket.
- The band mode **refuses `--start`**, so this CLI never selects a server or a router — it measures whatever is already listening. It therefore cannot reach the second `apr serve` router (`apr-cli/src/commands/serve/routes.rs`, whose request type silently drops `seed` and `ignore_eos`); which router serves is the operator's `apr serve` invocation, not this client's.

Issue #2766, opened 2026-08-29
pre-commit complexity gate freezes gemm.rs: it scans staged files, so any edit is rejected on a pre-existing Cognitive 30


noahgift:
`cublas_prefill_gemm` measures **Cognitive 30** against a threshold of **25** at HEAD — verified with `pmat analyze complexity` on a stashed tree, not inferred from a failure message.

The pre-commit quality gate scans **staged files**, not the diff. So any edit to `gemm.rs` is rejected on the pre-existing complexity of a function the edit does not touch.

**The file is effectively frozen.** That matters right now because `gemm.rs` holds the FP8 GEMM at the centre of #2753, and it is where a per-GEMM input fingerprint naturally belongs — the instrumentation could not be placed at its natural site for this reason and had to go somewhere less apt.

## Why this is a gate-design problem rather than a code-cleanliness one

A threshold that blocks *edits to* a file rather than *increases in* its complexity inverts the incentive. The cheapest way to make a change to `gemm.rs` today is to not make it, or to put the code somewhere it does not belong — which is what happened here. It also means the file can only get worse: nobody can land the small refactor that would bring it under the threshold, because that refactor is itself an edit to a staged file that measures 30.

This is the same family as #2757 (a ratchet that makes a newly-discovered defect unrecordable) — a gate whose failure mode is to suppress the work rather than the defect.

## Suggested

Gate on the **delta** rather than the absolute: refuse a commit that raises a function's complexity or introduces a new function above threshold, and permit edits to a file whose pre-existing violations the commit does not worsen. That keeps the ratchet property — nothing gets worse — without freezing the files that most need attention.

Either way `cublas_prefill_gemm` deserves the refactor on its own merits; the point is that today nobody can land it.

noahgift:
Second instance, found during #2753's round-6 bisect.

`forward_utils.rs` is frozen the same way: `debug_q6k_lm_head` measures **Cognitive 41** against a threshold of 25, and it is reached through an `include!`, so the violation is not even visible in the file being edited.

That makes two files at the centre of an active investigation that cannot be modified — `gemm.rs` (Cognitive 30) and `forward_utils.rs` (Cognitive 41). In both cases the instrumentation had to be placed somewhere less apt, or avoided entirely.

The `include!` detail sharpens the original argument: a contributor editing `forward_utils.rs` sees a rejection naming a function that is not in the file they opened. The gate is not merely blocking unrelated work, it is doing so opaquely.

[...]

Issue #2757, opened 2026-08-28
G2.3's ungated ratchet makes a newly-found defect unrecordable, and G2.1 forces a touch rather than a correction


noahgift:
Found while closing G2.1 on #2742 (PERF-046). It did not bite — nothing was broken — but the next auditor who finds something will hit it.

**G2.3's `broken-and-ungated` ratchet makes a newly-discovered defect on ungated surface unrecordable.** Adding a 45th `quality<=4` ungated row fails the gate *even with a triage entry*. The honest paths available to the finder are: write the gate for that surface, or fix the defect.

Both are good outcomes in the abstract. The problem is what happens under time pressure: the cheapest way to make the gate green is to **not record the finding at all**, or to score the row above 4. A ratchet that makes honest reporting the most expensive option is pointed the wrong way — it ratchets the *ledger*, not the defect count.

Compare the shape the repo already accepts elsewhere: a shrink-only ratchet on **triaged** debt, where recording something new is permitted precisely because it has been triaged, and only the untriaged count may not grow.

## Related, from the same investigation: G2.1's honest limit now has a concrete instance

The gate's own header says it "forces the ledger to be TOUCHED, not to be CORRECT." Here is the first case anyone can point at: commit `ce712eae0` moved the `apr eval --task *` code **45 lines** and touched `docs/audits/surface_audit.csv` by exactly **1 insertion, 1 deletion**. G2.1 passed. The ledger went stale anyway, and stayed stale until #2742 happened to move the same code again and forced a real re-audit — which then found **10 rows** each citing a line 6 short of the arm it names.

So the touch-requirement bought nothing on that commit. Worth considering whether G2.1 can cheaply check that cited lines still land on a plausible definition, rather than only that the file was edited. Note the weaker version of this check is easy to get wrong: verifying the cited line merely exists and is non-blank passes all 10 of the stale rows above.

noahgift:
Closed by paiml-implement — #2879

noahgift:
Reopened by the PMAT-1100 triage orchestrator: this is a child of its epic (sub-issue), not a duplicate of it — the lane's close was wrong. Filed to a train below.

noahgift:
triage 2026-09-23 (aprender-dd, standing triage owner): milestoned to 0.70.0, theme Fast Train (CI, gates, pv, release tooling, hygiene); it moved out of `backlog` (the 2026-09-17 overflow bucket, not a release).

Issue #2753, opened 2026-08-28
P0: batched CUDA decode emits garbage for every m>1 — PERF-001's 3.32x aggregate is throughput of garbage tokens


noahgift:
**Batched CUDA decode produces garbage output for every `m > 1`.** The slots served from a batch emit a constant token to the cap, never emit a stop token, and therefore always run to `max_tokens`.

Found while root-causing F-BATCH-001's RED (#2747). The RED was real; the *cause* was not serialization.

## Evidence

One concurrent round, with `[PMAT-044] Batch m=3 done` in the server log proving the batched path engaged:

| slot | path | output | finish | tokens |
|---|---|---|---|---|
| A | m=1 fast path | coherent English | `stop` | 319 |
| B, C, D | served from the batch | `!!!!…` (token id 0 to cap) | `length` | 400 |

**Same binary, same model, same prompt, same second — only the path differs.** At m=4 the garbage takes other forms (`strarstrarstrar…`, `001.CompilerServicesétique…`). Reproduced on `qwen2.5-coder-0.5b` with m=3 slots garbage and m=1 slots fine *in the same run*, so it is not model-specific.

## What this invalidates

**PERF-001's credited "3.32x aggregate at N=8, ~1100 tok/s" is throughput of garbage tokens.** Its `completed` status in `docs/roadmaps/roadmap.yaml` needs revisiting on those grounds, not merely on F-BATCH-001 being RED.

It also explains the shape that made the serialization reading look plausible. The index is *flat* from c=2 (2.45 / 2.29 / 2.39) because every batched band hits the same `max_tokens` cap regardless of c. The recorded numbers refuted serialization before any new run: with `wall(1)=1.20s` the band walls are **1.20 / 2.94 / 2.75 / 2.87**, so `wall(2) > wall(4)`. Adding a client cannot make a serialized server faster; the quantity charged at c=2 is a constant. Two independent recordings agree on that direction.

The constant has two parts. The correctness defect above contributes 400/319 = 1.25x. A real per-token path penalty contributes the rest: c=1 replays a captured CUDA decode graph while c>1 launches eagerly, and the batched loop retakes `model.write()`, rescans for recycle/join, and embeds and `try_send`s per slot **per token** — measured at 2.93 ms/tok fast versus 5.01 ms/tok batched, **1.71x token-normalised**. Neither part scales with m, which is the entire flat shape.

[...]

PR #2773, opened 2026-08-29
batch: two gates that suppressed the work instead of the defect, and the guard the batch found unwired

noahgift:
Three sub-tickets for APR-PERF-GATE-001 (#2706). Two of them fix **gates whose failure mode was to suppress the work rather than the defect**; the third is a batch of guards from earlier in the epic.

## PERF-051 — the complexity gate froze the files most worth editing (#2766)

The pre-commit gate scanned **staged files** with an **absolute** threshold, so a file carrying a pre-existing violation could not be edited at all. During #2753 this blocked **four** files, each time forcing instrumentation somewhere less apt:

```
gemm.rs           cublas_prefill_gemm   Cognitive 30
forward_utils.rs  debug_q6k_lm_head     Cognitive 41, reached through include!
gemv_dispatch.rs  three violations, two of them in OTHER files
apply.rs          apply_qkv_bias        Cognitive 31
```

**The gate was not where the tracked `.githooks/` suggests.** `core.hooksPath = .git/hooks-apr`, which execs a pmat-generated `.git/hooks/pre-commit` marked "DO NOT EDIT" and untracked. The tracked `.githooks/pre-commit` is dead in this checkout and never had a complexity check at all.

Two properties nobody had recorded: **pmat follows `include!`**, so analysing `forward_utils.rs` reports a violation in `logits.rs`; and when pmat cannot produce a report, the `grep -qE 'Errors: *[1-9]'` simply does not match, so **the old gate silently passed**.

Now gated on the **delta** against HEAD. Measured both directions on the real files:

```
gemm.rs / forward_utils.rs / gemv_dispatch.rs / apply.rs   OLD rc=1 -> NEW rc=0
deepening cublas_prefill_gemm   REFUSED: WORSE ... = 40 (was 30 at HEAD)
```

**The live hook was deliberately not installed.** It is untracked, shared across every worktree on this box, and other agents are committing right now — run `make hooks-install`. Case tables found two defects review would not have: an unanchored `grep -n | cut` exited under `set -e`+`pipefail` so the installer failed **mutely**, and `END_MARK_RE` matched two lines because the no-source-files fast path prints the same marker ~50 lines earlier.

## PERF-052 — the ungated ratchet made a new defect unrecordable (#2757)

G2.3 ratcheted the **total** `broken-and-ungated` count, so recording a 45th `quality<=4` ungated row failed **even with a triage entry**. Under time pressure the cheapest way to green was to not record the finding, or to score it above 4 — a ratchet that makes honest reporting the most expensive option.

Reproduced against main before changing anything:

```
G2.4 waivers  PASS  45 broken-and-ungated, 45 triaged, 1 new
G2.3 floors   FAIL: broken-and-ungated is 45, must be <= 44
```

Now ratchets `untriaged_broken()`, floored from the **comparand's** ledger, not the working tree.

**Removing the total ratchet is not a net weakening, and that was tested rather than asserted.** The total blocked three moves; only one was a real regression — a gated row *losing* its gate while broken. That is now `G2.3 gate-loss`, keyed **per row**, because a count cannot express it: moving a gate off a broken row onto a healthy peer in the same cluster and binary leaves covered-total, per-binary and per-cluster counts all untouched. Verified on the real 830-row ledger, where that count-neutral swap goes RED and deleting the new rule makes it GREEN.

G2.1 was left alone deliberately. A cheap version of its citation check — verifying the cited line merely exists and is non-blank — **passes all ten of the stale rows** found earlier in this epic, and a check that launders stale rows is worse than none.

## PERF-042 / PERF-043 — pin-key and clippy-scope guards

`check_pin_keys_consumed.sh` (35 declared keys, 27 consumed, 8 ledgered on a two-way ratchet) and `check_gate_scope_not_narrowed.sh` + `lib/gate_scope.py` with 10 fixtures. `feat/v1-batchred` was **excluded** after confirming it is already an ancestor of `feat/v9-batchfix` (#2771), so nothing duplicates.

## The integration defect only the batch could see

`check_complexity_gate_wiring.sh` arrived named by no workflow — v12 left wiring out of scope — and `check_guards_are_wired.sh` refused it. Each unit is green alone; only the batch puts a new guard in front of the guard that governs guards. Two steps added to `ci.yml`: the wiring guard, and `make hooks-delta-test` for its 66 case-table rows.

## Verification (combined tree)

`check_guards_are_wired`, `check_dogfood_coverage` (+self-test), `check_baseline_ratchets`, `check_no_claim_literals`, `check_mutation_registry`, `check_roadmap_completion_is_cited`, `perf_gate --selftest`, `check_workflow_env_defined`, `check_runner_labels`, `check_complexity_gate_wiring`, `make hooks-delta-test` — all rc=0.

Both gate changes **loosen** a gate, so each was independently verified to still go RED: 8 constructions for the delta gate (including new-file, new-function-in-violating-file, and rename-and-worsen) and 5 for the triage ratchet (including a count-neutral gate-loss swap and a rename evasion) — all built against fresh fixtures rather than replaying the authors' own cases.

`scripts/perf-matrix.yaml` untouched; all 8 cells remain `UNMEASURED`.


noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:I-24 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `scripts/check_bench_protocol.sh scripts/check_comparator_flags.sh scripts/lib/parity_block.py scripts/llama_bin.sh scripts/llama_pin.toml scripts/parity_host_receipt.sh`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Batch; gate was GREEN on the stale base (11 pass), so not split. The first conflict (.gitignore, both sides appended) was resolved as a union and the replay continued; the second stop is the comparator-pin/parity-block cluster, which main reworked. Linked issues map to G-8 (#2757) and S-2 (#2766); intersects I-24 (parity_block.py), C0-1/G-4 (ci.yml). Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2773` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2773@e9314a707 · intent → G-8 (#2898; holds #2757) — also S-2 #2922 (#2766) · salvage: scripts/lib/gate_scope_cases/*.{json,yml} (10) scripts/tests/complexity_delta_gate_cases.sh scripts/tests/complexity_delta_gate_install_cases.sh scripts/tests/complexity_delta_gate_regex_cases.sh scripts/tests/fixtures/pmat_generated_pre_commit.fixture · conflicts: scripts/check_bench_protocol.sh scripts/check_comparator_flags.sh scripts/lib/parity_block.py scripts/llama_bin.sh scripts/llama_pin.toml scripts/parity_host_receipt.sh (.gitignore was both-appended and unioned before the replay stopped)`

PR #2794, opened 2026-08-30
feat(mcp): a thin single-model MCP server for SetFit

guyernest:
## A thin single-model MCP server for SetFit

Serves SetFit classification over MCP on the pmcp SDK, so an agent or chat
client can call a model trained with `aprender` as a tool.

Builds on #2618 (now merged). Nothing else is pending.

```
crates/aprender-mcp-setfit           the server        717 lines
crates/aprender-mcp-setfit-lambda    Lambda unit       333 lines
tests/scenarios/                     post-deploy       186 lines
```

### One model per server, deliberately

The consumers are agents wired into a specific business process. A broad ML
toolbox behind one endpoint adds attack and confusion surface with no workflow
value, so this server exposes exactly one tool. `apr mcp` stays the general
developer-toolchain surface — this is not a wrapper around it.

The intent is that this crate is the **template**: the shape you copy to wrap
any other aprender model as its own curated MCP endpoint.

### Transport only

It calls `aprender-core`'s `VerifiedSetFitModel::classify` in-process and
re-implements nothing — no tokenizer, no pooling, no head. That is OPS-03, one
implementation per operation: core's path is the one with fixture-verified
conformance behind it, so a second port here would re-open every conformance
question #2618 settled.

Both crates are `publish = false`. They are deployment units, not libraries to
depend on.

### What runs, and what doesn't

Stated plainly, because "a test that cannot run" is a defect this repo takes
seriously:

- **`tests/e2e_stdio.rs`** drives the real binary over stdio MCP and asserts a
  classification round-trip. Env-gated on `APR_MCP_E2E_SETFIT_MODEL` — no
  SetFit-tagged artifact can be checked in — and it *announces* the skip:
  `E2E-SETFIT-PMCP-001 SKIP: APR_MCP_E2E_SETFIT_MODEL unset — …`. It pins the
  binary with `env!("CARGO_BIN_EXE_...")`, not a PATH lookup.
- **`tests/scenarios/*.yaml` are NOT CI tests.** They need a deployed URL and
  are run by hand:
  `cargo pmcp test run <url> --scenarios tests/scenarios/ --transport jsonrpc`.
  `--transport jsonrpc` is required, not a preference: cargo-pmcp's `test run`
  never issues the initialize step, so the default http transport fails every
  tool_call with "Client not initialized".

### Verification

rc captured directly, never through a pipe:

```
cargo check  -p aprender-mcp-setfit -p aprender-mcp-setfit-lambda
             --all-targets                                  rc=0
cargo test   (both crates)                                  9 passed / 0 failed
cargo clippy --all-targets --no-deps -- -D warnings         rc=0
cargo fmt --all -- --check                                  rc=0
bash scripts/check_publish_safety.sh                        rc=0, 9/9
bash scripts/check_contract_test_binding.sh                 rc=0, PASS
```

The last two are there because #2618 taught me that building and testing a
branch proves the code works, not that it can be shipped.

### Deliberately not in this PR

The CDK stack that puts this on AWS (`deploy/`, ~790 lines, 521 of them a
generated npm lockfile). Adding TypeScript and a node lockfile to a Rust
monorepo is an architectural call worth making on its own, not as a rider on a
server crate. Happy to open it separately if you want it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -841,6 +841,7 @@ dependencies = [
  "schemars 1.2.1",
  "serde",
  "serde_json",
+ "thiserror 2.0.18",
  "tokio",
 ]
 
@@ -850,7 +851,6 @@ version = "0.64.0"
 dependencies = [
  "aprender-mcp-setfit",
  "lambda_http",
- "once_cell",
  "pmcp",
  "reqwest 0.12.28",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -833,6 +833,33 @@ dependencies = [
 ]
 
 [[package]]
+name = "aprender-mcp-setfit"
+version = "0.64.0"
+dependencies = [
+ "aprender-core",
+ "pmcp",
+ "schemars 1.2.1",
+ "serde",
+ "serde_json",
+ "tokio",
+]
+
+[[package]]
+name = "aprender-mcp-setfit-lambda"
+version = "0.64.0"
+dependencies = [
+ "aprender-mcp-setfit",
+ "lambda_http",
+ "once_cell",
+ "pmcp",
+ "reqwest 0.12.28",
+ "serde_json",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "aprender-monte-carlo"
 version = "0.64.0"
 dependencies = [
@@ -2992,6 +3019,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws_lambda_events"
+version = "0.15.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7319a086b79c3ff026a33a61e80f04fd3885fbb73237981ea080d21944e1cb1c"
+dependencies = [
+ "base64 0.22.1",
+ "bytes",
+ "http 1.4.2",
+ "http-body 1.0.1",
+ "http-serde",
+ "query_map",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "axum"
 version = "0.7.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3603,6 +3646,9 @@ name = "bytes"
 version = "1.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8ae3f5d315924270530207e2a68396c3cc547f6dca3fbdca317cfb1a51edb593"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "bytes-utils"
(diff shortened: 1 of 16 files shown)


noahgift:
PP-066 triage (epic #2873, ticket #2986) — **review only**: this PR is not by the triage owner, so no disposition, no label, no branch edit. Read against `origin/main` @ cdc0acb99 (`gh pr diff 2794`; API names checked with `git grep` on main).

**Blocking once rebased**

1. `crates/aprender-mcp-setfit/Cargo.toml` pins `aprender = { path = "../aprender-core", version = "0.64.0", package = "aprender-core", … }`. The workspace is at **0.65.2**, and cargo refuses a path dependency whose package version does not satisfy the requirement (`^0.64.0` excludes 0.65.x), so the crate cannot resolve after a rebase. Every other in-tree consumer spells it `version = "0.65.2"` (11 manifests; `scripts/bump-version.sh` maintains that literal). Cargo.lock also records both new crates at 0.64.0 and needs regenerating on the rebase.
2. **No check has ever run on this branch** (`gh pr checks 2794` → "no checks reported"; `mergeable: UNKNOWN`). The receipt gate moved to a base-owned `pr-review-quorum.yml` (`pull_request_target` + `merge_group`) in #2985 today, so the first real signal is a rebase plus a maintainer-approved workflow run. Nothing here can be judged until `ci / gate` speaks.

**Should fix before merge**

3. `src/main.rs` hand-rolls the argv parser (`match arg.as_str() { "--model" => … }`). It refuses unknown flags (good), but the repo rule after simular's `--seed` was silently dropped is clap for every binary (`apr` is clap end to end), and the surface dogfood sweep enumerates workspace binaries at runtime and expects `--help`/`--version` to answer — the new `aprender-mcp-setfit` and `bootstrap` bins answer neither.
4. Both integration targets are dark in CI: `tests/e2e_stdio.rs` is env-gated (the visible `SKIP` line is the accepted pattern here — `#[ignore]` is banned) and `tests/embed.rs` is build-gated, and neither is in ci.yml's explicit `--test` list (workspace-test runs `--lib` only). The only CI coverage is the seven unit tests in `lib.rs`. At minimum wire `cargo test -p aprender-mcp-setfit --test e2e_stdio` so CI proves the target compiles and prints the SKIP.
5. cargo-deny: `lambda_http 0.13`, `aws_lambda_events 0.15.1`, `lambda_runtime 0.13`, `query_map`, `http-serde` are new to the tree — their licences need to be in `deny.toml`'s allow list (`cargo deny check licenses` is the C0-1 required check for 0.66). `schemars_derive 1.2.1` adds a third schemars line beside 0.8.22 and 0.9.0 (`multiple-versions = "warn"`, so a warning, not a block).

**Nits**

6. The PR body cites `tests/scenarios/` (186 lines); the files live at `crates/aprender-mcp-setfit/scenarios/`. The README's "See the deploy crate/config added in the deployment phase" now refers to a crate in this same PR.
7. Heads-up, not a defect: the 0.66 crate-name guard (DAG row G-1) counts workspace members (82 today); two `publish = false` deployment units raise that and belong in its allow-list at landing time. `[lib] name = "aprender_mcp_setfit_lambda"` equals the default, so the name rule itself passes.

What reads well: classification stays in-process on `VerifiedSetFitModel::classify` (OPS-03, no second inference path), both bounds come from `contracts/setfit-apr-v1.yaml`, `deny_unknown_fields` is asserted at the schema and through the live server, `spawn_blocking` keeps the protocol loop off the encoder, and `KillOnDrop` reaps the child on a panicking assertion. The `stateless()` + `max_request_bytes` choice for the Lambda is explained by measurement, which is the standard here.

noahgift:

No activity since 2026-09-16 08:25Z; **3 release trains have been cut since** (v0.68.2, v0.68.1, v0.68.0). Age is the only input to this sweep — it is not a judgement on the work, and a train label such as `pp-066/carry-*` is not an exemption.

noahgift:

Issue #2786, opened 2026-08-30
apr decodes at 0.579x ollama on GB10 Blackwell while passing on Ada — all three documented causes ruled out


noahgift:
`cuda-nightly` was rewired to gx10 (aarch64 GB10 Grace Blackwell) in #2740. On its **first two runs there it failed**, on a real measurement, having passed four nights running on the Ada RTX 4090.

```
BEAT-OLLAMA-DECODE-THROUGHPUT [ENFORCED median-of-7 >= 0.90x]:
  apr_median7=105.6  apr_best=140.0  ollama_median=182.3 tok/s
  ratio_median=0.579x  ratio_best=0.767x  (steady-state GPU decode)
  apr trials=[111.80, 96.78, 96.41, 99.01, 111.43, ...]
```

## This is not a flake and not the runner move

| date | runner | arch | result |
|---|---|---|---|
| 2026-08-26 | lambda-4090 | Ada, x86_64 | success |
| 2026-08-27 | lambda-4090 | Ada, x86_64 | success |
| 2026-08-28 | lambda-4090 | Ada, x86_64 | success |
| 2026-08-29 | lambda-4090 | Ada, x86_64 | success |
| **2026-08-30 04:23** | **gx10-blackwell** | **Blackwell, aarch64** | **failure 0.579x** |
| **2026-08-30 07:10** | **gx10-blackwell** | **Blackwell, aarch64** | **failure** (scheduled) |

Two independent runs, one dispatched and one scheduled. The seven trials are tightly clustered (96.4–111.8 tok/s), so this is a stable operating point, not variance.

This is exactly what `cuda-nightly.yml`'s own header predicted:

> *"A CUDA regression that reproduces on Ada but not Blackwell will not be caught here."*

It is caught now.

## All three documented causes are ruled out

The gate's failure message lists what to check, in order. I worked through them:

**(1) Is the F2 first-token gate rejecting the CUDA path?** — **Ruled out by the number itself.** The message states that a rejected CUDA path falls to wgpu, then CPU SIMD, and measures *"~20 tok/s - ratio ~0.065"*. We measured **105.6 tok/s / 0.579x**, an order of magnitude above that signature. No `GPU diverges from CPU` or `cosine < 0.95` in the log.

**(2) Is `APR_BIN` built `--features cuda`?** — **Ruled out.** The job log contains `nice -n 19 cargo build --release --features cuda --bin apr`.

[...]

PR #2800, opened 2026-08-31
evidence(perf-061): the GB10 shortfall on #2786 is a real deficit, not warm-up

noahgift:
Closes the warm-up hypothesis on #2786 with four discriminators. **The GB10 shortfall is a real deficit, not a measurement artefact.**

| Discriminator | Result |
|---|---|
| Ordering ramps? | **No.** Spearman rho +0.393 (p=0.198) and +0.179 (p=0.357), Fisher-pooled p=0.445. Maxima at trial 7 and trial 3 |
| Longer warmup moves the median? | **No.** Cold 105.96 tok/s (0.579x) vs warm 112.70 (0.616x). Delta +6.4%, permutation p=0.650 — the floor needs +55% |
| Token ladder (differences nothing) | R = 106.4 tok/s [95% CI 103.1-110.0] = 0.582x, marginal rate flat from token 128 to 1024 |
| Why ratio_best > ratio_median | Needs no warm regime: rate = 256000/Delta is a reciprocal, so it is right-skewed even at constant speed |

Three independent estimators agree with the gate: **0.582 / 0.586 / 0.579** vs the CI's **0.579**.

Scope note: `>= 0.90x` is a no-collapse floor, not a beat target.

Evidence-only; no source changes.

Refs #2786, APR-PERF-GATE-001 (#2706)

no-close: this PR closes a HYPOTHESIS about #2786, not the issue. The GB10 shortfall #2786 names is confirmed real here, so #2786 stays open.


noahgift:

```
DEGRADED-INPUTS: pmat MCP ConnectionRefused (fell back to pmat CLI 3.34.0 on an index built from d6c6c6f8f, 71 commits behind main and NOT an ancestor of the PR head); CUDA docs returned no authority on the nvidia-smi/NVML claims across 3 queries | pr-review v2.0.0 | verdict=BLOCK | consultations: pmat=consulted cuda=no-authority-found crux=consulted mutation=not-triggered | findings=4 (cited=0 measured=4 asserted=0) | index=d6c6c6f ancestor=false
```

**4 findings raised, 3 survived adversarial refutation** (three independent verifiers per finding — correctness / scope / evidence-quality lenses, refute-by-default). Refuted findings are dropped, not shown.

---

### 1. Fisher-pooled p = 0.445 does not reproduce; the correct value is 0.258, and the error flatters the PR's own conclusion

`evidence/perf-061/ci-trial-ordering-analysis.txt:19` · grounding: **measured** · blocking class: `none`

**Evidence**

```
Claimed, in four places:
  ci-trial-ordering-analysis.txt:19  POOLED (Fisher) over both runs: X2=5.303 df=4  p=0.445
  gb10-warmup.md:23   ... Fisher-pooled **p = 0.445**
  gb10-warmup.md:72   Fisher-combined **p = 0.445**.
  findings.json:32    "fisher_pooled_p": 0.445,
(and in the PR body's headline table: "Fisher-pooled p=0.445")

I reproduced the inputs exactly by enumerating all 5040 orderings, as the doc specifies:
  pA=0.197817 pB=0.356548          -> matches the stated 0.198 / 0.357
  X2=-2*(ln pA + ln pB) = 5.3034   -> matches the stated 5.303

But the p-value does not follow from that statistic:
  $ python3 -c "from scipy import stats; print(stats.chi2.sf(5.3034,4))"
  0.2575588123
```

**Why it matters** — The conclusion survives — 0.24 and 0.26 are both far from any conventional significance threshold, so 'no warm-up ramp in the trial ordering' still holds, and the two component p-values (0.198, 0.357) are stated correctly. But the number is wrong by a factor of ~1.8 in the direction that makes the PR's own claim look stronger than it is, and it is the headline statistic for discriminator 1 in the PR body. Everything else in this document reproduced to the digit, which is what makes this one stand out: a reader who spot-checks anything else will find it exact and extend that trust to this. Fix the four occurrences to 0.258 (or 0.241 for the exact permutation null, which is the estimator the rest of discriminator 1 uses and is the more consistent choice).

Not refutable on the scope lens, and the math holds up under independent reproduction. SCOPE: all three cited files are additions in this PR, not pre-existing on main. `git show origin/main:evidence/perf-061/ci-trial-ordering-analysis.txt` returns "fatal: path ... does not exist in 'origin/main'"; `git log origin/main -- evidence/perf-061/` is empty (the directory has never been on main); and `git diff --name-status $(git merge-base origin/main pr2800-check) pr2800-check` marks ci-trial-ordering-analysis.txt, gb10-warmup.md and findings.json all as `A` (added). The merge-base boundary does not apply. MATH: I independently enumerated all 5040 orderings without scipy and got A rho=0.3929 p=0.197817 and B rho=0.1786 p=0.356548 (matching the PR's 0.393/0.198 and 0.179/0.357), X2=-2*(ln pA+ln pB)=5.3034 (matching the PR's 5.303), and chi2 survival at df=4 via the closed form (1+x/2)e^(-x/2) = 0.25756 -- not 0.445. df=6 gives 0.50553; no df yields 0.445. The three quantities on line 19 (X2=5.303, df=4, p=0.445) are mutually inconsistent on that single line, so this is not the reviewer feeding in different inputs. OTHER REFUTATION ROUTES CLOSED: not a negative control (the surrounding text is a substantive discriminator write-up concluding "There is no ramp"; no "MUST fail" or "SELF-TEST PASSED" framing anywhere near it); grounding `measured` is honest (pasted scipy output, which I reproduced by a second method); blocking is `none`, so spec 1 and spec 7 do not apply; and all four claimed occurrences exist verbatim in the PR head (ci-trial-ordering-analysis.txt:19 "POOLED (Fisher) 

</details>
---

### 2. ramp.sh's 'must be 0' contamination check has an inverted exit status: it exits 1 when clean and 0 when contaminated

`evidence/perf-061/ramp.sh:36` · grounding: **measured** · blocking class: `none`

**Evidence**

```
The final two lines of ramp.sh, which is committed as one of 'the measurement harnesses, exactly as run' (gb10-warmup.md file table):
  35: echo "=== CB-006-OUT lines (must be 0) ==="
  36: cat "$OUT/apr.stdout" "$OUT/apr.stderr" | grep -c "CB-006-OUT"

Because `grep -c` exits 1 when the count is zero, and this is the last command of a script running `set -uo pipefail` with no `-e`, the script's status is the negation of the check it is labelled with. I reproduced the exact construct against a clean and a contaminated fixture:

  --- CLEAN (0 contamination = the GOOD case) ---
  === CB-006-OUT lines (must be 0) ===
  0
  script exit status = 1
  --- DIRTY (1 contamination line = the BAD case
```

**Why it matters** — This is the repo's own named recurring class — CLAUDE.md 'Verification Discipline' rule 1 and the SIGPIPE/pipefail memory entry, whose worst variant is described as 'the INVERTED site ... EPIPE PASSES a safety check'. Here the polarity is inverted by grep -c rather than SIGPIPE, with the same effect. Today the impact is bounded: the count is printed and was read by eye, the headline number (0 CB-006-OUT lines across 66 invocations) is corroborated independently by gb10exp.sh's per-invocation cb column, which captures grep -c's stdout and ignores its status, and nothing in the repo invokes ramp.sh and checks $?. But the file is committed under evidence/ as the reproduction recipe, so the next person who wraps it in CI or an `if ramp.sh; then` gets a contamination gate that fires only when the run is clean. Terminate the script with an explicit comparison, e.g. `n=$(grep -c ... || true); echo "$n"; [ "$n" -eq 0 ]`.

Finding survives on the evidence-quality lens. (1) The quote is verbatim: origin/feat/y3-gb10:evidence/perf-061/ramp.sh is exactly 36 lines, line 4 is `set -uo pipefail` (no -e), and line 36 `cat "$OUT/apr.stdout" "$OUT/apr.stderr" | grep -c "CB-006-OUT"` is the last executed statement, so the script's status IS that pipeline's status. (2) The `measured` mark is honest — I independently reproduced the identical construct against clean and contaminated fixtures and got prints `0`/exit=1 for clean and prints `1`/exit=0 for dirty, matching the pasted output. (3) Not pre-existing: `git show origin/main:evidence/perf-061/ramp.sh` returns "does not exist in 'origin/main'", and `gh pr view 2800` lists ramp.sh as an added file — the whole evidence/perf-061 directory is new in this diff. (4) Not a negative control: there is no "it MUST fail" or "SELF-TEST PASSED" marker near it; gb10-warmup.md:275 files it under "the measurement harnesses, exactly as run" and never discloses or relies on an exit status. (5) Grading is internally consistent — `blocking: none` alongside `grounding: measured` with pasted output, so no spec-1 violation of a blocking_class on an asserted finding. (6) I also verified the finding's own bounding claim rather than trusting it: gb10exp.sh:23 is `local cb; cb=$(... | grep -c "CB-006-OUT")` under `set -uo pipefail` with no -e, so it captures stdout and discards status, and the trailing cb column in gb10-raw-trials.tsv is 0 on every ladder row — the corroboration the reviewer cites is real, so impact is genuinely bounded and was not overstated. The only stretch 

</details>
---

### 3. The dispersion contrast quotes 'ollama CV 0.41-0.75%' while a sibling file in the same PR computes 22.13% for one of the two cited runs

`evidence/perf-061/gb10-warmup.md:217` · grounding: **measured** · blocking class: `none`

**Evidence**

```
Claimed:
  gb10-warmup.md:217  - apr CV 12.7-15.0% vs ollama CV 0.41-0.75%, same box, minutes apart.
  findings.json:90    "evidence": "apr CV 12.7-15.0% vs ollama CV 0.41-0.75%, same box, minutes apart. ..."

Computed by the PR itself, one file over:
  ci-trial-ordering-analysis.txt:24  A 33292383055 04:23  apr     n=7 ... CV=14.00%
  ci-trial-ordering-analysis.txt:25  A 33292383055 04:23  ollama  n=5 min= 100.85 max= 182.90 median= 182.35 range= 45.0%  CV=22.13%

I reproduced all four CVs from the committed trial series: apr A 14.00%, ollama A 22.13%, apr B 15.02%, ollama B 0.41%. The local session's ollama CV is 0.75% (median 182.86), also reproduced. So the stated 0.41-0.75% range covers
```

**Why it matters** — The section this line anchors ('What the harness does get wrong — noise, not bias') rests on a two-sided dispersion contrast: apr's estimator is intrinsically noisy, ollama's is intrinsically tight. That conclusion is defensible and I believe it is correct — the 22.13% is driven entirely by one runaway trial (100.85), and the PR does disclose that separately in side finding B. But the range as written reads as complete and is not, and a reader who reads only this section gets a contrast that would partly invert for run A. Either widen the range to 0.41-22.1% with the one-line runaway caveat inline, or state the exclusion explicitly ('excluding run A's runaway trial, see side finding B'). Costs one clause and removes the only place in an otherwise scrupulous document where a number was selected rather than reported.

Could not refute; every check confirms it. (1) Not pre-existing: `git show origin/main:evidence/perf-061/gb10-warmup.md` returns "does not exist in origin/main" — all 13 PR files are new. (2) Quotes are verbatim at PR head b74a2b1ad: gb10-warmup.md:217 "apr CV 12.7–15.0% vs ollama CV 0.41–0.75%" and ci-trial-ordering-analysis.txt:25 "ollama n=5 ... CV=22.13%". (3) I recomputed all five CVs by hand from the committed series and each reproduces: ollama A [154.99,182.79,182.90,182.35,100.85] mean 160.776 sd 35.58 -> 22.13%; apr A 14.00%; apr B 15.02%; local ollama [182.94,183.19,182.78,180.25] -> 0.752% median 182.86 (matches gb10-analysis.txt:13-14); local apr cold gate from gb10-raw-trials.tsv -> 12.67%. (4) The asymmetry is real and load-bearing: the apr range 12.7-15.0 is exactly min-max over all three sources, while the ollama range 0.41-0.75 covers only B and local and drops A's 22.13%, the largest value on that side. (5) Run A is in scope of the same bullet list — line 218 uses run A's ollama median (182.35 -> 182.28 = 0.04%) one line below the range that omits run A's ollama CV. (6) Run A is the headline run: line 4 frames the doc on ratio_median 0.579x, line 113 says "The cold arm reproduces CI run 33292383055 exactly: 0.579x", line 172 tabulates 0.579x/0.647x for A/B. (7) Grepping all five evidence files for CV|22.13|0.41|0.75|runaway|outlier|exclud finds no line scoping the range. (8) Not a negative control — no "MUST fail"/"SELF-TEST PASSED" marker; this is prose, not a guard row. (9) Grounding valid: `measured` with primary artifacts pasted verbatim and all derive

</details>

---

noahgift:
Heads-up: #2829 adds a rule to `check_pass_grep_anchored.sh` that this branch's
`evidence/perf-061/ramp.sh` trips. Verified by running the new guard against
this branch's actual file:

```
INVERTED-EXIT evidence/perf-061/ramp.sh:36
              final command: cat "$OUT/apr.stdout" "$OUT/apr.stderr" | grep -c "CB-006-OUT"
              the script EXPORTS this grep status as its own:
                  input with NO match -> exit 1
                  input WITH a match  -> exit 0
```

Line 36 is the script's final command, so `ramp.sh` returns grep's status:
**exit 1 when the log is clean, exit 0 when it is contaminated** — the opposite
of the `=== CB-006-OUT lines (must be 0) ===` banner one line above. A caller
checking `rc` gets the inverted answer.

One-line fix, and it makes the intended polarity explicit:

```bash
n=$(cat "$OUT/apr.stdout" "$OUT/apr.stderr" | grep -c "CB-006-OUT")
echo "$n"
[ "$n" -eq 0 ] || exit 1
```

(or just `exit 0` if the count is only meant to be read by a human). Whichever
lands first, the other will need this.

noahgift:
Re-verified after rebasing #2829 onto `385db8a02`: unchanged. Rule 2 run against all three of this branch's evidence scripts at `18474b10d` (`ramp.sh`, `gb10exp.sh`, `matched.sh`) finds exactly one hit, still `evidence/perf-061/ramp.sh:36` — the other two end in `echo` and are clean. The fix in the comment above still applies as written.

noahgift:
PP-066 triage (epic #2873, ticket #2986): **CARRY-0.67:W-G**. Linked issue #2786 sits in the carried 0.67 row W-G (issue #2942, `docs/specifications/pp-066-dag.yaml`). Converted to draft and labelled `pp-066/carry-0.67`; not closed. Evidence stays attached to #2786 until W-G opens.

noahgift:

No activity since 2026-09-17 13:33Z; **2 release trains have been cut since** (v0.68.2, v0.68.1). Age is the only input to this sweep — it is not a judgement on the work, and a train label such as `pp-066/carry-*` is not an exemption.

noahgift:
**Triage disposition, cop, 2026-09-20: moved 0.71 → 0.70.**

**0.70's theme is "verbs that are fast"** — #3602's batched-prefill correctness fix, the TTFT ratchet built on #3604's before/after receipts, and decode parity against llama.cpp by reference. **This PR's subject is that theme**, and it was sitting two trains out from the release that will actually do the work.

Measured today and relevant to this row, so it is not re-derived from scratch:

- **apr decode is ~370 tok/s against llama.cpp's 179–189** on the same box and file — roughly **2× faster**. The gap is entirely **pre-generation**.
- **TTFT is 14–16 s against 1.8–2.9 s**, and **67 % of it is `f2_validate_qwen35`** — a full **CPU reference forward** run before the user's first token, on every invocation, saturating at ~9.5 s because `QWEN35_F2_PROBE_MAX = 64`. Split measured: **90–93 % CPU reference, 6–9 % GPU probe** (#3604 caches it).
- On the **dense** path, `--gpu` is currently **2–3× slower** than `--no-gpu` — the GPU result fails its own cosine check (`0.4153@1`) and falls back after the user has paid (#3602).

**What that changes for this PR:** any prefill/decode ratio measured before #3604 lands is measured in a regime that is about to move by ~9 s. **State which regime the numbers were taken in** — that is the difference between a result and a snapshot, and it is the caveat that will be invisible to a reader six weeks from now.

Milestone moved; no judgement on the work itself, which nobody has reviewed in twenty days.

noahgift:

PR #2801, opened 2026-08-31
perf(serve): batched CPU prefill — apr prefilled at decode rate (#2787)

### LINKED ISSUE #2787: apr prefills at decode rate on CPU: no batched prefill outside src/cuda/, 6.5x slower than llama.cpp on the same box
Found by PERF-060 while calibrating `intel` and `mini` for the measurement campaign.

**apr prefills at decode rate on CPU.** Measured prefill-to-decode ratio: **1.07× on `intel`, 1.02× on `mini`**. A GEMM-based prefill should be many times faster than token-by-token decode; these are indistinguishable.

`crates/.../generate_quantized.rs:568` loops `forward_single_with_cache` over **every prompt token**. Every batched `prefill_*` in the tree lives under `crates/aprender-serve/src/cuda/`. So there is no batched prefill on the CPU path at all — it is decode, run once per prompt token.

## The comparator gap, on the same box and the same file

| | prefill tok/s |
|---|---|
| apr, `intel` | **8.14 – 8.97** |
| llama.cpp, `intel`, same GGUF | **53.8 – 57.5** |

Roughly **6.5× slower**. Recorded here as a host-capability reference, **not** as an Arm B number — that llama.cpp build is the wrong pin (`version: 1 (35bee03)`) and cannot be an oracle.

## What it costs the campaign

Every replicate measured, W1 shape (513-token prompt, 128 generated, c=1):

| host | prefill | decode | `t_req` |
|---|---|---|---|
| `intel` (Xeon W-3245, load 19–24) | 8.14 – 8.97 | 6.57 – 7.66 | **73.9 – 81.9 s** |
| `mini` (M4) | 2.371 – 2.372 | 2.318 | **270.9 – 271.1 s** |

`evidence/perf-054/campaign-scope.md` assumed 40/6 and 60/8 tok/s. **Prefill is 4.8× optimistic on `intel` and 25.3× on `mini`**; `t_req` is 2.3× and 11.1× off. `intel`'s apr-lane cell goes from an estimated 8.9 h to **19.2 – 21.3 h**.

Note the decode estimate for `intel` (6 assumed vs 6.57–7.66 measured) was the one figure that held, and was slightly *pessimistic*. The bandwidth model of the hardware was sound. The error was assuming apr prefills with a GEMM.

## Recommendation

**Land batched CPU prefill before scheduling either measurement slot.** It takes `intel`'s cell to roughly **7 h** — below the original estimate — and it moves `mini` from *unmeasurable* to measurable (see the timeout blocker filed separately: `mini`'s 270 s `t_req` exceeds the protocol's hard 120 s per-request timeout, so every request aborts).

Scheduling the campaign first would spend ~19–21 h on `intel` and produce a guaranteed-empty receipt on `mini`, then have to be redone — and per **I-9 a band may not be re-run to green**, so those runs would be spent.

This is also a product defect independent of the campaign: CPU serving is 6.5× off the comparator on prefill for every user, not only for the benchmark.

noahgift:
**Reproduced before fixing.** On lambda, `origin/main` a866988e4, the W1 model and shape (qwen2.5-coder-7b-instruct-q4_k_m.gguf, 513-token prompt, 128 generated):

```
prefill 8.609 tok/s   decode 7.760 tok/s   ratio 1.109   t_req 76.1 s
```

apr prefilled at decode rate. That reproduces #2787 on a third host (the issue records 1.07x on intel, 1.02x on mini) and it is the signature of a missing GEMM, not slow hardware.

## Mechanism

`generate_quantized.rs` looped `forward_single_with_cache` over every prompt token, and every batched `prefill_*` in the tree lived under `src/cuda/`. There was no batched CPU matmul to call: `fused_matmul_k_quants` looped matvec for `seq_len > 1`, and `prefill_batch`'s own comment said as much. A 7B Q4_K_M model streams ~4.4 GB of weights per matvec sweep, so prefill paid that **once per prompt token**.

## What this does

`quantize::batched_matmul` multiplies n activation rows against one weight tensor with the parallel axis still on the output dimension, so a rayon task loads a weight row once and reuses it across the chunk. `forward_prefill_batched` runs the prompt in 128-token chunks.

Attention is deliberately **not** batched — each query still calls `attention_with_cache_gqa_into` against the cache prefix in position order.

Refs #2787, APR-PERF-GATE-001 (#2706)

no-close: #2787 is cited as context for the measurement; this PR reports a result and does not close it.


noahgift:

```
DEGRADED: pmat index d6c6c6f is not an ancestor of head 62b7ee9b (71 commits behind origin/main), and the pmat MCP refused auth — duplication was checked by git grep over origin/main, not by the semantic index | pr-review v2.0.0 | verdict=DEGRADED | consultations: pmat=consulted cuda=no-authority-found crux=consulted mutation=consulted | findings=5 (cited=0 measured=5 asserted=0) | index=d6c6c6f ancestor=false
```

**5 findings raised, 1 survived adversarial refutation** (three independent verifiers per finding — correctness / scope / evidence-quality lenses, refute-by-default). Refuted findings are dropped, not shown.

---

### 1. Three of five prefill falsifiers cannot detect a deleted activation function — the synthetic model's lm_head saturates to 16 distinct logit values

`crates/aprender-serve/src/gguf/inference/forward/falsify_batched_prefill.rs:162` · grounding: **measured** · blocking class: `none`

**Evidence**

```
MUTATION (scratch worktree, PR head 62b7ee9bc): commented out the SwiGLU activation in the batched FFN — prefill_batched.rs:407 `// MUTANT: ops::silu(g);` — which makes the batched path compute a mathematically different function.

  test ...falsify_batched_prefill_kv_cache_is_bitwise_identical ... FAILED
  test ...falsify_batched_prefill_logits_are_bitwise_identical ... ok
  test ...falsify_greedy_continuation_emits_the_same_tokens ... ok
  test ...falsify_chunk_size_does_not_change_the_answer ... ok
  test result: FAILED. 7 passed; 1 failed; 0 ignored; 15782 filtered out

Only the KV-cache test fired. PROBE of the reference logits on covered_model():
  PROBE len=512 distinct_bit_patterns=1
```

**Why it matters** — The property the PR is named for — "the logits it returns for the last prompt token are BITWISE the logits the per-token loop returns" — is asserted by a test that survives deletion of the FFN activation. The suite as a whole is not defenceless (falsify_batched_prefill_kv_cache_is_bitwise_identical did fire, and my RoPE mutation turned three tests RED), so this is not an unprotected change. But the protection rests on one test, and the three logits-based ones read as independent confirmations while contributing almost no discrimination. The authors identified the saturation and routed one discrimination case around it; the three falsifiers were left pointed at it. Making covered_model()'s lm_head non-saturating (or asserting on the hidden state instead of the logits) would convert three near-inert assertions into real ones.

I could not refute it on any of the five refutation grounds, and the source independently predicts the reviewer's measured numbers.

1. NOT true of main. `git show origin/main:crates/aprender-serve/src/gguf/inference/forward/falsify_batched_prefill.rs` → "fatal: path ... does not exist in 'origin/main'". The whole file is +375/-0 in PR #2801 (`gh pr view 2801 --json files`). This is code the diff introduces.

2. NOT a negative-control misread. The file's deliberate negative controls are the two `discrimination_*` tests ("these stay GREEN when the falsifiers above are RED"). The finding does not touch them — it names `falsify_batched_prefill_logits_are_bitwise_identical`, `falsify_greedy_continuation_emits_the_same_tokens`, and `falsify_chunk_size_does_not_change_the_answer`, all three of which are genuine falsifiers whose stated job (module doc, items 2 and 3) is exactly to go RED when the batched path computes something different.

3. Grounding is legitimately `measured`, and the mutation provably ENGAGED. Both mutation runs paste real `cargo test` summaries, and in each run at least one test went RED (`..._kv_cache_is_bitwise_identical` under the silu mutation; 3 failures under the RoPE mutation) — so the edit was compiled and did change behaviour. That rules out the usual "mutation never rebuilt" false positive. `blocking: none`, so spec 1's "blocking_class on an asserted finding" trap does not apply.

4. Both cited mutation sites are real and correctly located at the PR head 62b7ee9bc. prefill_batched.rs:407 is `ops::silu(g);` inside `prefill_ffn_activate`; prefill_batc

</details>

---

noahgift:
PP-066 triage (epic #2873, ticket #2986): **CARRY-0.67:W-G**. Linked issue #2787 sits in the carried 0.67 row W-G (issue #2942, `docs/specifications/pp-066-dag.yaml`). Converted to draft and labelled `pp-066/carry-0.67`; not closed. Rebase when W-G opens (blocked by W-B-kernel).

noahgift:

No activity since 2026-09-17 13:33Z; **2 release trains have been cut since** (v0.68.2, v0.68.1). Age is the only input to this sweep — it is not a judgement on the work, and a train label such as `pp-066/carry-*` is not an exemption.

noahgift:
**Triage disposition, cop, 2026-09-20: moved 0.71 → 0.70.**

**0.70's theme is "verbs that are fast"** — #3602's batched-prefill correctness fix, the TTFT ratchet built on #3604's before/after receipts, and decode parity against llama.cpp by reference. **This PR's subject is that theme**, and it was sitting two trains out from the release that will actually do the work.

Measured today and relevant to this row, so it is not re-derived from scratch:

- **apr decode is ~370 tok/s against llama.cpp's 179–189** on the same box and file — roughly **2× faster**. The gap is entirely **pre-generation**.
- **TTFT is 14–16 s against 1.8–2.9 s**, and **67 % of it is `f2_validate_qwen35`** — a full **CPU reference forward** run before the user's first token, on every invocation, saturating at ~9.5 s because `QWEN35_F2_PROBE_MAX = 64`. Split measured: **90–93 % CPU reference, 6–9 % GPU probe** (#3604 caches it).
- On the **dense** path, `--gpu` is currently **2–3× slower** than `--no-gpu` — the GPU result fails its own cosine check (`0.4153@1`) and falls back after the user has paid (#3602).

**What that changes for this PR:** any prefill/decode ratio measured before #3604 lands is measured in a regime that is about to move by ~9 s. **State which regime the numbers were taken in** — that is the difference between a result and a snapshot, and it is the caveat that will be invisible to a reader six weeks from now.

Milestone moved; no judgement on the work itself, which nobody has reviewed in twenty days.

noahgift:

Issue #2790, opened 2026-08-30
apr run --gpu on the 7B silently falls back to CPU after the F2 parity gate rejects at cosine 0.9278: any tok/s from that path is a CPU number with a GPU label


noahgift:
Found by PERF-059 while root-causing the batched-decode garbage.

**`apr run --gpu` on the 7B has never been a CUDA run.** The F2 parity gate rejects the CUDA path at **cosine 0.9278 (position 46)** on 3 of 3 runs, and the process falls back through wgpu to CPU. The 1.5B does not exhibit this.

So **any 7B tok/s figure taken from that path is a CPU number wearing a GPU label.**

This is the provenance failure this epic exists to remove, in a new place. It is the same family as:

- **#2754** — `tokenization.method` asserting `server_usage` over client-counted tokens
- **#2779** — `wgpu = ["inference"]` reporting a GPU backend it does not enable
- **#2780** — `provenance.model` echoing a clap default, so a 0.5B run passed as `default`

Each is a field or a label that describes an intent rather than what happened, with nothing cross-checking it. Here the label is the `--gpu` flag itself: the user asks for GPU, the parity gate silently declines, and the run reports throughput as though the request had been honoured.

## Why the fallback is the defect, not the rejection

The F2 gate rejecting a numerically-divergent CUDA path is **correct behaviour** — that is what it is for, and given #2767/#2770/#2789 it is rejecting for good reason. The defect is that the fallback is **silent and unrecorded**: nothing in the output, and nothing in a receipt, distinguishes "ran on GPU" from "asked for GPU, was refused, ran on CPU".

Compare **I-2 (resolved-versus-requested)**, which this epic already applies to `--gpu-layers`: a run must record what it *resolved*, not what was *requested*. The same rule plainly applies here and is not being enforced.

## Suggested

1. **Make the fallback loud.** A refused accelerator request should be a stated condition — visible on stdout and carried in the receipt — not a quiet downgrade. `ensure_accelerator_available()` already exists for the explicit-`--gpu` case; this is the parity-gate path reaching the same outcome by a different route.
2. **Record resolved-vs-requested for the compute path**, exactly as I-2 requires for layer counts. A receipt whose `compute_class` says `cuda` for a run that executed on CPU is unfalsifiable provenance, and `bench_receipt.py` would validate it.

[...]

Issue #2780, opened 2026-08-29
provenance.model is the join key that makes cross-host comparison 'impossible to express', and neither end validates it: a 0.5B run passed as 'default'


noahgift:
Found by PERF-055 while producing the first receipt the gate's real mode has ever accepted.

**The first real receipt read `"model": "default"` and the gate said PASS.**

`provenance.model` is the join key whose own docstring says cross-host comparison *"should be impossible to express, not merely discouraged."* **Neither end validates it.**

- **The producer** copies the operator's `--model` flag straight through, and clap's default for that flag is the literal string `default`. Present on `feat/r1-2719` and on `feat/v7-receipt` line 223 — it is not among v7's three provenance fixes (#2754/#2755/#2756).
- **The server does not constrain it.** A run serving a 0.5B Qwen answered `"model":"llama3-70b-there-is-no-such-model"` without complaint.
- **The validator checked absence only**, recording it through `_join_key_incomplete` — written at line 56 and **read by zero other lines in the repo**. A value that is present and meaningless passed.

So the field that exists to make a cross-host comparison unexpressible would happily join a 0.5B run on one host to a 70B run on another, and every schema check would pass.

Same family as #2754 (`tokenization.method` asserting `server_usage` over client-counted tokens) and #2779 (`wgpu` reporting a backend it does not enable): a provenance field nothing cross-checks against what actually happened. This one is worse because it is the **join key** — it is what makes two receipts comparable at all.

## Fixed in PERF-055 (`808f1a9b2`)

Present-and-meaningless is now an error. Absence stays reported-not-failed, which is the documented #2696 deferral that `00_pristine.json` depends on.

Measured on a real artifact rather than a fixture:

```
placeholder + unfixed gate  -> PASS
placeholder + fixed gate    -> FAIL
honest receipt + fixed gate -> PASS
```

Case floor 14 -> 18, each mutation-verified, including an **over-reach** discrimination case (make everything a placeholder) that turns `15_join_key_real_values` RED — without it the rule could be "reject every join key" and still read green.

## Still open, and it is the real fix

The producer should **read `provenance.model` from the server** rather than echo the operator's flag. `GET /api/tags` returns the loaded model; `/v1/models` returns only the echoed placeholder. Until then the guard catches obvious placeholders but cannot catch an operator who passes a plausible wrong name — and per I-9 a band may not be re-run to green, so a mislabelled cell is a spent run.

noahgift:
[...]

PR #2803, opened 2026-08-31
fix(perf-gate): two provenance fields recorded intent, not outcome (#2790, #2780)

noahgift:
#2790 and #2780 are the same defect twice: **a receipt field that repeats what the operator asked for instead of what the machine did.**

This is the enforcement-integrity class — a provenance field recording intent cannot falsify anything, because it agrees with the request by construction.

## Part 1 — `apr run --gpu` on the 7B has never been a CUDA run (#2790)

Measured on lambda-vector, apr built from a866988e4 with `--features cuda`. The receipt recorded the requested device, not the resolved one, so a silent CPU fallback was indistinguishable from a GPU run in every receipt ever produced.

## Part 2 — the resolution line

Adds `scripts/check_compute_resolution_line.sh` and `scripts/perf-receipt-fields.yaml`, so a receipt must carry what was **resolved**, and a run that fell back is visible as a fallback.

Relates to the recorded I-2 rule: a run must record what it resolved, not what was requested.

Refs #2790, #2780, APR-PERF-GATE-001 (#2706)


noahgift:

```
DEGRADED: pmat index d6c6c6f is 71 commits behind main and NOT an ancestor of the PR head (MCP ConnectionRefused; pmat CLI 3.34.0 used instead) | pr-review v2.0.0 | verdict=DEGRADED | consultations: pmat=consulted cuda=consulted crux=consulted mutation=consulted | findings=6 (cited=1 measured=5 asserted=0) | index=d6c6c6f ancestor=false
```

**6 findings raised, 3 survived adversarial refutation** (three independent verifiers per finding — correctness / scope / evidence-quality lenses, refute-by-default). Refuted findings are dropped, not shown.

---

### 1. compute_resolution.rs cites `scripts/check_compute_class_vocabulary.sh` as the guard that catches vocabulary divergence; that script has never existed

`crates/aprender-serve/src/infer/compute_resolution.rs:57` · grounding: **measured** · blocking class: `none`

**Evidence**

```
Source claim (compute_resolution.rs:53-57):

    /// The token set is `COMPUTE_CLASSES` in `scripts/lib/bench_receipt.py`. It is
    /// duplicated here rather than derived because a receipt schema and an
    /// inference engine cannot share a type across a language boundary — but the
    /// guard `scripts/check_compute_class_vocabulary.sh` fails when they diverge.

Measured:

    $ git ls-tree -r origin/feat/y6-provenance --name-only | grep -i "compute_class_vocab"
    (no output)
    $ git log --all --oneline --diff-filter=A -- 'scripts/check_compute_class_vocabulary.sh'
    (no output)
    $ git show origin/feat/y6-provenance:.github/workflows/ci.yml | grep -in compute_class_vocabulary

```

**Why it matters** — The file admits it is creating a second definition of a shared vocabulary and justifies that by naming a guard which does not exist. The one test covering it is a THIRD copy: `const RECEIPT` is a frozen literal inside the same crate, so it fires only if the Rust enum gains a token — it cannot fire if scripts/lib/bench_receipt.py's COMPUTE_CLASSES changes, which is the only divergence the comment claims is guarded. This is the repo's own "assertion that cannot fire" class, introduced by a PR about enforcement integrity. Either write the guard (read COMPUTE_CLASSES out of bench_receipt.py and diff it against the enum) or delete the sentence that says one exists.

Verified correct on all counts against origin/feat/y6-provenance (working tree not used). (1) The claim is in the diff: compute_resolution.rs:53-57 says "the guard `scripts/check_compute_class_vocabulary.sh` fails when they diverge". (2) That script has never existed: `git grep -n check_compute_class_vocabulary origin/feat/y6-provenance` returns exactly one hit — the doc comment itself; `git log --all --diff-filter=A -- scripts/check_compute_class_vocabulary.sh` is empty; a `git cat-file -e "$b:scripts/check_compute_class_vocabulary.sh"` loop over every remote branch found it on none (so it is not pending in a stacked branch); ci.yml wires only scripts/check_compute_resolution_line.sh (lines 761/763). (3) The similarly-named real guard is a different guard: check_compute_resolution_line.sh is a case table over apr_class_from_log() in parity_host_receipt.sh (log classification) and never touches the Rust enum — `git grep "ComputeClass\|compute_resolution.rs" origin/feat/y6-provenance -- scripts Makefile .github` returns ZERO hits, and no script reads COMPUTE_CLASSES (the only tree-wide hits are bench_receipt.py:31,117,119 plus Rust prose). (4) The test genuinely cannot fire on the divergence claimed: compute_resolution.rs:502 hardcodes `const RECEIPT: [&str;5]` and asserts each Rust wire() is IN it, so removing/renaming a token in bench_receipt.py's COMPUTE_CLASSES leaves it green; receipt.rs:655 is a second hardcoded copy of the same tuple. (5) NOT pre-existing — `git show origin/main:crates/aprender-serve/src/infer/compute_resolution.rs` errors with "does not exist in 'ori

</details>
---

### 2. The guard is not mechanically linked to the Rust that writes the line it parses, and the Rust test that would pin the format is tautological

`scripts/check_compute_resolution_line.sh:1` · grounding: **measured** · blocking class: `none`

**Evidence**

```
MUTATION RUN (8 mutations applied to the extracted producer `apr_class_from_log` in scripts/parity_host_receipt.sh, guard re-run after each; producer restored and diff-verified clean after every round):

  M1 revert the refusal-exclusion (the historical defect)   rc=1 failrows=5   KILLED
  M2 tail -1 -> head -1 (last-wins broken)                  rc=1 failrows=2   KILLED
  M3 drop metal from the case allowlist                     rc=1 failrows=2   KILLED
  M4 drop cuda from the case allowlist                      rc=1 failrows=3   KILLED
  M5 remove the resolution-line read entirely               rc=1 failrows=6   KILLED
  M6 drop the ^ anchor on the resolution line               rc=0 failro
```

**Why it matters** — The guard's own header states the rule it then breaks one level up: "The function under test is EXTRACTED from parity_host_receipt.sh rather than copied here. A copy is a second definition, and a second definition is how the two band schemas happened (PERF-004)." The LINE FORMAT is exactly such a second definition — a Rust `format!` and a shell `sed -n 's/^apr-compute: .*resolved=\([a-z]*\).*/\1/p'` — with nothing tying them. Change RESOLUTION_LINE_PREFIX and: the Rust test still passes (it compares the constant to itself), the shell guard still passes (all 17 rows are frozen .txt/heredoc text, not produced by the binary), and the live pipeline silently falls back to banner classification. The cheap fix is one row that greps the constant out of compute_resolution.rs and asserts the sed matches a line built from it. M6 is the same gap in miniature and is low severity on its own.

The core technical claim verifies true on concrete inspection of the PR head, and every refutation angle fails.

NOT PRE-EXISTING: `git show origin/main:scripts/check_compute_resolution_line.sh` and `origin/main:crates/aprender-serve/src/infer/compute_resolution.rs` both return "does not exist". `apr_class_from_log()` on main is the banner-only a866988e4 form with no resolution-line reader. The coupling is introduced by this diff.

THE TAUTOLOGY IS EXACT: compute_resolution.rs:252 builds the line as `format!("{RESOLUTION_LINE_PREFIX} requested={} resolved={} ...")` and line 447 asserts `line.starts_with(RESOLUTION_LINE_PREFIX)` — the same const on both sides, so it holds for any value. A full-branch `git grep "apr-compute"` finds the literal in exactly ONE Rust location: the const definition (line 50). No Rust test pins the literal.

NO MECHANICAL LINK: parity_host_receipt.sh:91 independently re-types `sed -n 's/^apr-compute: .*resolved=\([a-z]*\).*/\1/p'`. `git grep -l RESOLUTION_LINE_PREFIX -- scripts .github contracts` returns 0 files. ci.yml lines 761/763 wire the guard as an explicitly "Text-only check, no build" — it reads only parity_host_receipt.sh and evidence/perf-062/*, never executing the Rust. Extra corroboration the finding did not use: compute_resolution.rs justifies the cross-language vocabulary duplication by citing "the guard `scripts/check_compute_class_vocabulary.sh` fails when they diverge" — that script exists on neither the branch nor main. The only cross-language pin the code claims is absent.

NOT A NEGATIVE-CONTROL MISREAD: I reran both entry point

</details>
---

### 3. Two published, literal-constructible public structs gain required fields with no semver signal in this PR

`crates/aprender-serve/src/infer/inference_result.rs:28` · grounding: **measured** · blocking class: `none`

**Evidence**

```
InferenceResult is `pub`, has all-public fields, and no #[non_exhaustive]:

    /// Result from inference
    #[derive(Debug, Clone)]
    pub struct InferenceResult {
        ...
        pub used_gpu: bool,
        pub compute: ComputeResolution,      <-- added
    }

Same for InferenceConfig (infer/mod.rs:43): `pub struct InferenceConfig` gains `pub compute_request: ComputeRequest`, no #[non_exhaustive].

Measured blast radius inside the tree — the signature of a constructible-struct break:
  $ git diff origin/main...origin/feat/y6-provenance | grep -cE "^\+ +compute: (Default::default\(\)|ComputeResolution)"
  69
  $ git diff origin/main...origin/feat/y6-provenance --name-only -G "compute:
```

**Why it matters** — 69 one-line struct-literal fixups across 22 in-tree files are the direct measurement that external `InferenceResult { .. }` / `InferenceConfig { .. }` construction breaks; realizar (aprender-serve) is a published crate, so this is cargo-semver-checks' `constructible_struct_adds_field`. I am deliberately NOT setting blocking_class here even though §7 lists "breaking API surface with no semver bump": this repo bumps all 78 crates once at release-cut time via scripts/bump-version.sh, so "no version change in the PR" is the normal state of every PR that touches a pub struct, and a class that fires on nearly every PR cannot meet §7's ≥90% precision admission rule. The actionable item is smaller than a block: mark both structs #[non_exhaustive], or record that 0.64.0 -> 0.65.0 must be a 0.x-breaking minor, so the signal is not left to the next release cut to rediscover.

Survives on my scope lens and on every other refutation criterion. (1) NOT true of main: `git show origin/main:crates/aprender-serve/src/infer/inference_result.rs` shows InferenceResult terminating at `pub used_gpu: bool,` with no `compute` field, and `git show origin/main:crates/aprender-serve/src/infer/mod.rs | grep -n compute_request` returns zero hits. The PR head (f9fe389d, confirmed identical to refs/pull/2803/head) adds `pub compute: ComputeResolution` at exactly line 28 of inference_result.rs — the cited anchor — and `pub compute_request: ComputeRequest` at mod.rs:77, plus a new `with_compute_request` builder at 207. Both fields are introduced by this diff. (2) The `measured` grounding is real and reproduces byte-for-byte against merge-base a866988e: the two greps return 69 and 22, and both Cargo.toml version lines return "0.64.0". My first run returned 0/0 and 0.61.0 only because my local FETCH_HEAD had resolved to a stale adf16c00; re-fetching origin/main + the PR head made the finding's exact commands reproduce verbatim. (3) Premises hold: `grep -c non_exhaustive` on the PR-head file returns 0, all fields are `pub`, crates/aprender-serve/Cargo.toml carries no `publish = false` and declares `[lib] name = "realizar"` with `documentation = "https://docs.rs/realizar"`, and both structs are in the published surface twice over via an un-cfg-gated `pub mod infer;` (lib.rs:328) and `pub use infer::{… InferenceConfig, InferenceResult}` (lib.rs:458-461) — so `constructible_struct_adds_field` genuinely applies. (4) The two ci.yml `semver` matches are a `pv --version` string

</details>

---

noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:R-0 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `crates/apr-cli/src/commands/test_llm_band.rs scripts/parity_host_receipt.sh`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Linked issue #2790 is in R-0's blocker list (R-0 expires 2026-09-19); also intersects R-0's dispatch.rs. Must land before R-0's worktree opens. Also edits .github/workflows/ci.yml. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2803` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2803@53bc2bf32 · intent → R-0 (#2904; its blocker list holds #2790) · salvage: crates/apr-cli/src/commands/test_llm_band.rs crates/aprender-serve/src/api/tests/{chat_completion_03,completion_request_response,deep_apicov_02,format_chat_02,imp_140c,model_metadata}.rs crates/aprender-serve/tests/{api_coverage,api_deep_coverage,infer_deep_coverage,property_api}.rs · conflicts: crates/apr-cli/src/commands/test_llm_band.rs scripts/parity_host_receipt.sh`

PR #2804, opened 2026-08-31
fix(ci): sweep the remaining 103 EPIPE-inverting `producer | grep -q` sites — 32 of them fail SILENTLY OPEN

noahgift:
Follow-up to #2799 (which fixes the two sites that took `guard-runner-labels` red on #2776). Same defect, 48 more files, 103 more sites.

`grep -q` exits at its FIRST match and closes the pipe. If the producer still has a `write()` to perform it takes EPIPE, and under `set -o pipefail` the pipeline reports the **producer's** 141/1 rather than grep's 0 — so **a match is read as a non-match**.

## The universe

Re-derived from `origin/main` with `git show`, not the checkout (the checkout is 50+ commits behind on an old branch and does not even contain `check_verifier_pinning.sh`). **129 raw hits across 58 files** in `*.sh`, `.github/workflows/*.yml`, `Makefile`. Of those:

| | |
|---:|---|
| 23 | **not live code** — 20 comments documenting this exact hazard, 3 fixture strings inside guard case tables |
| 1 | **not a pipe** — `[ -f x ] \|\| grep -q p FILE` is a `\|\|`, not a `\|` |
| **105** | **live pipelines** — the real universe |

## Polarity — the inverted half is the dangerous one

EPIPE only ever turns a **match** into a read-as-non-match. So the damage is decided by what a match *means* at each site:

| count | class | what a match means | what EPIPE does |
|---:|---|---|---|
| 57 | **NORMAL** | "expected thing is present" | fabricates an absence → **false RED**. Noisy, self-announcing. This is what hit #2776. |
| **32** | **INVERTED** | "violation found" | drops the detection → **silent false GREEN**. A safety check that passes. |
| 16 | BRANCH | capability probe / retry / diagnosis | degrades; no verdict flips |

**The 32 inverted sites**, by file:

`check-simd-target-feature.sh:110` · `install-hooks.sh:73` · `sync-models_test.sh:91,96` · `validate-book-code.sh:58,67` · `chaos-baseline.sh:89,107` · `book-contracts.yml:148` · `book-gate.sh:183` · `check_book_examples_executable.sh:78` · `check_cargo_install_private_root.sh:78,93,103,107,234` · `check_cascade_covers_all_crates.sh:90` · `check_contract_test_binding.sh:248` · `check_dogfood_shim.sh:366,368` · `check_format_sovereignty.sh:66` · `check_multiplatform_dogfood.sh:224` · `check_pass_grep_anchored.sh:123,129` · `check_publish_safety.sh:45` · `check_pv_version_parse.sh:91` · `check_readme_claims.sh:291` · `check_runner_labels.sh:24` · `dogfood.sh:952,955` · `qwen-story.sh:301` · `runner-disk-guard.sh:85`

Five are the same sub-class — `producer | grep -q X || continue`, where a lost match silently **removes an item from the guard's own universe**: `check_cascade_covers_all_crates.sh:90`, `check_dogfood_shim.sh:366` and `:368`, `check_pass_grep_anchored.sh:123`, `check_runner_labels.sh:24`.

## Measurement — the size story was wrong in both directions

The governing variable is **not** payload size. It is whether the producer still has a `write()` pending when grep exits. Two ways that happens:

**(a) burst producer, more than ~96 KB left to write after an early match**

| bytes | 32768 | 65536 | 98304 | 131072 | 262144 | 1048576 |
|---|---|---|---|---|---|---|
| match **EARLY** | 0/200 | 0/200 | 3/200 | **200/200** | **200/200** | **200/200** |
| match **LATE** | 0/200 | 0/200 | 0/200 | 0/200 | 0/200 | 0/200 |

**Match position is the variable nobody had held constant.** A large payload with the needle at the end never reproduces.

**(b) incremental producer still alive when grep exits — 100% loss at ANY size**

```
{ echo NEEDLE; sleep 0.02; echo tail; } | grep -q NEEDLE   ->  lost 100/100   (30 bytes)
{ echo NEEDLE;             echo tail; } | grep -q NEEDLE   ->  lost   0/100
```

So `apr run … | grep -q`, `curl … | grep -q`, `cargo … | grep -q` are the shapes to fear — not the big ones.

**Is any guard permanently fail-open?** No — and I checked rather than assumed. Largest producer in the tree is 40346 bytes (`grep -v` over `dogfood.sh`); every producer writes in one burst; every real needle sits late. Measured **0/300** on all five files that actually match at `check_dogfood_shim.sh:366`, and **0/40** on `cargo package --list | grep -q` (28143 bytes) at `check_publish_safety.sh:45`. These are stochastic fail-opens (~1/20000 at load 13.6), not permanent ones — but a false GREEN is unobservable, so it accrues in silence while a false RED announces itself.

## Before / after

Command-producer class (the 34 sites converted to capture-then-here-string):

```
BEFORE  cmd | grep -q            lost: 200/200
AFTER   grep -q <<< "$(cmd)"     lost:   0/200
```

## Verification

**14 guard self-tests / case tables**, all rc=0, 0 unexpected FAIL rows. (`check_pv_version_parse.sh --self-test` prints 3 FAIL rows by design — byte-identical to `origin/main`; they are the table proving the pre-#2559 extractor turns RED.)

**7 mutations, each RED then GREEN on revert:**

| # | guard | mutation | mutated | revert |
|---|---|---|---|---|
| M1 | `check_runner_labels.sh` | add a self-hosted job with no discriminating label | **rc=1** | rc=0 |
| M2 | `check_cargo_install_private_root.sh` | add a self-hosted `cargo install` with no private root | **rc=1** (`SHARED-INSTALL … job mut-install`) | rc=0 |
| M3 | `check_workflow_path_filters.sh` | drop `crates/aprender-core/src/**` from both filters | **rc=1** | rc=0 |
| M4 | `check_dogfood_shim.sh` | add a SECOND install-sweep runner | **rc=1** (`FAIL 1b`) | rc=0 |
| M5 | `check_guards_are_wired.sh` | unwire `check_runner_labels.sh` from every workflow | **rc=1** (`grew 4 -> 5, NEW: check_runner_labels.sh`) | rc=0 |
| M6 | `check_pass_grep_anchored.sh` | its own built-in proof: plant the PMAT-CI-PASSGREP-001 pattern | **RED** | — |
| M7 | `check_publish_safety.sh` | repoint the needle at a path that IS in the package list | **rc=1** | rc=0 |

M5 took three attempts, and the two failures are worth recording: `DISABLED_check_runner_labels.sh` still contains the substring `check_runner_labels.sh`, and the guard is named in `coverage-nightly.yml` as well as `ci.yml`. Both invalid mutations looked like a guard that could not fire. Only removing the invocation lines from **both** workflows produced the real RED.

Also green: `check_shell_lint_ratchet.sh` (bashrs over all 48 touched scripts), `check_apr_bin_pinned.sh` (164 files), `check_beats_gated.sh`, `check_baseline_ratchets.sh`, `check_no_hand_rolled_parsers.sh`, `check_sourced_libs_option_neutral.sh`, `check_hermetic_stdin_tests.sh`, `check_exclude_anchored.sh`, `check_story_json_streams.sh`, `check_no_timing_in_required.sh`, `check_workflow_env_defined.sh`, `check_duplicate_bin_names.sh`. `bash -n` clean on all 48. Tree-wide re-scan: **0 live sites remain** outside `check_verifier_pinning.sh` (deliberately left to #2799 to avoid a conflict).

Confirmed the sweep did not shrink any guard's universe: `check_pass_grep_anchored.sh` reports the same `2 checked` before and after.

## The fix

- producer is a **variable** → `grep -q PAT <<< "$var"` (no subprocess at all)
- producer is a **command** → capture, then here-string

Never by disabling `pipefail` — that trades a rare false verdict for a permanent blind spot.

Four sites needed hand treatment rather than the mechanical rewrite: `check_cargo_install_private_root.sh:234` and `check_workflow_path_filters.sh:106` (two greps on one line), `cascade-publish.sh:144` / `check_guards_are_wired.sh:97` / `check_cascade_covers_all_crates.sh:90` (multi-line pipelines, captured into a local first), `scientific-swap-benchmark.sh:324-325` (inside a command substitution).

Three of the 48 files are workflows (`book-contracts.yml`, `cuda-nightly.yml`) — one-line edits inside `run:` blocks, the same construct as everywhere else, no job/trigger/matrix change. `scripts/perf-matrix.yaml` is untouched.

### DIFF
--- a/.github/workflows/book-contracts.yml
+++ b/.github/workflows/book-contracts.yml
@@ -131,7 +131,7 @@ jobs:
         run: |
           FAIL=0
           for f in book/src/chapters/ch*.md; do
-            if ! head -1 "$f" | grep -q "PCU:"; then
+            if ! grep -q "PCU:" <<< "$(head -1 "$f")" ; then
               echo "::error::No PCU frontmatter: $f"
               FAIL=1
             fi
@@ -145,7 +145,7 @@ jobs:
           FAIL=0
           for md in $(find book/src -name "*.md" -not -name "SUMMARY.md"); do
             hdr=$(head -1 "$md")
-            if echo "$hdr" | grep -q 'PCU:'; then
+            if grep -q 'PCU:' <<< "$hdr" ; then
               CONTRACT=$(echo "$hdr" | grep -oE 'contract:[[:space:]]*[^ |]+' | sed 's/^contract:[[:space:]]*//')
               if [ -z "$CONTRACT" ]; then
                 echo "::error::$md has PCU header but no contract: field"
--- a/.github/workflows/cuda-nightly.yml
+++ b/.github/workflows/cuda-nightly.yml
@@ -213,7 +213,7 @@ jobs:
             echo "::error::APR_GGUF not found on gx10: $APR_GGUF"
             exit 1
           fi
-          "$HOME/.local/bin/ollama" list | grep -q "$OLLAMA_MODEL" || {
+          grep -q "$OLLAMA_MODEL" <<< "$("$HOME/.local/bin/ollama" list)" || {
             echo "::error::ollama model not pulled: $OLLAMA_MODEL"
             exit 1
           }
--- a/crates/aprender-compute/scripts/check-simd-target-feature.sh
+++ b/crates/aprender-compute/scripts/check-simd-target-feature.sh
@@ -107,7 +107,7 @@ check_file() {
                 fi
 
                 # Check if function body contains SIMD intrinsics
-                if echo "$fn_body" | grep -qE "$pattern"; then
+                if grep -qE "$pattern" <<< "$fn_body" ; then
                     # Function uses SIMD intrinsics
                     if [[ "$has_target_feature" == false ]]; then
                         # Missing #[target_feature] attribute!
--- a/crates/aprender-orchestrate/scripts/build-wasm.sh
+++ b/crates/aprender-orchestrate/scripts/build-wasm.sh
@@ -28,7 +28,7 @@ echo "================================="
 echo
 
 # Check if wasm32 target is installed
-if ! rustup target list | grep -q "wasm32-unknown-unknown (installed)"; then
+if ! grep -q "wasm32-unknown-unknown (installed)" <<< "$(rustup target list)" ; then
     echo -e "${YELLOW}⚠️  wasm32-unknown-unknown target not installed${NC}"
     echo "Installing..."
     rustup target add wasm32-unknown-unknown
--- a/crates/aprender-present/scripts/install-hooks.sh
+++ b/crates/aprender-present/scripts/install-hooks.sh
@@ -70,7 +70,7 @@ for f in $PTOP_IMPL; do
 done
 
 # Run interface tests for ptop changes
-if echo "$STAGED" | grep -q 'crates/presentar-terminal/src/ptop/'; then
+if grep -q 'crates/presentar-terminal/src/ptop/' <<< "$STAGED" ; then
     echo "Running interface tests for ptop changes..."
     if ! cargo test -p presentar-terminal --features ptop --test cpu_exploded_async --quiet 2>/dev/null; then
         echo ""
--- a/crates/aprender-serve/scripts/perf-gate.sh
+++ b/crates/aprender-serve/scripts/perf-gate.sh
@@ -48,7 +48,7 @@ trap "kill $SERVER_PID 2>/dev/null; wait $SERVER_PID 2>/dev/null" EXIT
 
 # Wait for server
 for i in $(seq 1 30); do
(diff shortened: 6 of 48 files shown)


noahgift:

```
DEGRADED: pmat MCP unreachable, CLI fallback index d6c6c6f is NOT an ancestor of head adf16c0 and covers Rust only (shell/YAML diff unreachable) | pr-review v2.0.0 | verdict=DEGRADED | consultations: pmat=consulted cuda=no-authority-found crux=consulted mutation=consulted | findings=3 (cited=0 measured=3 asserted=0) | index=d6c6c6f ancestor=false
```

**3 findings raised, 0 survived adversarial refutation** (three independent verifiers per finding — correctness / scope / evidence-quality lenses, refute-by-default). Refuted findings are dropped, not shown.

No findings survived verification. Consultations: `pmat=consulted cuda=no-authority-found crux=consulted mutation=consulted`.

---

Issue #2789, opened 2026-08-30
FP8 E4M3 decode fires at m>=5, so CB-006's two named conditions do not hold at m>=6 — and no configuration is correct across bands 1/4/8/16


noahgift:
Found by PERF-059 while root-causing #2785 (since closed as premise-falsified).

**FALSIFY-CB-006's "two named conditions" were only ever tested at m=2 and m=4. They do not hold at m≥6.**

The boundary is one predicate — `fp8_will_fire = gpu_profile.fp8_decode && m >= 5` — at `batched_qkv.rs:119`, `batched_ffn.rs:111` and `cublas_prefill/attention.rs:1020`. It was **predicted from the source before testing, then confirmed**: `FP8_DECODE=0` restores m=6/8/16 exactly and changes nothing at m≤4.

## The ladder, measured on one binary

| STREAM_LEGACY | CUBLAS_THR | FP8_DECODE | c=1 | c=2 | c=4 | c=6/8/16 |
|---|---|---|---|---|---|---|
| unset | 4 (def) | def | ref | – | **GARBAGE** | – |
| unset | 32 | def | – | – | **GARBAGE** | – |
| 1 | 4 (def) | def | ref | – | wrong | – |
| 1 | 32 | def | ref | ref | **ref** (m=4) | wrong |
| 1 | 32 | 0 | GARB | GARB | **ref** (m=4) | **ref** |
| 1 | 4 (def) | 0 | ref | ref | ref (m=3) | wrong |

## The consequence for APR-PERF-GATE-001 (#2706)

**No configuration is correct across bands {1, 4, 8, 16}.** The two working arms are disjoint at the ends — row 4 is correct at c=1/2/4 and wrong at c≥6; row 5 is correct at m=4 and c≥6 and garbage at c=1/2.

**Arm A divides `agg(c)` by `agg(1)`**, so a cell cannot be spliced from two configurations. Every band in a cell has to come from one binary under one set of conditions, and no such set currently exists. That makes **Arm A unmeasurable on the CUDA hosts right now**, independent of the wall-clock and receipt questions.

## Sequencing

[...]

Issue #2767, opened 2026-08-29
legacy-stream cuMemcpy races non-blocking streams: 260 host-transfer sites assume an ordering CUDA does not provide


noahgift:
Root cause of #2753's residue, and it is systemic rather than a slip.

`CudaStream::new` creates this crate's streams with **`CU_STREAM_NON_BLOCKING`**, which CUDA explicitly excludes from legacy default-stream ordering. `GpuBuffer::copy_from_host` and `copy_to_host` are **`cuMemcpyHtoD` / `cuMemcpyDtoH` — legacy-stream transfers**.

A legacy transfer does not order against a non-blocking stream. So every host transfer in the batched decode path races the kernels in flight, while the surrounding code is written throughout as though the transfers were ordered against it.

## Proven by mutation, one binary, one env var

```
CU_STREAM_NON_BLOCKING (default)          SELF cosine 0.517985   FAILS
APR_STREAM_LEGACY=1 (CU_STREAM_DEFAULT)   SELF cosine 1.000000   PASSES
```

SELF runs the batched forward twice on the same input and the same state and requires identical logits. This **restores the invariant** rather than improving a number. SENSITIVE and PERTURBED still diverge under the knob, so the probe has not gone uniformly green.

One mechanism accounts for the entire #2753 thread: the allocation-size dependence (timing), the all-NaN and all-zero logits (a partially written buffer; a fresh allocation reads as zero, which argmaxes to token 0 — the original constant-token garbage), and the first-call/second-call asymmetry.

## The scale

**260 legacy-stream host-transfer sites in `aprender-serve` alone** — 206 `copy_to_host`, 31 `copy_from_host`, 15 `copy_from_host_at`, 8 `copy_to_host_at` — including the pointer arrays, `seq_lens` and `positions` that in-flight kernels are reading.

## Two repairs, and why the default was left alone

**(a) Create the streams with `CU_STREAM_DEFAULT`.** One line. Restores the ordering the code already assumes — and **serialises the overlap**. This crate creates several streams (`compute_stream`, `transfer_stream`, `s1..s3`) deliberately for concurrency, and CORRECTNESS-012b moved work onto `self.stream` precisely for ordering. Flipping them all to the legacy stream serialises them against it.

**(b) Convert the transfers to the stream-ordered async variants that already exist** in `transfer.rs` (`copy_from_host_async`, `cuMemcpyDtoHAsync`), with an explicit sync before each host read. Preserves the overlap. 260 call sites.

[...]

Issue #326, opened 2026-02-24
feat: BERT encoder inference for cross-encoder reranking (.apr)


noahgift:
## Summary

Verify and implement end-to-end BERT encoder inference in aprender, enabling cross-encoder reranking models (e.g., `BAAI/bge-reranker-base`) to run as `.apr` models via trueno SIMD. This is the sovereign-stack alternative to ONNX Runtime / fastembed for neural reranking.

## Motivation

trueno-rag's RAG pipeline achieves MRR 0.952 with semantic hybrid (BGE-small + BM25 RRF). Cross-encoder reranking is the standard next step to push MRR toward 0.97+. Lexical reranking was tested and rejected (regressed to MRR 0.876 — term overlap disrupts semantic ordering).

Rather than adding `ort` (ONNX Runtime) as a dependency, the sovereign approach is:
1. Convert cross-encoder ONNX/SafeTensors → `.apr` via `apr import`
2. Run BERT inference natively via aprender/trueno SIMD
3. Same pattern as whisper-apr (whisper model → `.apr` → pure Rust inference)

### Model Architecture: BERT Cross-Encoder

Cross-encoders are BERT-base models with a classification/regression head:

```
Input:  [CLS] query_tokens [SEP] passage_tokens [SEP]
         ↓
      BERT Encoder (12 layers, 768d, 12 heads)
         ↓
      CLS pooling (extract [CLS] token embedding)
         ↓
      Linear head (768 → 1) → sigmoid → relevance score [0, 1]
```

### Target Models

| Model | Params | Dim | Source |
|-------|--------|-----|--------|
| `BAAI/bge-reranker-base` | 109M | 768 | HuggingFace SafeTensors |
| `cross-encoder/ms-marco-MiniLM-L-6-v2` | 22M | 384 | HuggingFace SafeTensors |

#### 1. BERT Tensor Name Mapping (Architecture::Bert)
Verify `map_name()` handles all BERT tensors:
- `bert.embeddings.word_embeddings.weight`
- `bert.embeddings.position_embeddings.weight`
- `bert.embeddings.token_type_embeddings.weight`
- `bert.embeddings.LayerNorm.{weight,bias}`
- `bert.encoder.layer.{N}.attention.self.{query,key,value}.{weight,bias}`
- `bert.encoder.layer.{N}.attention.output.dense.{weight,bias}`
- `bert.encoder.layer.{N}.attention.output.LayerNorm.{weight,bias}`
[...]

PR #2809, opened 2026-08-31
fix(cuda): arm FALSIFY-CB-008 — the rule that forbade #2753 could never run (#2753)

### LINKED ISSUE #2753: P0: batched CUDA decode emits garbage for every m>1 — PERF-001's 3.32x aggregate is throughput of garbage tokens

noahgift:
## The rule that forbade #2753 could never run

`contracts/continuous-batching-v1.yaml` CB-008 has said **"No frozen slots — all M slots produce distinct tokens per decode step (not constant)"** since v1.0. aprender#2753 is that rule failing in production for four releases: every slot served from a batch emitted **one token id to the `max_tokens` cap** and never reached a stop token. CB-008's `test:` field named a `BATCHED_DECODE_TRACE` log — **the variable did not exist**, so the evidence it pointed at could never have been read.

### Reproduced first, on `origin/main` at its default configuration

RTX 4090 sm_89, `apr 0.64.0 (745fa8588)` == `origin/main`, `qwen2.5-coder-1.5b-instruct-q4_k_m`, greedy, 120-token cap. `[PMAT-044] Batch m=3 done` in the server log proves the batched path engaged:

```
[CB-008] step=0   m=3 token_ids=[151662, 151662, 151662]
...                   the SAME id for steps 0..115
[CB-008] step=119 m=3 token_ids=[98401, 98401, 98401]

c=4:  [0] OK       'Compilers are software programs that tra'   <- m=1 fast path
      [1] DIVERGED '</iang.").'                                 <- from the batch
      [2] DIVERGED '</iang.").'
      [3] DIVERGED '</iang.").'
```

`APR_STREAM_LEGACY=1` in that **same binary** turns the garbage into coherent English. So at HEAD the freeze **is** the #2767 legacy-stream race and nothing else — the three structural defects (#2771) and the Q6_K dequant (#2783) already landed.

## What this carries, and what is new

**Merged unchanged from `feat/v14-streamdefault` (PR #2776 — green, queued, evicted, still open; not my work):** `CU_STREAM_DEFAULT` as the stream flag. That is the fix for the freeze, and CB-008 cannot be green without it. **If #2776 lands first, this reduces cleanly to the commit on top.**

### 1. `realizar::cb008_frozen_slots` — the rule as one pure function

Ten discrimination controls that run in **ci.yml's ordinary workspace `--lib` line, on hosts with no CUDA**. They assert the checker *rejects* the recorded signatures verbatim (token id 0 to the 400-token cap; id 151662 for 116 of 120 steps), rejects a long stall that variety would otherwise excuse, treats a too-short stream as a **failure** rather than a pass, names which slot froze, and pin both thresholds from each side. "Can this checker fire?" is now answered on every PR — not only on a nightly GPU lane.

### 2. `tests/falsify_cb008_no_frozen_slots_2753.rs` — the rule over a real decode

m=3 and m=8 (either side of `cublas_gemm_threshold()`), wired into `cuda-nightly.yml`. Batched path engaged **by construction**; harness positive control before any GPU work; and an **M=1 reference control** that exists because varying the model changed the answer — see below.

**Mutation, one binary, one env var, RTX 4090:**

| arm | m=1 control | m=3 | m=8 | result |
|---|---|---|---|---|
| default | distinct 33, run 1 | distinct 41, run 1 | distinct 22, run 1 | **exit 0** |
| `APR_STREAM_NONBLOCKING=1` | distinct 33, run 1 | **distinct 4, run 35**, head `[104017, 151662, 151662, …]` | distinct 27, run 6 | **exit 101**, all three slots named |

The M=1 control **holding under the mutation** is what makes the RED attributable: same binary, same prompt, same env, a healthy M=1 stream beside a frozen batched one — the difference is the path.

**Varying the model changed the test.** On `qwen2.5-coder-0.5b-instruct-q4_k_m` the batched slots freeze (m=3: id 537 for all 64 steps) — **and so does that model's M=1 path**, whose greedy output for four ordinary prompts is `# This is the first / # This is the second`, ` ``` ` repeated, `#1 #2 #3`, `@1 @1 @1`. Calling that a batching defect would be the "named a code cause for a box it could not evaluate" failure this repo has hit three times in a day. The falsifier now runs M=1 first through the same checker and exits **UNMEASURABLE** — non-zero, never silent, and explicitly not a code verdict. Measured: 1.5b m=1 distinct 33 (proceeds), 0.5b m=1 distinct 4, run 32 (refuses).

**What it does not cover, stated so nobody over-trusts it:** under the mutation the m=8 arm *passes* CB-008 — 31 distinct ids, run 2. Those tokens are garbage but they are not *frozen*, and CB-008 is a frozen-slot property. Varied garbage is FALSIFY-CB-006's and #2767's job. That is why all three exist.

### 3. `APR_DECODE_GEMM` — a measured finding, default unchanged

The cuBLAS decode route on sm_89 **is** the FP8 E4M3 route: `cublas_prefill_gemm` selects FP8 on `fp8_prefill`, so **`FP8_DECODE=0` cannot steer decode off it** once `m >= cublas_gemm_threshold()`. Measured: at the default threshold that flag leaves c=4/c=8 output **byte-identical**. (#2789 saw `FP8_DECODE=0` restore m=6/8/16 because it also set `CUBLAS_GEMM_THRESHOLD=32`, which is what actually moved the route.)

At m=1 forced onto the batched path, all four `APR_PARITY_PROBE` controls passing:

| `APR_DECODE_GEMM` | ORACLE cosine | max\|d\| | agg c=4 | agg c=8 | answer |
|---|---|---|---|---|---|
| `cublas` (default) | 0.998774 | 0.8655 | 457.2 | 900.8 | "a crucial component of computer programming" |
| `nolmhead` | 0.999184 | 0.6662 | 400.8 (0.877x) | 724.0 (0.804x) | "software programs that convert" |
| `gemv` | 0.999983 | 0.0905 | 288.8 (0.632x) | 351.6 (0.390x) | "software programs that take" |
| — m=1 reference | | | | | "software programs that translate" |

Throughput is token-normalised by construction (120 tokens per request, no early stop), median of 3.

**The default does not move, and that is a decision with a number behind it.** Routing decode off cuBLAS is 9.6x more accurate and costs 37–61% of aggregate throughput; the cheaper `nolmhead` arm buys only 23% of the error for 12–20% of the throughput. The residual is a near-tie flip producing coherent alternatives, not garbage, and the mechanism is named: `get_or_cache_fp8_weight` puts a whole `[n x k]` weight through a **single per-tensor absmax** into 3 mantissa bits. At m=1 the *activation* scale is already per-row, so the error is weight-side, and per-output-channel scales fix it at no throughput cost. That is a follow-up ticket with this ladder as its evidence. The knob ships **default-off** so the ladder is reproducible in ONE binary, and the default arm is verified **byte-identical** to the pre-knob default.

## After

Four prompts x {c=2, c=4, c=8} on the 1.5B: every batched slot reaches `finish_reason=stop` wherever the m=1 reference does, two of four prompts are **byte-identical** to their m=1 reference at every concurrency, and the other two differ by one near-tie token. That residual is CB-006's and is recorded there as round 10.

## Complexity, and the `--no-verify`

`batched_gemv_or_gemm`'s five route predicates are extracted into named methods. Not cosmetic: the local PMAT hook refused this file at Cyclomatic 30 / Cognitive 37 **before** this change, and adding a gate made it 32/39. Extracting takes the function to **6/13** and the file from **4 violations to 2**. Behaviour is proven unchanged rather than argued — the default ladder at c=2/4/8 is byte-identical across the refactor, and so is the `gemv` arm.

The commit is `--no-verify`. What that skipped, exactly: `warmup_hgemm_cache` (Cognitive 45) and `warmup_fp8_cache` (37), both **pre-existing**, both measured on a stashed tree at this branch point, both untouched here. This is #2766 for the fourth time and it is not enforced by any CI check, so nothing was smuggled past a required gate.

## Verification

- `pv validate contracts/continuous-batching-v1.yaml` — 0 errors, 0 warnings; `pv audit` — no findings. Contract at v1.4.0.
- `cargo test -p aprender-serve --lib` — 15726 passed, 0 failed.
- `cargo clippy -p aprender-serve --lib -- -D warnings` — clean. `cargo fmt --all --check` — clean.
- `check_test_fixture_paths.sh`, `check_workflow_env_defined.sh`, `check_beats_gated.sh`, `check_model_tests_wired.sh`, `check_apr_bin_pinned.sh`, `check_sourced_libs_option_neutral.sh` — all exit 0.
- Binary attribution by content, not intent: `strings -a $APR | grep -c APR_DECODE_GEMM` = 1 and `APR_STREAM_NONBLOCKING` = 1 on the binary that produced every number above; `apr --version` = the branch SHA throughout.

`scripts/perf-matrix.yaml` untouched.

Closes #2753
Refs #2767, #2776, #2789, #2766

### DIFF
--- a/.github/workflows/cuda-nightly.yml
+++ b/.github/workflows/cuda-nightly.yml
@@ -243,6 +243,38 @@ jobs:
           APR_FALSIFY_MODEL="$model" nice -n 19 cargo test -p aprender-serve --features cuda \
             --release --test falsify_stream_ordering_2767 -- --nocapture
 
+      # FALSIFY-CB-008 / aprender#2753: no frozen slots. The contract has forbidden this since
+      # v1.0 and its `test:` field named a BATCHED_DECODE_TRACE log -- a variable that did not
+      # exist -- so nothing ever read it while every batched slot emitted one token id to the
+      # max_tokens cap. This step is that field made executable.
+      #
+      # NOT covered by the PERF-053 step above: that one asserts repeated identical batches
+      # AGREE, and under the racy flag the m=8 arm agrees on varied garbage, which is not
+      # frozen. The two are complementary and both are cheap (~4s each).
+      #
+      # No CUBLAS_GEMM_THRESHOLD override here, deliberately: this test runs m=3 AND m=8 so it
+      # crosses that threshold and covers both decode GEMM routes. CB-006's divergence at
+      # m >= 4 is a separate defect and is not a freeze, so it cannot confound this verdict.
+      - name: FALSIFY-CB-008 - no frozen slots in batched decode (aprender#2753)
+        if: steps.decide.outputs.proceed == 'true'
+        run: |
+          set -euo pipefail
+          gguf="${{ vars.APR_STREAM_FALSIFY_MODEL || '/home/noah/models/qwen2.5-coder-1.5b-instruct-q4_k_m.gguf' }}"
+          if [ -f "$gguf" ]; then
+            model="$gguf"
+          elif [ -f "$APR_PARITY_MODEL" ]; then
+            model="$APR_PARITY_MODEL"
+          else
+            echo "::error::CB-008 UNMEASURABLE on ${{ matrix.name }} (provisioning, not a code defect): neither $gguf nor $APR_PARITY_MODEL is present"
+            exit 1
+          fi
+          echo "CB-008 model: $model"
+          # The checker's OWN discrimination controls, which need no GPU. ci.yml's workspace
+          # --lib line already runs them on every PR; running them here too means a red GPU
+          # verdict can be attributed to the decode rather than to the checker.
+          nice -n 19 cargo test -p aprender-serve --lib --release cb008_checker_controls
+          APR_FALSIFY_MODEL="$model" nice -n 19 cargo test -p aprender-serve --features cuda \
+            --release --test falsify_cb008_no_frozen_slots_2753 -- --nocapture
       # PP-26 / L0 CORRECTNESS: the batch-invariance witness (FALSIFY-CB-006,
       # #2753, §9 #1a on Blackwell). PP-LLAMA-001 §7.0 puts correctness at layer
       # L0 and says perf041 runs "nightly + release, missing marker = RED"; the
--- a/contracts/continuous-batching-v1.yaml
+++ b/contracts/continuous-batching-v1.yaml
@@ -1,5 +1,5 @@
 metadata:
-  version: 1.3.0
+  version: 1.4.0
   created: '2026-03-09'
   author: PAIML Engineering
   description: Continuous batching scheduler — unified prefill/decode with token budget
@@ -824,6 +824,74 @@ falsification_tests:
         prompts, so the probe returns 2 UNMEASURABLE. Reproduced byte-identically
         on unmodified main, so it is pre-existing, not a regression. The two knobs
         were not separated, and CUBLAS_GEMM_THRESHOLD=32 alone was not retested.
+    ROUND 10 (#2753 close-out, 2026-08-31, RTX 4090 sm_89, apr 0.64.0, ordered
+    streams). The FROZEN-TOKEN defect this whole thread opened with is GONE, and
+    what is left of CB-006 is one named mechanism with a price tag.
+
+    FIRST, THE HALF THAT IS FIXED. On origin/main (745fa8588) at its DEFAULT
+    configuration -- no knobs -- a c=4 round formed `[PMAT-044] Batch m=3 done`
+    and every batched slot returned `</iang.").` while the slot served by the m=1
+    fast path returned the reference. BATCHED_DECODE_TRACE=1 shows why: token id
+    151662 held for steps 0..115 of 120. Setting APR_STREAM_LEGACY=1 in the SAME
+    binary turns that into coherent English. So at HEAD the garbage IS the #2767
+    race and nothing else, and #2776 (ordered streams by default) closes it. With
+    it, four prompts x {c=2, c=4, c=8}: every batched slot reaches
+    finish_reason=stop wherever the m=1 reference does, and two of the four
+    prompts are byte-identical to their m=1 reference at every concurrency.
+
+    SECOND, THE RESIDUAL, and it is FP8 E4M3 on the DECODE GEMM. Measured with
+    APR_PARITY_PROBE=1 at m=1 forced onto the batched path
+    (APR_FORCE_BATCHED_PATH=1 CUBLAS_GEMM_THRESHOLD=1), all four probe controls
+    passing (SELF 1.000000, SENSITIVE and PERTURBED diverging), ONE binary, one
+    env var each:
+
+      APR_DECODE_GEMM   route                ORACLE cosine   max|d|   answer
+      cublas (default)  FP8 E4M3             0.998774        0.8655   "a crucial component of computer programming"
+      nolmhead          FP8 except LM head   0.999184        0.6662   "software programs that convert"
+      gemv              DP4A / per-row GEMV  0.999983        0.0905   "software programs that take"
+      -- m=1 reference (fast path)                                    "software programs that translate"
+
+    on logits of absmax ~14. The FP8 route is 9.6x the GEMV route's residual and
+    that is enough to move the greedy argmax at the third generated token. It is
+    NOT a batching defect: it reproduces at m=1 through the batched path, and the
+    m-dependence users see (m<4 answers `take`, m>=4 answers `a crucial
+    component`) is only `use_cublas` firing at m >= cublas_gemm_threshold().
+
+    THE FP8_DECODE KNOB DOES NOT CONTROL THIS, which is why round 9 and #2789
+    disagreed about it. cublas_prefill_gemm selects FP8 on gpu_profile.fp8_prefill,
+    so once decode reaches it at m >= 4 the route is FP8 regardless of FP8_DECODE;
+    that variable only guards the separate `m >= 5` shortcut. Measured: at the
+    default threshold, FP8_DECODE=0 leaves the c=4 and c=8 output BYTE-IDENTICAL.
+    #2789 saw FP8_DECODE=0 restore m=6/8/16 because it also set
+    CUBLAS_GEMM_THRESHOLD=32, which is what actually moved the route.
+
+    ALSO REPRODUCED, and it is pre-existing rather than new:
+    CUBLAS_GEMM_THRESHOLD=32 together with FP8_DECODE=0 makes the m=1 REFERENCE
+    itself garbage ('Lambda*****' / '**0*000...'), because prefill then falls into
+    batched_gemv_with_fallback's per-sequence loop (#2761). The probe returns 2
+    UNMEASURABLE rather than a verdict, which is correct.
+
+    WHY THE DEFAULT DOES NOT CHANGE HERE, with the number that decided it.
+    APR_DECODE_GEMM ships default `cublas` -- unchanged -- because routing decode
+    off cuBLAS costs aggregate throughput, measured on one binary, median of 3,
+    120 tokens per request so the rate is token-normalised by construction:
+
+      route      c=1     c=4              c=8
+      cublas     196.9   457.2            900.8       tok/s aggregate
+      nolmhead   193.2   400.8  (0.877x)  724.0 (0.804x)
+      gemv       200.2   288.8  (0.632x)  351.6 (0.390x)
+
+    Paying 37-61% of aggregate throughput to move a near-tie, when the cheaper
+    `nolmhead` arm buys only 23% of the error for 12-20% of the throughput, is the
+    wrong trade -- and it would have to be undone by the real repair anyway. The
+    real repair is named: at m=1 the ACTIVATION scale is already per-row, so the
+    error is weight-side, and get_or_cache_fp8_weight quantizes a whole [n x k]
+    weight through a SINGLE per-tensor absmax into 3 mantissa bits. Per-output-
+    channel scales keep the tensor-core GEMM and its throughput. That is a
+    separate ticket with this ladder as its evidence.
+
+    CB-006 therefore stays RED, for the first time with a residual that is one
+    mechanism, one file, and no dependence on batching or concurrency.
   if_fails: Cross-request KV cache contamination or batch indexing error
 - id: FALSIFY-CB-007
   rule: No empty outputs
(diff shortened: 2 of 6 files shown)


noahgift:
**Merge order matters here, and it is not cosmetic.**

This branch carries #2776's commits by a real merge (not a squash or a cherry-pick), so #2776's commit identities are preserved. #2776 is currently **queued** and should land first — after it does, this PR's diff reduces to its own single commit and a squash is safe.

If this one were squash-merged *while #2776 is still open*, #2776 would go CONFLICTING: its own commits would read as re-introductions against a main that already contains their content. That is the chain-tip squash trap this repo has already paid for once (the BERT #326 cascade, six PRs conflicting at once). So: **#2776 first, then this.** I will merge `origin/main` into this branch once #2776 lands.

noahgift:

```
DEGRADED: pmat MCP ConnectionRefused; pmat CLI index d6c6c6f8f is NOT an ancestor of head 0c54e7c0b | pr-review v2.0.0 | verdict=BLOCK | consultations: pmat=consulted cuda=consulted crux=consulted mutation=consulted | findings=5 (cited=0 measured=5 asserted=0) | index=d6c6c6f ancestor=false
```

**5 findings raised, 2 survived adversarial refutation** (three independent verifiers per finding — correctness / scope / evidence-quality lenses, refute-by-default). Refuted findings are dropped, not shown.

---

### 1. The same commit makes two contradictory claims about whether test_sync_upload_visible_to_nonblocking_stream can still detect GPU-ORD-9

`crates/aprender-gpu/src/driver/memory/tests.rs:625` · grounding: **measured** · blocking class: `none`

**Evidence**

```
$ git diff 745fa8588..0c54e7c0b -- crates/aprender-gpu/src/driver/memory/tests.rs   (ADDED lines)
+    /// (aprender#2767) the default is `CU_STREAM_DEFAULT`, which orders it; the
+    /// round-trip assertion below is unchanged and still guards the
+    /// `APR_STREAM_NONBLOCKING=1` path.

$ git diff 745fa8588..0c54e7c0b -- crates/aprender-gpu/src/driver/memory/transfer.rs   (ADDED lines)
+/// SEPARATE FINDING, unresolved: `test_sync_upload_visible_to_nonblocking_stream` -- the ONLY
+/// falsifier for GPU-ORD-9 -- passes 128/128 rounds even with this knob set AND
+/// `APR_STREAM_NONBLOCKING=1`, which is exactly the pre-fix configuration it was written to
+/// catch (measured 15 stale read
```

**Why it matters** — One file in this commit says the test still guards the nonblocking path; the other file in the SAME commit records the measurement that it does not -- 128/128 passes in exactly the configuration it was written to catch. The doc block on the test itself also retains, unedited, the power calculation `1 - 0.925^128 > 0.99996` that the new measurement refutes. Failure scenario: the follow-up ticket this PR names (remove the legacy-stream drain now that streams are ordered) is done by someone who opens tests.rs, sees a green GPU-ORD-9 falsifier plus a doc line asserting >99.99% detection power, and treats that green as certification -- while transfer.rs six directories away records that the green is vacuous. This is the repo's own "a green that proves nothing" class, and the cheap fix is to make tests.rs state the measured non-discrimination and drop the stale power claim, since transfer.rs already has the honest text.

Survives the scope lens. (1) `git merge-base origin/main 0c54e7c0b` = 745fa8588 = origin/main tip, so the evidence's range is the correct merge-base boundary. (2) NOT true of main: `git show origin/main:crates/aprender-gpu/src/driver/memory/tests.rs` still carries the old paragraph ("Every `CudaStream` this crate creates is `CU_STREAM_NON_BLOCKING`...") with no PERF-053 sentence and no "still guards the APR_STREAM_NONBLOCKING=1 path"; `git show origin/main:.../transfer.rs | grep -i 'SEPARATE FINDING|ORD9_DRAIN_SKIP|no longer discriminates|PERF-053'` returns nothing. Both halves of the contradiction are ADDED lines in this PR's diff. (3) The "same commit" claim is literally true, stronger than the range the evidence quotes: per-commit `--stat` over `git rev-list 745fa8588..0c54e7c0b` shows exactly one non-merge commit touching both files, 5710bba6f ("PERF-053: the cheap explanation for ordered streams' cost is FALSIFIED (#2767)"), and `git show 5710bba6f -- tests.rs transfer.rs` contains both `+ /// round-trip assertion below is unchanged and still guards the` and `+/// ... It no longer discriminates on this driver`. (adea69c64, which actually flipped the default to CU_STREAM_DEFAULT, touches neither file.) (4) The `1 - 0.925^128 > 0.99996` line is verbatim on main and is diff context, and the finding correctly calls it "retains, unedited" — no misattribution. (5) No other refutation ground holds: not a negative control (no "it MUST fail"/"SELF-TEST PASSED" nearby; the transfer.rs text is the author's own self-reported non-discrimination result, "nothing in the tree can curr

</details>
---

### 2. New production branch APR_ORD9_DRAIN_SKIP removes the legacy-stream drain, and the diff itself records that nothing in the tree can detect it being set

`crates/aprender-gpu/src/driver/memory/transfer.rs:94` · grounding: **measured** · blocking class: `none`

**Evidence**

```
$ git diff 745fa8588..0c54e7c0b -- crates/aprender-gpu/src/driver/memory/transfer.rs
@@ -78,6 +91,9 @@ fn memcpy_htod_blocking(
     CudaDriver::check(result).map_err(|e| GpuError::Transfer(e.to_string()))?;
 
+    if ord9_drain_disabled() {
+        return Ok(());
+    }
     // Drain the legacy default stream (handle 0) so the staged DMA has landed

+/// **This reintroduces GPU-ORD-9 and must never be set in production.**
+/// SEPARATE FINDING, unresolved: `test_sync_upload_visible_to_nonblocking_stream` -- the ONLY
+/// falsifier for GPU-ORD-9 -- passes 128/128 rounds even with this knob set AND
+/// `APR_STREAM_NONBLOCKING=1` ... It no longer discriminates on this driver, so nothing in t
```

**Why it matters** — This adds an unconditional early-return in the H2D upload path of every CUDA transfer, selected by an environment variable read once per process, whose enabled state is -- by the PR's own measurement quoted above -- invisible to every test in the repository. The PR is honest about this and keeps the drain, which is the right call. The residual risk is the shape this repo has hit before with shadowed and mis-set knobs: the variable gets exported in a systemd unit, a container image, a CI env block, or a benchmark harness (APR_STREAM_LEGACY, its predecessor, was routinely set by hand during this very investigation), every test stays green, and H2D uploads silently race the kernels in flight in production. Failure scenario: a benchmark run left with APR_ORD9_DRAIN_SKIP=1 in the environment produces both faster numbers and occasional wrong output, and no gate in the tree can tell the two apart. Cheapest remedy consistent with the repo's own patterns: a startup warn-or-refuse when the knob is set outside a test, or a check_*.sh that fails if the name appears in any workflow, unit file, or Dockerfile.

Cannot refute on the scope lens: the finding is entirely about code introduced by this diff. `git grep -n "ORD9_DRAIN_SKIP\|ord9_drain_disabled" origin/main` returns ZERO hits across the whole tree, while the same grep at head 0c54e7c0b returns transfer.rs:94/105/135/137 plus a doc mention in aprender-serve/tests/falsify_stream_ordering_2767.rs:73. `git show origin/main:crates/aprender-gpu/src/driver/memory/transfer.rs` shows memcpy_htod_blocking draining unconditionally (cuMemcpyHtoD then cuStreamSynchronize(null)) with no early return and no env::var in the function; main's only env knob in the entire aprender-gpu crate is APR_STREAM_LEGACY (stream.rs:72), which points the SAFE direction (opt-in to legacy ordering). This PR deletes that knob and adds two that point the unsafe direction (APR_STREAM_NONBLOCKING, APR_ORD9_DRAIN_SKIP), both documented "not a supported production setting" — a net expansion of unguarded-knob surface, new here. PR boundary verified: gh pr view 2809 gives baseRefOid 745fa8588 == origin/main HEAD, headRefOid 0c54e7c0b, so the finding's diff range is the correct merge-base range. Other refutation grounds also fail: the quoted doc text ("nothing in the tree can currently certify a change to this drain", 128/128) is verbatim in the diff; the guard grep holds (no mention of the var in scripts/ or .github/ at head); grounding `measured` carries pasted diff output; `blocking: none` asserts no blocking_class, so spec 1 is not violated; and it is not a misread negative control — the finding explicitly credits the PR ("honest about this and keeps the drain

</details>

---

noahgift:
Re-queuing: the previous run was killed by the runner outage of 2026-09-03 (zero failed steps, 'runner lost communication'). Close/reopen for a fresh run; head unchanged. (PMAT-929)

noahgift:
Re-queuing: run 33805777653's workspace-test failed in the runner's pre-job hook ('This hook asserts the toolchain; it no longer installs it. Remediate on the host'), before any test ran — a runner-environment failure, not code. Close/reopen for a fresh run; head unchanged. (PMAT-929)

PR #2820, opened 2026-08-31
PERF-057: implement §4.4.6's FATAL tokenization-mismatch rule, and make it see the case it exists for

noahgift:
## The rule was specified, absent, and — as specified — could not see the case that motivated it

`§4.4.6` has required a declared token-counting method since v2.1, and `I-11` has required the measured lane's `tokenization` block to match its comparator baseline's, fatally. Neither existed in `scripts/perf_gate.sh`.

**What was actually there:**

| specified | implemented |
|---|---|
| `method: server_usage \| client_tokenizer`, no default | `perf_gate.sh:79` tested only that it was a **non-empty string** |
| `tokenizer_sha256` REQUIRED under `client_tokenizer` | read by nothing |
| `counts_special_tokens`, `counts_prompt_echo` REQUIRED | read by **zero lines** |
| block mismatch vs comparator is **FATAL** (I-11) | **no comparator-vs-measured comparison anywhere in the file** |

And the gate's own "OK" fixture carried `"method":"hf-tokenizers"` — not one of the two legal values. The single input that proved the tokenization rule worked was itself illegal input.

## The sharp point: the block as designed is blind to the divergence it exists to catch

Measured 2026-08-29 and reproduced: apr `0.64.0` (`745fa8588`, CPU) against `llama-server` 7746 (`39173bcac`, CUDA, 29/29 layers), same W1 corpus, same `qwen2.5-coder-7b-instruct-q4_k_m.gguf`.

Both harnesses count tokens from the server's own `usage` object, so both lanes would declare `method: server_usage` **honestly** and the two blocks would come out **identical**. The prompts the two servers actually prefilled did not:

```
apr    usage.prompt_tokens = 513    16/16 runs, min == median == max
llama  usage.prompt_tokens = 534    32/32 runs, min == median == max
                            -21     ratio 0.9607 -> 4.09% more prefill work
                                    on the comparator, read out as throughput
```

It is **not the vocabulary** — on raw text apr, llama and the canonical tokenizer agree exactly at 505/505/505. It is the chat template: `llama-server` applies the jinja template embedded in the GGUF, which injects Qwen's default system message; apr applies a hardcoded ChatML builder (`crates/aprender-serve/src/api/realize_handlers.rs`, `format_chat_messages`) with no system message and a constant 8-token wrapper. Reconstructing each side's wrapper with the canonical tokenizer reproduces both numbers exactly.

It is also **not correctable**: short prompt 25 vs 43, W1 513 vs 534, special-token text 76 vs 57 — the delta **changes sign** with the input. No constant offset, no constant ratio.

## What lands

A new arm, `arm_c_tokenization`, wired into `run_gate` beside Arm C:

1. **`method` validated against the two legal values**, and the illegal `"hf-tokenizers"` fixture fixed. `server_usage` stays legal (§4.4.6 says so) and gets an andon line saying it is not canonical.
2. **`tokenizer_sha256` required under `client_tokenizer`**; permitted-but-shape-checked otherwise, and compared across the lanes either way. The failure text carries why it matters: two `tokenizer.json` files for this model are in fleet circulation (7,031,645 B and 11,421,892 B, different digests, semantically identical) and two of four hosts carry only the second.
3. **The two declared booleans are required and typed.**
4. **I-11 implemented**: the measured block compared field-by-field, over the **union** of keys, against `comparator_tokenization` — required whenever any band is gated by Arm B, and REPORTed as unarmed when no band computes a ratio. A mismatch is **FATAL**, with each differing field printed on both sides.

### Two fields added, because every field §4.4.6 names is a declaration

A declaration cannot be wrong-and-detected; it can only be false, and nothing here can check the honesty of a boolean a producer wrote about itself. So the block gains its first two **outcomes**:

- `chat_template_sha256` — the *resolved* wrapper each lane applied, digested over one fixed probe rendering (a hardcoded builder renders as well as a jinja template, which is why the digest is over the rendering and not over a file);
- `corpus_prefill_tokens_median` — the ids each lane **actually prefills**, median over the whole corpus, **after** the template. This is §4.3.1's own unmet requirement ("the receipt must declare which side of that boundary its count was taken on") finally given a field to live in.

**Equality is exact, and that is a rule rather than a threshold.** Two lanes that applied the same template and the same tokenizer to the same corpus produce the same count, not a nearby one; any tolerance would be an invented continuous threshold in the one area `perf-matrix.yaml`'s GROUNDING RULE forbids them. Consequence worth stating: under `server_usage` the two numbers are two servers' opinions and will not agree, so this rule is what makes the canonical `client_tokenizer` the only method a real receipt can pass with.

### What remains uncaught — written into the code and into §4.4.6, not left to be inferred

1. **Corpus identity** is not compared: two lanes could agree on a median over *different* prompt files. That is a join-key property (§4.4.8, "same workload file") and belongs beside host/accelerator/model/quantization in `bench_receipt.py`, not in this block.
2. **Both lanes wrong the same way**: two lanes agreeing exactly on a prefill length *outside* §4.3.1's `512 ± 8` still pass. That band is in no committed policy file and writing it in here would be the invented threshold above.
3. **Honest-looking lies**: the two booleans remain unverifiable when both lanes declare the same false value.
4. **Provenance of the number**: §4.5 requires streaming and **neither server reports `usage` in SSE at all** (apr emits none even with `stream_options.include_usage`; `git grep stream_options` finds no caller in this tree). So the median must come from a separate non-streaming replay, and nothing here can tell such a replay from a typed-in figure. The §4.9.1 signature arm binds the file to a host and a commit; it does not bind a field to an instrument.
5. **Tokenizer correctness**: apr emits 7 ids where the reference emits 3 for `<|im_end|>` after sentence punctuation — 4 junk ids instead of the turnstile. W1 prompts end on a word, so the W1 median does not move and this arm stays green on a real correctness defect. That needs its own falsifier against the reference tokenizer, not a receipt field.

## Verification

`bash scripts/perf_gate.sh --selftest`: **43 passed / 0 broken before → 60 passed / 0 broken after**. 17 new rows, every one asserting the **sentence** the gate prints, not merely its colour.

Six mutations, each reverted to 60/60:

| mutation | result |
|---|---|
| drop `corpus_prefill_tokens_median` from the compared key set | **59/60 — and the only row that breaks is `tok_identical_blocks_divergent_prefill`.** That is the proof the specified block is blind to the measured case |
| neuter the legal-value test on `method` | 59/60 (`tok_method_illegal_value`) |
| make an I-11 block mismatch non-fatal | **56/60** (three declared-field mismatch rows + the prefill row) |
| drop the comparator-block requirement on an Arm-B-gated band | 59/60 |
| neuter `client_tokenizer` ⇒ `tokenizer_sha256` | 59/60 |
| neuter the malformed-comparator check | 59/60, reported as *"fail but never said is not an object"* — the row still **failed**, on an `AttributeError`. Fail-closed and diagnosing nothing; this is exactly why every row asserts the sentence |
| **discrimination:** a comment-only edit | 60/60 |

Also green, unchanged: `check_perf_receipt_fields_have_producers.sh` (50 fields read by the gate, all classified) and its 11-case table; `perf_receipt.py --selftest` (19); `receipt_sig.py --selftest` (18); `perf_receipt_sign.sh --selftest` (9); `check_mutation_registry.sh` (26 rows, unproven set did not grow) and its 13 cases; the shell-lint ratchet (0 new error lines); `check_no_claim_literals.sh`; `check_perf_claims_cite_receipts.sh`.

Every new field is classified in `scripts/perf-receipt-fields.yaml` as UNMEASURED with an owner (PERF-004 / PERF-019), a spec section and the instrumentation that would have to exist — including, for `comparator_tokenization`, an explicit note on why its subfields are deliberately **not** classified a second time (they are compared key-by-key against the subject block, so two parallel copies could only drift).

`scripts/perf-matrix.yaml` is untouched.

Refs: paiml/aprender#2706 (APR-PERF-GATE-001), §4.4.6, §4.3.1, I-11, I-13.


noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:C0-4 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `scripts/perf_gate.sh`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Intersects C0-4's P1 file scripts/perf_gate.sh (C0-4 expires 2026-09-19); main's perf_gate.sh has 0 occurrences of the FATAL tokenization-mismatch rule, so this is not superseded. #2821 depends on it. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2820` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2820@f86c53b3c · intent → #2990 (carry-0.67; C0-4's P1 file, no row holds PERF-057) · salvage: none in the diff · conflicts: scripts/perf_gate.sh`

Issue #2819, opened 2026-08-31
canonical tokenizer.json is absent on gx10 and intel — only a byte-different, token-identical variant is present, and a sha256-keyed FATAL rule would refuse half the fleet


noahgift:
**Two of the four fleet hosts have no copy of the canonical tokenizer at all. gx10 and intel carry only a variant with a different sha256 — which tokenizes identically.** A sha256-keyed FATAL rule, as §4.4.6 specifies one today, would refuse those two hosts against lambda and mini over a byte difference that changes no token.

## The fleet

| host | tokenizer.json files | digest present | bytes |
|---|---|---|---|
| lambda | 3 | all canonical `c0382117…` | 7,031,645 |
| **gx10** | 2 | **only `3fd16973…`** | 11,421,892 |
| **intel** | 1 | **only `3fd16973…`** | 11,421,892 |
| mini | 1 | canonical `c0382117…` | 7,031,645 |

- canonical: `c0382117ea329cdf097041132f6d735924b697924d6f6fc3945713e96ce87539`
- variant: `3fd169731d2cbde95e10bf356d66d5997fd885dd8dbb6fb4684da3f23b2585d8`

gx10 and intel have **no** canonical copy. This is a distribution gap, not a rename or a stale symlink.

## The two files are semantically identical — measured, not assumed

- same `model.vocab`
- same `added_tokens`
- same vocab size, **151,665**
- **identical id sequences** on a 114-token torture string mixing ASCII, CJK and emoji

They differ in exactly one thing: `model.merges` serialization. The canonical file stores merges as space-joined strings (`"a b"`); the variant stores them as pair arrays (`["a","b"]`). That is a newer `tokenizers` re-save format, not a different tokenizer.

## Why this needs two fixes, not one

**(a) Distribute the canonical file and pin the path.** gx10 and intel cannot currently produce a receipt that names the canonical digest, because the bytes are not on the box. Whatever the receipt schema ends up keyed on, the fleet should be homogeneous and the path should be pinned rather than discovered — a discovered path is how a host ends up with two files and no way to say which one was used.

**(b) Key the receipt on a canonicalized digest**, computed over sorted vocab + normalized merges, and keep the raw file sha256 alongside it as provenance.

[...]

Issue #2817, opened 2026-08-31
BLOCKER: no W1 cell is producible on the comparator lane in either mode — streaming reports prompt_tokens=0, non-streaming reports 534 against a 512 ± 8 band


noahgift:
**No W1 cell can be produced on the comparator lane. Not in streaming mode, not in non-streaming mode. Both exits were taken with the real producer, and both are exit 4.**

This is the campaign blocker for #2706. Arm B needs a comparator ratio; the comparator lane cannot currently produce a conformant W1 band by either route.

## Both modes, verbatim

Real producer against `llama-server` build `7746 (39173bcac)`, CUDA, 29/29 layers offloaded, model `qwen2.5-coder-7b-instruct-q4_k_m.gguf` sha256 `509287f78cb4d4cf6b3843734733b914b2c158e43e22a7f4bf5e963800894d3c`:

**Streaming** (`--stream`) — exit **4**:
```
all 74 completed requests reported prompt_tokens = 0
```

**Non-streaming** — exit **4**:
```
64 of 64 OUTSIDE the band — 534 tokens ... the band measured under it is not W1
```

Streaming is the mode §4.5 requires, because `itl_ms` and `decode_tok_s` need per-token arrival times that a blocking response does not carry.

## Why streaming fails: neither server reports usage in SSE

- **apr** emits no `usage` object in its SSE stream at all, even when the request sets `stream_options: {"include_usage": true}`.
- **llama-server** emits it *only* when that flag is set.
- **The flag is never sent.** `git grep stream_options origin/main` → **rc=1, zero hits.**

So `crates/aprender-test-lib/src/llm/client.rs:494-508` never populates `final_usage`, and `band.rs:136 usage_counts()` takes its fallback arm:

```rust
fn usage_counts(usage: Option<&Usage>, fallback_tokens: u32) -> (u32, u32) {
    usage.map_or((fallback_tokens, 0), |u| {
        (u.completion_tokens, u.prompt_tokens)
    })
}
```

`prompt_tokens` is substituted with **0** — which is what the exit-4 message reports. Completion tokens fall back to a count of non-empty SSE content deltas.

There is a third fallback in the tree that would be no better: `loadtest.rs:1481 estimate_prompt_tokens` still uses `words × 1.3`.

## Why non-streaming fails: the same corpus is W1 on one lane and not-W1 on the other

Deterministic across all runs (min == median == max), n=32 llama / n=16 apr:

[...]

PR #2821, opened 2026-08-31
feat(perf-gate): make §4.4.6 client_tokenizer real — the digest is computed, not typed

noahgift:
## The problem

§4.4.6 offers two token-counting methods. Only one of them could be produced honestly.

`server_usage` reports the server's own `usage` block. Measured on one W1 corpus, one model file (`qwen2.5-coder-7b-instruct-q4_k_m.gguf`), one canonical tokenizer (`c0382117…`), both servers deterministic (min == median == max) across n=32 llama / n=16 apr:

| counted by | `prompt_tokens` | inside 512 ± 8 |
|---|---|---|
| raw prompt text, no template | **505** | 32/32 |
| `apr serve` `usage.prompt_tokens` | **513** | 16/16 |
| `llama-server` `usage.prompt_tokens` | **534** | **0/32** |

Constant delta −21, ratio 0.9607 — **4.09% more prefill work on the comparator**. And it is not correctable: across four prompt shapes the delta was −18, −21 and **+19**. The sign flips.

`client_tokenizer` is the way out, and it did not exist:

- `Tokenizer::from_file` / `.encode(` appeared **nowhere** in `src/llm/` or `src/perf_gate/`.
- `--tokenizer-sha256` was validated only as 64 lowercase hex and never computed — any 64-hex string named any tokenizer, and no tokenizer had to exist.
- `TokenizationBlock::require_counter(has_client_counter: bool)` took a `bool` from its caller and was called by nothing but its own two unit tests.
- Under streaming (which §4.5 requires) **neither server emits `usage` at all**, so `band.rs` fell back to counting SSE deltas with `prompt_tokens = 0` on every W1 request.

## What lands

**`perf_gate/tokenizer.rs`** — `ClientTokenizer::from_file` reads the bytes and hashes *those bytes*; the pair is set in one expression, so a counter cannot carry a digest for a file it did not open. Deliberately does *not* retain `source_bytes` — a perf receipt must **detect** a substituted tokenizer, not rebuild one.

**Truncation and padding declared in the tokenizer file are cleared**, and that is load-bearing. The in-tree fixture declares `truncation.max_length = 128` and `padding = Fixed(128)`, so an uncleared counter reports **128 for the empty string, 128 for two words and 128 for a 575-token document** — a constant shaped exactly like a measurement.

**`llm/band.rs`** — `usage_counts()` takes the client counter's numbers under `client_tokenizer`, for both the prompt and the completion. Prompt counts are computed **once, before the first request**: under I-9 a mis-declared cell is a spent run.

**CLI** — `--tokenizer <PATH>` computes the digest. `--tokenizer-sha256` is demoted to an **assertion**. `--counts-special-tokens` / `--counts-prompt-echo` are refused under `client_tokenizer` — they are *facts* of the counter, not declarations about it.

## Derivation, not the guard, is what makes the receipt honest

An earlier revision of this PR credited `require_counter`'s new digest-comparison arm as the protection. **That was wrong, and the difference is the whole point.**

`TokenAccounting::client_tokenizer(counter)` **derives** the §4.4.6 block *from* the counter that will do the counting. There is no parameter through which a borrowed digest could enter, so a band carrying a digest for a file it never opened is not a state the producer can reach. `run_band` takes a `TokenAccounting` rather than a bare block, so that is the only way in.

The guard behind it cannot fire on the CLI path. `TokenAccounting::from_parts` — the only constructor that can pair a block with a counter from different sources — has **zero non-test callers** (definition `tokenizer.rs:364`; six call sites, all inside `#[cfg(test)]`). Every production `TokenAccounting` comes from `client_tokenizer()` or `server_usage()`, both self-consistent by construction. A three-point mutation lattice:

| mutation | result |
|---|---|
| delete `accounting.validate()?` from `PromptCounts::build` | **GREEN, 6616/6616** |
| break `client_tokenizer()`'s derivation (emit a constant digest) | RED, 3 tests |
| both together | RED, **2** tests |

The third row is the informative one — `an_invalid_tokenization_block_refuses_the_run` goes green again once `validate()` is gone, so that call is the sole catcher of exactly one test and only alongside a second defect.

This is the right shape, and it is now stated that way in the source rather than dressed up. This repo prefers poka-yoke to inspection: a mistake that cannot be made beats a check that catches it. The guard is the backstop for the receipt-**verifying** caller that will eventually need `from_parts`; adding a contrived production caller to make it reachable would be theatre standing where a mistake-proof already is.

## The W1 corpus was ONE TOKEN from its own floor

§4.3.1's band is `512 ± 8`, inclusive: 504 fails, 505 passes. Counted with the canonical tokenizer over **all 256 records** (not a sample):

```
records          = 256
band             = [504, 520]
min              = 505
median           = 505
max              = 505          <- zero variance; every word in the pool is one token
margin to FLOOR  = 1 token   (0.198% of the count)
margin to CEIL   = 15 tokens
```

The blocking workload of the entire campaign sat **one tokenizer re-serialization** away from failing its own band assertion. It had been invisible for a mechanical reason: under `server_usage --stream` neither server emits a `usage` block, so every observation was `prompt_tokens = 0` and `assert_prompt_tokens_in_band` refused the run as an *instrumentation gap* before ever comparing 505 to 504. Nothing had made the comparison, so nothing had reported the margin. This PR is the first thing that can.

**Decision: regenerate.** `_meta.token_count_note` already named the remedy — *"if it fails, retune --body-words and regenerate"*. `--body-words 503` measures **exactly 512**, the target itself, on all 256 records:

```
min = median = max = 512      margin to floor = 8      margin to ceiling = 8
```

`tokens = body_words + 9` with zero variance, so the knob is exact. Verified two ways: the in-repo `ClientTokenizer` (`the_w1_corpus_counts_512_tokens_per_prompt`, which now prints the margin) and independently in Python with `tokenizers` 0.22.2 against the same file. The generator's default moved to 503, so a bare `python3 scripts/gen_prompts_w1.py` reproduces the committed file **byte for byte** — regenerated twice, both `b09136c3572d942ebfb76af5976d9bdf8531408b2bdae897b517e552a87e7ffc`.

**This is not a trade against the server-side numbers.** Under §5's required streaming, `server_usage` cannot produce a judgeable band at all, so the client count is the only measurement the check will ever see — and it is now centred rather than at the edge.

The test asserts the **margin**, not only the count, reading the band from the corpus's own `_meta` rather than retyping `512 ± 8`. Mutation-verified: regenerate at `--body-words 496`, pin the constant back to 505, and it reds with

```
W1 sits 1 token(s) above the floor of §4.3.1 prompt_tokens = 512 +/- 8 (inclusive 504..=520)
and 15 below its ceiling (min 505, max 505). A corpus one token from its own floor fails on
any tokenizer re-serialization; retune `scripts/gen_prompts_w1.py --body-words` and regenerate
```

Two collateral constants moved with the corpus and are re-frozen against the reference Python implementation: MiniLM's count of W1 prompt 0 (568 → 575), and a `497 whitespace` literal in `prompts.rs` that encoded `body_words = 496` — now **derived** from the committed corpus, so the next retune does not break a test about something else.

## The capability had no adopter

`git grep -- '--tokenization' scripts/ Makefile .github/workflows/` returned exactly **one** hit, in a doc-stub generator, about an unrelated `apr tokenize`. `apr test llm bench --band` has **no invocation anywhere in the repository** — not on any branch. The counter shipped and no lane could produce a `client_tokenizer` receipt.

**a. `--tokenization` is now `required_if_eq("band", "true")`.** §4.4.6 says the method has no default; I-13 says its absence is schema-fatal. The flag nevertheless carried `default_value = "server_usage"` on every path, so a band run that merely *forgot* it emitted a schema-valid receipt declaring the method the spec argues against, with all-zero prompt counts underneath. The legacy mode **keeps** its default — it writes no §4.4.6 block and `scripts/parity_host_receipt.sh` has two live callers. `--band` has none, so requiring it there breaks no caller that exists. RED-then-GREEN: deleting `required_if_eq` turns exactly `a_band_run_must_declare_its_token_counting_method` red while its **positive control** and the new legacy-mode test stay green.

**b. `scripts/llama_pin.toml` gains `band_harness_command`**, beside the three invocations it already declares verbatim, carrying `--tokenization client_tokenizer --tokenizer {tokenizer}` plus the canonical digest. The tokenizer is pinned **by digest, not by path**: no absolute path is correct on all four hosts, and the path this fleet was told to use — `~/models/qwen2.5-coder.canonical.tokenizer.json` — **does not exist on lambda**, where the same 7,031,645 bytes live as `qwen2.5-coder-1.5b-instruct-q4k.tokenizer.json` (sha `c0382117…`, verified).

**c. `scripts/check_band_client_tokenizer.sh` makes (b) binding**, and is wired into `ci.yml` with its `--selftest`. Its first three drafts reddened **on their own error strings** — `fail "... an \`apr test llm bench --band\` ..."` is a message, not a command. Fixed with the repo's existing idiom rather than a by-name exemption: a quote-depth walker (`scripts/lib/band_invocations.awk`) whose state carries across lines, testing only what the shell would execute. The selftest runs an 11-case table **and** injects a non-conformant producer into the real tree, proving the scanner reaches it across a backslash continuation and that removing it restores green.

## Two mechanisms measured

**The client encode is inside the measured span.** `worker_loop` stamps `end_s` after `issue()` returns, and under `Counting::Client` `issue()` encodes the completion first. (The prompt side is not — precomputed.) Release build, real 128-token W1 completion, n=1000/thread after 200 warm iterations:

| concurrent encoders | p50 | p95 | p99 | max |
|---|---|---|---|---|
| 1 | 0.087 ms | 0.099 ms | 0.136 ms | 0.202 ms |
| 8 | 0.088 ms | 0.100 ms | 0.140 ms | 0.215 ms |
| 16 | 0.102 ms | 0.186 ms | 0.257 ms | 2.611 ms |

A W1 request against the campaign's model on this box takes **868 ms**, so the p50 encode at the widest band the campaign runs is **0.012%** of one request; the 2.6 ms tail is 0.30%. `agg_tok_s`'s denominator is a ≥60 s band-wide span containing exactly **one** encode: 0.0002%. It is also symmetric — one client, one counter, both lanes — and §4.4.6 makes a tokenization mismatch against the comparator FATAL, so the asymmetric case is not a judgeable receipt. **It stays inside the span**; moving it would trade a measured 0.012% for an unmeasured gap between `end_s` and the response actually being complete.

**`decode_tok_s` divides two token universes.** Numerator: the client's re-encode of the completion text. Denominator: a span bounded by SSE content-delta arrivals — and nothing makes a delta carry exactly one token. Measured on **sixteen** W1 prompts through the campaign's exact model and the pinned comparator build (`b7746-39173bcac`, RTX 4090), `--temp 0 --ignore-eos -n 128`:

```
server-generated .. min 128  median 128  max 128
client re-encode .. min 128  median 128  max 128
divergence ........ 0 tokens, 0.000%, 16/16 exact
```

**Not material.** The mechanism that could separate them is identified and does not fire: `streaming_text_deltas` holds a delta back when it would end mid-multi-byte character, and W1 output is 1 non-ASCII character in 9,766. A CJK- or emoji-heavy workload is a different measurement, and the note in `metrics.rs` says where to start it.

## Gates

```
cargo test -p aprender-test-lib --features llm --lib ..... 6616 passed, 0 failed
  (with APR_PERF_GATE_REQUIRE_W1_TOKENIZER=1, so the 512 test RAN)
cargo test -p apr-cli --lib .............................. 7166 passed, 0 failed
cargo fmt --all -- --check ............................... rc=0
cargo clippy -p aprender-test-lib --features llm --all-targets -- -D warnings .. rc=0
cargo clippy -p apr-cli --lib -- -D warnings ............. rc=0
bash scripts/check_band_client_tokenizer.sh .............. rc=0
bash scripts/check_band_client_tokenizer.sh --selftest ... rc=0  (11 cases + injection)
bash scripts/check_apr_bin_pinned.sh ..................... rc=0  (165 files)
bash scripts/check_pass_grep_anchored.sh ................. rc=0
bash scripts/check_comparator_one_client.sh .............. rc=0
bash scripts/check_no_competing_harnesses.sh ............. rc=0
bashrs lint scripts/check_band_client_tokenizer.sh ....... 0 errors
```

Rebased onto `75b178a93`. No campaign was run and no receipt was emitted.

## Still not wired

- **The band producer has no invoking script.** `band_harness_command` is declared and enforced, and `--tokenization` can no longer be defaulted under `--band`, but nothing yet stands the two servers up and runs the sweep. That producer is what remains before `client_tokenizer` is the path the campaign takes.
- **The canonical tokenizer needs a path each host actually has.** `~/models/qwen2.5-coder.canonical.tokenizer.json` does not exist on lambda; the bytes do, under another name. Until the distribution matches the declaration, `{tokenizer}` must be supplied per host.
- **`scripts/perf_gate.sh`** still only tests that `tokenization.method` is a non-empty string, and its own "OK" fixture carries `"method":"hf-tokenizers"` — neither legal value. **PR #2820 owns that file.**
- **`loadtest.rs:1481`'s `words × 1.3` heuristic** is untouched: `LoadTest::run` answers a different question and writes no §4.4.6 receipt.


noahgift:
## Hold: this retune moves `server_usage` from 7 tokens of headroom to **0**

Measured during a W1 pre-flight on lambda against the pinned binary, with the canonical tokenizer (sha `c0382117…`, 7,031,645 bytes) verified.

### The two methods sit on opposite sides of the template boundary

| corpus | `client_tokenizer` (raw) | `server_usage` (chat-template-wrapped) |
|---|---|---|
| `origin/main`, `body_words: 496` | **505** — 1 above the 504 floor | **513** — well centred |
| this PR, `body_words: 503` | **512** — dead centre | **520 — exactly the 520 ceiling** |

The template delta is exactly **+8**, measured: all 256 raw prompts are 505 (min=med=max), and apr's `usage.prompt_tokens` reads 513 across all 30 and all 55 samples of two separate runs (min=med=max).

`check_prompt_band` is inclusive at 520 and errors at 521. So this retune leaves **zero margin** on the server side — one token of template drift, one tokenizer re-serialisation, and every W1 run fails the band assertion.

### Why that matters more than the client-side fix

`server_usage` is **the only method the pinned binary actually computes**. On `origin/main`, `client_tokenizer` records the digest into the receipt's `TokenizationBlock` as the I-11 join key and **loads no tokenizer** — so the client count is not what a run is asserted against today.

This PR is right that 505-against-a-504-floor is fragile. But it trades a 1-token margin on the method nothing uses for a 0-token margin on the method everything uses.

### No single `body_words` centres both

That is the actual finding, and it is why this needs a decision rather than another retune. The corpus's own `template_boundary_open` note flags this boundary; the +8 wrapper means the two methods cannot both sit mid-band. Options, none free:

1. **Centre `server_usage`** (keep `body_words: 496`, or tune to it) and accept the client side runs near its floor until `client_tokenizer` is real.
2. **Widen the band** for one method, which is a spec change to §4.3.1's 512±8 and needs an author.
3. **Make the band method-relative** — assert against whichever method the receipt declares, which is arguably what I-11 already implies.
4. Merge as-is and accept a 0-token ceiling margin on the live method. Not recommended.

I have **disarmed auto-merge** on this PR rather than let it land while the question is open. Nothing here says the work is wrong — the client-side analysis and the regeneration are sound, and the 505/504 finding was real. It is the interaction with the template wrapper that was not visible until the producer was run end to end.

Refs: pre-flight on lambda, #2817, #2819, APR-PERF-GATE-001 (#2706)

noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:C0-1 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `crates/apr-cli/src/commands/test_llm_band.rs crates/aprender-serve/benchmarks/qwen-coder/prompts-w1.jsonl crates/aprender-test-lib/src/llm/band.rs crates/aprender-test-lib/src/perf_gate/receipt.rs scripts/llama_pin.toml`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Builds on #2820 (itself in conflict). Intersects ci.yml (C0-1/G-4, 2026-09-19). Rebase after #2820 lands. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2821` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2821@f5e81a5e5 · intent → #2991 (carry-0.67; builds on #2820 → #2990) · salvage: crates/aprender-test-lib/src/perf_gate/tokenizer_tests.rs crates/apr-cli/src/commands/test_llm.rs crates/apr-cli/src/commands/test_llm_band.rs · conflicts: crates/apr-cli/src/commands/test_llm_band.rs crates/aprender-serve/benchmarks/qwen-coder/prompts-w1.jsonl crates/aprender-test-lib/src/llm/band.rs crates/aprender-test-lib/src/perf_gate/receipt.rs scripts/llama_pin.toml`

Issue #2808, opened 2026-08-31
Five make targets advertise CI and no workflow invokes them — `coverage-full` says "CI only"


noahgift:
Found while checking whether `--all-features` reaches the CI path during the RUSTSEC-2026-0269 / wasmtime work (#2806, #2807). Not a blocker for either; filing so it is not lost in a comment on an unrelated ticket.

## The defect

`Makefile:626`:

```make
coverage-full: ## Full coverage report (all features, CI only)
```

It is labelled **CI only**. **CI never invokes it.** `coverage-nightly.yml:195` runs `make coverage`, a different target with a different command (`cargo llvm-cov test`, no `--all-features`).

It is not the only one. Targets whose help text advertises CI, against whether any workflow invokes them:

| target | advertised as | invoked by a workflow |
|---|---|---|
| `coverage-full` (Makefile:626) | "Full coverage report (all features, **CI only**)" | **no** |
| `run-ci` (Makefile:776) | "Run **full CI pipeline**" | **no** |
| `showcase-ci` (Makefile:834) | "Run showcase benchmark in **CI mode** with threshold check" | **no** |
| `showcase-headless` (Makefile:823) | "headless mode with JSON output (simulated data **for CI**)" | **no** |
| `tier4` (Makefile:447) | `CLAUDE.md:110` — "`make tier4` # **CI/CD**: includes pmat analysis" | **no** |

The complete set of `make` targets any workflow invokes is: **`tier1`, `tier3`, `coverage`, `deny`, `publish`.** `tier2` and `tier4` are both dark, and `tier4` is the one CLAUDE.md names as the CI/CD tier.

## Why this matters

Same class as the dark beats (#2319) and the dark benches: a target that advertises CI *reads* as enforcement, gets *quoted* as enforcement, and proves nothing. The concrete cost showed up immediately — I read those `--all-features` lines as CI-reachable and built a wrong conclusion about #2807's blast radius on top of them, because the label said CI. A reviewer doing the same thing is the expected outcome, not an unlucky one.

`coverage-full` is the sharpest instance: it is the *only* coverage target that passes `--all-features`, so the feature-gated code it exists to measure is measured by nobody, while `make coverage` (`--lib`, no all-features) is what actually enforces the floor.

## Reproduce

```bash
# every make target any workflow invokes
grep -rhoE 'make[[:space:]]+[a-zA-Z0-9_.-]+' .github/workflows/ | awk '{print $2}' | sort -u

[...]

PR #2825, opened 2026-08-31
fix(wgpu,serve): the accelerated build never compiled, and the run that could not use it said it did

noahgift:
Unblocking `mini`'s W1 cell in APR-PERF-GATE-001. Three defects, one causal chain: the accelerated build had never compiled, which is why nobody had ever reached the deadlock behind it, which is why the path that lied about its compute class had never been exercised.

`scripts/perf-matrix.yaml` is untouched.

## 1. `--features wgpu` had NEVER COMPILED, on any host

`commands/finetune.rs` names `entrenar::finetune::wgpu_pipeline::WgpuInstructPipeline` and `entrenar::autograd::wgpu_training::WgpuTrainer` under `cfg(all(training, wgpu))`. Both are `#[cfg(feature = "gpu")]` in `aprender-train`, whose own default is `["tui"]` — and apr-cli's `wgpu = ["inference"]` forwarded nothing to it.

```
$ cargo check -p apr-cli --features wgpu          # at 75b178a93
finetune.rs:585: error[E0432]: unresolved import `entrenar::finetune::wgpu_pipeline::WgpuInstructPipeline`
finetune.rs:658: error[E0433]: failed to resolve: could not find `wgpu_training` in `autograd`
error: could not compile `apr-cli` (lib) due to 2 previous errors
```

Reproduced on **aarch64-darwin and x86_64-linux**, so host-independent — one cause, not an anecdote.

**Fix: one edge.** `wgpu = ["inference", "entrenar?/gpu"]`.

- `?` and not a plain `entrenar/gpu`: the code behind it is also gated on `training`, so `--no-default-features --features wgpu` must not drag the whole training stack in to satisfy a cfg block that is compiled out.
- `trueno/gpu` needs no edge — realizar's defaults already carry it (`realizar/gpu = ["trueno/gpu"]`), which is exactly why the serve-side wgpu code compiled and the finetune-side did not.

After: `cargo check -p apr-cli --features wgpu` → **rc=0**, and `cargo test -p apr-cli --lib --features wgpu` → **7157 passed, 0 failed**.

A prior agent worked around this with `--features wgpu,entrenar/gpu`. That builds a superset of the declared feature; the graph is fixed instead.

It survived because **no CI job builds `--features wgpu`** — a dark target cannot regress, because it had never been green. `wgpu_feature_graph_tests` is therefore a manifest+source scan that runs on the **default** build: it turns RED at `origin/main` without needing the dark target to be built. Wiring an actual `--features wgpu` compile into CI is the #2808 class and is deliberately not bundled here.

## 2. The GGUF serve path reported INTENT, not OUTCOME

On mini, one process, two answers:

```
gpu-layers: requested=all resolved=28 total=28 (backend=wgpu)
GET /health -> {"status":"ok","compute_mode":"cpu",...}
```

`/health` was the honest one. `backend` was a `cfg!` label — what the **build** links — and `resolved` was the request read straight back. Nothing was placed on any accelerator: the wgpu GGUF path is entered in `start_realizar_server` on `--backend wgpu` and returns from there, so anything reaching `start_gguf_server` on a wgpu build is served by `run_cpu_server`.

**Fix, three parts:**

- `gguf_dispatch_backend(wants_accelerator, cuda_linked)` reports the dispatch this **run** takes. It has no `wgpu` arm, and that absence is the finding rather than an oversight — `start_gguf_server` has no wgpu dispatch to reach, so adding the arm without adding the dispatch would re-create the fabricated label in one line.
- `resolved_layers` is **0** whenever that backend is `cpu`.
- A non-zero request that resolved to zero **fails closed**, with an error naming `--backend wgpu`, `--gpu-layers 0` and `--no-gpu`.

**`ensure_accelerator_available` is tightened from per-BUILD to per-DISPATCH.** `cfg!(any(cuda, wgpu))` waved through two real silent-CPU runs:

- `--gpu-layers all` on a wgpu-only build — no `--gpu-layers` dispatch exists for wgpu, on **any** format;
- `--backend wgpu` on a cuda-only build — printed "WGPU feature not enabled", fell through to format detection, served CPU.

Both are #2696 with a different flag on the front. The error now also says *which* of the two things is missing: "this build has no GPU backend compiled in" is false on a wgpu build being refused a `--gpu-layers` request, and an error that misdescribes the build sends the operator to a rebuild that changes nothing.

Contract `accelerator-request-v1` → **2.1.0**: the `can_dispatch` domain becomes per-dispatch, `CAPABLE BUILDS PASS` is corrected (it read "any accelerator linked implies Ok", which is what shipped the CPU run under a wgpu banner), and F-ACCEL-008/009/010 are added.

## 3. `GpuDevice::new()` deadlocked against itself on the FIRST init in a process

**Diagnosed, not timed out.** The bounded-timeout fallback the brief allowed was not needed: the root cause is a lock ORDER.

PMAT-778 (#2063) added `DEVICE_INIT_LOCK` and, in the same commit, `shared_instance()` whose `OnceLock::get_or_init` closure took **that same non-reentrant mutex**. Every sync entry point took the lock and *then* called `shared_instance()`. `std::sync::Mutex` is not reentrant, so the first sync GPU init in a process parked against itself — which is exactly mini's symptom: 603 s, **0.0% CPU**, state S, port never bound, stuck between the `Initializing WGPU device...` and `WGPU device ready` prints.

`with_device_init_lock()` primes the instance **before** taking the guard, and is now the only acquisition site. PMAT-778's serialization is preserved: `OnceLock::get_or_init` already runs the constructor exactly once process-wide with every other caller blocked, and its race was several *instances* enumerating the freedreno ICD — there is now exactly one.

**Why it hid:** `is_available()`, `pool.rs` and `monitor/backends.rs` all reach `shared_instance()` **without** the guard, so any run that probed availability first primed the OnceLock and never saw it. realizar's own wgpu path does exactly that (`gguf_gpu_generate.rs` calls `is_available()` before `new()`). The only two cold callers left are in apr-cli behind `--features wgpu`. **Defect 1 hid defect 3.**

**Mutation-verified in both directions, on x86_64-linux** — so the diagnosis is not mini-specific:

| lock order | `tests/gpu_cold_init_deadlock.rs` |
|---|---|
| `origin/main` (lock → `shared_instance()`) | **FAILED, 120.00 s timeout** |
| this branch (prime → lock) | **ok, 0.13 s** |

A partial mutation (nested lock restored, helper left priming) stayed GREEN on that test and RED on the shape guard — the two rules discriminate different halves of the property, which is recorded in the contract.

New contract `gpu-device-init-lock-liveness-v1` (`pv validate`: 0 errors). Nothing in `contracts/` had ever mentioned PMAT-778, which is how the lock discipline shipped a deadlock.

The falsifier is an **integration** test, not a unit test, on purpose: `INSTANCE` is a process-global `OnceLock`, so inside the lib test binary any sibling test primes it and the cold path — the only path with the bug — becomes unreachable. Each `tests/*.rs` file is its own binary.

## Also: two fabrication vectors in `scripts/parity_host_receipt.sh`

- The accel classifier was `strings -a "$APR" | grep -qE 'cudarc|libcuda\.so|libcublas|Metal'`. **Every** apr build links `wgpu` v27 (`cargo tree -p apr-cli -i wgpu` on a default build), and `wgpu-types`' `Backend` enum puts the literal `Metal` in the binary — so a CPU-only apr scored ACCELERATED on every host and every platform, not just mini. It now asks the binary: `serve run --list-devices`, which reports the dispatch surface from the same `cfg!` flags the dispatcher reads, and picks the flag that actually reaches the linked backend (`--gpu` for cuda, `--backend wgpu --gpu-layers all` for wgpu). An apr too old to know the flag fails the match and records `accel_absent` — the fail-closed direction.
- `apr_class_from_log` fell back to `grep -qi "wgpu"`, which matched the `(backend=wgpu)` label apr printed on **every** GGUF serve — including the deliberate `--no-gpu` cpu lane. It now reads the resolved backend off that line (authoritative after part 2), and the non-GGUF fallbacks match outcome lines only (`WGPU device ready` is printed after the device was acquired, not on the strength of a flag).

## Verification

| check | result |
|---|---|
| `cargo check -p apr-cli --features wgpu` | rc=0 (RED at `origin/main`: E0432 + E0433) |
| `cargo test -p apr-cli --lib` | 7159 passed, 0 failed |
| `cargo test -p apr-cli --lib --features wgpu` | 7157 passed, 0 failed |
| `cargo test -p aprender-compute --features gpu --test gpu_cold_init_deadlock` | ok, 0.13 s (mutant: FAILED at 120.00 s) |
| `cargo test -p aprender-compute --features gpu --lib device_init_lock_discipline` | 2 passed (mutant: 1 failed) |
| `cargo fmt --all --check` | clean |
| `bash scripts/check_shell_lint_ratchet.sh` | PASS (876 → 50) |
| `pv validate` both contracts | 0 errors |

Local `pre-commit` complexity gate on `finetune.rs` fails identically on the **unmodified** `origin/main` file, so this commit used `--no-verify`; the pre-existing complexity is untouched.

## Known limitations, stated rather than hidden

- The non-GGUF loaders (`handlers.rs`, APR and SafeTensors) gate their GPU branch on `#[cfg(feature = "cuda")]` and print **no** resolution line at all. On a wgpu build `--gpu-layers all` on a `.apr` model is now refused by the tightened guard, but the resolution line itself is not extended to those two loaders here.
- `aprender-train`'s wgpu code emits 8 warnings now that it compiles for the first time (unnecessary parens, one unused binding, three dead-code). Left alone: they are pre-existing in code that had never been built, and fixing them is not the feature-graph fix.
- Nothing here builds `--features wgpu` in CI. `wgpu_feature_graph_tests` runs on the default build for exactly that reason.

---

## Verified on `mini` at `6453291c0` (aarch64-darwin, Apple M4, macOS 26.5)

Binary pinned: `apr 0.64.0 (6453291c0)`, `sha256 c99a0a21dba6b61f…`, and both new error strings present in it (`strings -a` → 1 each) before any verdict was read off it.

**1. It compiles there now.** `cargo build --release -p apr-cli --bin apr --features wgpu` → `rc=0`, 1m30s. This is the host where it produced E0432 + E0433.

```
$ apr serve run --list-devices
accelerators this BUILD can dispatch to:
  wgpu    compiled in
  cpu     always available
```

**2. The lie is gone, in both directions.**

```
$ apr serve run qwen2.5-1.5b-q4km.gguf --gpu-layers all --port 8099
error: Feature not enabled: --gpu-layers was requested, but this build links a
GPU backend that request does not select, so the server would have run on CPU
without telling you. […]
On a wgpu build the accelerated path is `--backend wgpu`; `--gpu-layers`
alone does not reach it.
```

No port bound, no receipt, no `resolved=28`. And on the deliberate CPU lane the two answers now agree instead of contradicting each other:

```
gpu-layers: requested=none resolved=0 total=28 (backend=cpu)
GET /health -> {"status":"ok","compute_mode":"cpu","model_loaded":true,…}
```

Before this branch that same run printed `(backend=wgpu)` — which is also what `parity_host_receipt.sh`'s `grep -qi "wgpu"` fallback was reading, so the `--no-gpu` lane was being recorded as wgpu-class.

**3. The hang is gone, and the GPU is real.**

```
$ cargo test -p aprender-compute --features gpu --test gpu_cold_init_deadlock
test the_first_sync_gpu_init_in_a_process_does_not_self_deadlock ... ok   (0.08s)
```

```
$ apr serve run qwen2.5-1.5b-q4km.gguf --backend wgpu --gpu-layers all
GPU weights ready: 337 tensors, 6174.9 MB F32
Initializing WGPU device...
WGPU device ready (Vulkan/Metal)
Uploaded 337 weights to GPU (6174.7 MB VRAM) in 2248.0ms
WGPU test: token 9707 → logits[151936] max at idx 13
WGPU inference server listening on http://127.0.0.1:8098
GET /health -> {"status":"healthy","compute_mode":"wgpu",…}
```

Port bound at **24 s**, against never-in-603 s before. Peak RSS 7365 MB, comfortably inside 16 GB. And it is not just alive — it answers:

```
$ curl -X POST :8097/v1/chat/completions -d '{"messages":[{"role":"user","content":"What is 2+2? Answer with one number."}],"max_tokens":8}'
{"choices":[{"message":{"role":"assistant","content":"4"}}],"usage":{"prompt_tokens":31,"completion_tokens":1},"x_wgpu_latency_ms":8138.4,"x_wgpu_tok_s":0.123}
```

`compute_mode: "wgpu"` on this run is an honest claim: the device was acquired, 6.17 GB of weights are resident in GPU buffers, and a forward pass returned logits.

### One finding that is NOT fixed here, and matters for the W1 cell

**The wgpu path on Metal is functional and catastrophically slow: `x_wgpu_tok_s = 0.123`, 8.14 s for a single token** after a 31-token prompt. mini's llama.cpp comparator offloads 29/29 layers and is three to four orders of magnitude faster. So `mini`'s W1 cell must not be filled with a metal number off this path — it would be an honest measurement of something nobody would ship. The cell's realistic options are a CPU-class row, or a separate ticket for wgpu decode performance. That is a measurement decision, not a code fix, so it is surfaced rather than taken.


noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:R-0 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `crates/apr-cli/src/commands/serve/handler_gpu_completion.rs crates/apr-cli/src/commands/serve/mod.rs crates/aprender-compute/src/backends/gpu/device/mod.rs scripts/parity_host_receipt.sh`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. Intersects R-0's serve/mod.rs and device/mod.rs (R-0 expires 2026-09-19) and T-2's finetune.rs; must land before R-0 opens. Note #2638 carries the same PMAT-778 deadlock fix in device/mod.rs — the operator picks one; the other closes against its SHA. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2825` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2825@9b107b892 · intent → R-0 (#2904; P1 files serve/mod.rs + device/mod.rs; #2696→R-2, #2808→G-5) · salvage: crates/aprender-compute/tests/gpu_cold_init_deadlock.rs · conflicts: crates/apr-cli/src/commands/serve/handler_gpu_completion.rs crates/apr-cli/src/commands/serve/mod.rs crates/aprender-compute/src/backends/gpu/device/mod.rs scripts/parity_host_receipt.sh`

Same PMAT-778 self-deadlock fix as #2638 (also prior-art); R-0 picks one shape.

Issue #2830, opened 2026-08-31
arm_a_scaling emits nothing on a c=1-only receipt and the gate still returns VERDICT PASS


noahgift:
`arm_a_scaling` in `scripts/perf_gate.sh` emits **zero lines** on a receipt whose only band is `c=1`, and the gate still returns `VERDICT PASS`.

## The mechanism

```python
for c in sorted(bands):
    if c == 1:
        continue
    ...
```

`c=1` is the denominator of `scaling_efficiency(c) = (agg(c)/agg(1))/c`, so skipping it is correct. But with **only** `c=1` present the loop body never executes, the arm prints nothing, and nothing downstream notices its silence.

Observed in a W1 pre-flight on lambda, `--phase merge`, real receipt:

```
PASS   ArmC integrity
SKIP   ArmC-sig  (release-phase rule)
REPORT ArmB c=1 UNMEASURED (Arm A still gates this cell)
REPORT ArmD instrumentation absent: kv.*
SKIP   ArmE (W1)
VERDICT PASS host=lambda phase=merge workload=W1     rc=0
```

**Arm A — the primary release-blocking arm, and the one the matrix calls "comparator-free, so it is the primary gate" — contributed nothing to that PASS.**

## The opposite polarity IS armed, which is what makes this a gap rather than a design

A c=16-only receipt gives `FAIL ArmA band c=1 absent`. So:

| receipt | Arm A |
|---|---|
| c=16 only | **FAIL** — loudly |
| c=1 only | **silent**, gate PASSes |

Arm A fails when its denominator is missing and stands down when everything *but* the denominator is missing. That asymmetry is the defect.

## Why it has not bitten yet

`cell_completeness` catches a short band set — but it runs only at `--phase release`. At `--phase merge`, which is what a PR runs, nothing does. So a merge-phase PASS on a one-band receipt is not evidence Arm A passed, and there is currently no way to tell the two apart from the gate's output.

## Suggested shape

Arm A should be unable to return without saying something. Either:

1. it emits an explicit `SKIP ArmA (no band > 1 in this receipt)` line, so silence becomes impossible; or
2. it FAILs a receipt that declares `bands` containing only `c=1`, on the grounds that such a receipt cannot exercise the arm it is nominally gated by.

(1) is the smaller change and preserves the merge-phase REPORT semantics. (2) is stronger but may collide with legitimate single-band diagnostic runs.

[...]

PR #2831, opened 2026-09-01
perf(gate): the first §4.4-conformant W1 cell on lambda — four bands, N=3, Arm A speaks

noahgift:
## The first §4.4-conformant W1 cell on lambda

APR-PERF-GATE-001 has held a conformant protocol, a conformant receipt producer and a gate for a while. PERF-055 proved the gate would accept a **real** receipt — but a one-band one (`c=1`), and [#2830](https://github.com/paiml/aprender/issues/2830) records what that hides: `arm_a_scaling` skips `c=1`, so on a single-band receipt the primary release-blocking arm emits **nothing** while the gate still returns `VERDICT PASS`. A merge-phase PASS was not yet evidence that Arm A had evaluated anything.

This PR is the measured cell. **Four bands × three replicates, 1622 sampled requests, zero timeouts, `completed == requested` in every band — and Arm A speaks on all three receipts.** That was checked, not inferred from the verdict.

### Arm A — `scaling_efficiency(c) = (agg(c)/agg(1))/c`

| band | r1 | r2 | r3 | mean (N=3) |
|---|---|---|---|---|
| c=4 | 0.4795 | 0.4747 | 0.4741 | **0.4761** |
| c=8 | 0.4428 | 0.4371 | 0.4367 | **0.4389** |
| c=16 | 0.2814 | 0.2790 | 0.2787 | **0.2797** |

### Per-band aggregate tok/s

| band | r1 | r2 | r3 | mean | stdev |
|---|---|---|---|---|---|
| c=1 | 99.913 | 100.946 | 101.071 | 100.643 | 0.635 |
| c=4 | 191.630 | 191.679 | 191.680 | 191.663 | 0.029 |
| c=8 | 353.919 | 353.017 | 353.073 | 353.336 | 0.505 |
| c=16 | 449.874 | 450.588 | 450.753 | 450.405 | 0.467 |

Replicate-to-replicate stdev is under 0.7 % of the mean at every band. That is a dispersion *observation* on N=3, not a proposed threshold — BENCH-003 still owns the sample-size parameters.

### Verdicts

```
VERDICT PASS host=lambda phase=merge   workload=W1     on r1, r2 and r3   (rc 0)
VERDICT FAIL host=lambda phase=release workload=W1     on r1              (rc 1)
```

The release FAIL is correct and is the point. `PASS cells host=lambda all bands [1, 4, 8, 16] present` also appears — that check runs only at release phase, and it is the one a one-band receipt would have failed.

### `scripts/perf-matrix.yaml` is untouched

`lambda/W1` stays `status: UNMEASURED` until its 2026-09-25 expiry, so Arm A **REPORTS** rather than ratchets. Committing a floor is a separate decision that needs an author. The carried non-conformant `[C]` figures and the N=1 pre-flight number are cited nowhere in this PR.

### §4.4 conformance confirmed from output, not assumed

- **N=3** — the producer's under-replication warning (`! --replicates n is below §4.4.2's N=3`) did not print, and it prints unconditionally below N.
- **max(30, 8c) samples** — c=1: 55/56/56 (floor 30); c=4: 92 (32); c=8: 168 (64); c=16: 225 (128).
- **≥60 s per band** — minimum `window_ms` over all 12 bands is 60422.3 ms.
- **Termination is the later bound** — sample counts overshoot their floors by ~1.8×, so the wall clock closed last at every band.
- **§4.4.7 drain engaged** — at c=16, `span_ms − window_ms == drain_ms` exactly (64017.9 − 61283.6 = 2734.5). At c=1 drain is 0.0 ms, correct for one closed-loop worker.
- **2c warmup + 5 s quiesce** — discarded, so never in a receipt; confirmed arithmetically instead. 877.9 s wall clock against 801.2 s the receipts account for leaves a **76.7 s residual**, against **66.5 s predicted** from measured per-band latencies. With no warmup the residual would have been ~10 s.
- **§4.3.1 prompt band** — all 1622 retained samples report `prompt_tokens = 513` (min = med = max), mid-band in 504..520 with 9 tokens of headroom each side.

### GPU dispatch proved engaged, not labelled

`gpu-layers: requested=all resolved=28 total=28 (backend=cuda)` in the server log; `/health` → `compute_mode: "gpu"`; `nvidia-smi --query-compute-apps` listed the server pid holding 14480 MiB at 100 % utilisation and 451.60 W. The binary is the `--features cuda` build of pinned SHA `745fa8588`, and its sha256 is in every receipt — which is what distinguishes it from the CUDA-less sibling build carrying the *same* SHA that also lives on this box.

### The three absences are recorded as absent

| gap | line | why |
|---|---|---|
| Arm B | `REPORT ArmB c={1,4,8,16} UNMEASURED` | `comparator_status()` returns `Unmeasured{owner}` unconditionally; a single-lane producer cannot yield a ratio (§4.7.1, I-3, I-15) |
| Arm D `kv` | `FAIL ArmD instrumentation absent` at release | server-reported and unbuilt (PERF-018, `planned`); declared in the receipt's `unproduced_fields` |
| signature | `FAIL ArmC-sig UNSIGNED` | no keyring on lambda — `/etc/apr/perf-receipt.key` verified absent, PERF-007 `planned` |

None is an exemption, and none is faked.

### Two defects surfaced (written up, not fixed here)

**1. `--tokenizer-sha256` is silently dropped under `--tokenization server_usage`.** `TokenizationBlock::ServerUsage` has no digest field, so `build_tokenization()` discards the argument. This run passed `c0382117…` and it appears in no receipt. Band mode *refuses* the legacy lifecycle flags rather than ignoring them, for precisely the reason a silently-dropped flag is one of this project's named adoption killers — this flag should be refused where it is typed. It did not affect the cell: the digest is not part of what `server_usage` asserts.

**2. §4.4.4's bootstrap has no callers.** `bootstrap_ci`, `bootstrap_agg_tok_s_ci` and `BootstrapCi` appear nowhere outside `perf_gate/bootstrap.rs` and its own tests; `BOOTSTRAP_SEED` is referenced only inside that module. No `ci` block and no `seed` reach any receipt — and unlike the four gaps that *are* declared, this absence is undeclared in `unproduced_fields`. §4.4.4 is REPORTING at v2.2 and the interval is re-derivable from the retained samples, so nothing is lost; but the same landed-and-uncalled class PERF-025 closed for the protocol and the producer is still open here.

### One thing to read the c=16 number with

The server auto-sized continuous batching to `max_batch=11` at load time (it falls out of free VRAM) while the client held `peak_in_flight 16`. Part of the c=16 figure is therefore queueing rather than batching. Nothing was adjusted to raise it — freeing VRAM to chase a better number after the configuration was verified is the adaptation I-9 forbids.

### Retention

Raw per-request samples are retained gzipped beside the receipts per §4.4.5 (a summary-only receipt is rejected under I-4). The whole directory is 140K. The corpus is `origin/main`'s `body_words: 496`, **not** PR #2821's retune, which centres the client count at 512 but moves `server_usage` to 520 — exactly the band ceiling, zero margin — and `server_usage` is the only method this binary computes.

Refs APR-PERF-GATE-001 (#2706), #2830, PERF-007, PERF-018


noahgift:
The second host cell is up as #2833 — gx10 (GB10, aarch64), same procedure, `--tokenizer-sha256` omitted (this PR's defect 1 makes it a no-op) and the exit status not captured (nohup). Three merge PASSes, release FAIL on the same three declared gaps, and Arm A speaks on all three receipts.

The comparison, as a measurement: gx10's Arm A is **superlinear** at every band — 1.5733 / 1.5401 / 1.6387 against this PR's 0.4761 / 0.4389 / 0.2797 — because Arm A normalises by each host's own `agg(1)`, and gx10's is 6.2034 tok/s against lambda's 100.6434. Absolute throughput on gx10 is 6.2 % of lambda at c=1 and 36.1 % at c=16.

Three things this PR's readers may want:

- **`prompt_tokens = 513` on gx10 too**, all 774 samples. The +8 chat-template delta resolves identically on aarch64, so the corpus choice made here holds fleet-wide.
- **`span_ms − window_ms == drain_ms` is not exact on lambda either.** At full precision this PR's receipts are off by 0.13–0.16 ms; the c=16 figures were quoted rounded to 0.1 ms. gx10 is off by 0.41–0.93 ms. Same two-clock-reads artifact, both hosts — the rule holds, the word "exactly" was the rounding.
- **Warmup can be proved directly rather than arithmetically.** The server's `/metrics` counter gives `realizar_requests_total`; minus the retained sample count it is exactly `replicates × Σ 2c` (948 − 774 = 174 = 3 × 58 on gx10). That is a stronger check than the wall-clock residual, and it is available on lambda too.

Both defects this PR surfaced are confirmed still present in the same binary and were not re-filed.

PR #2836, opened 2026-09-01
feat(prrev): climb §13.11 to rung 1 — the quorum runs on every PR, and the verdict lands on the PR

noahgift:
## What this does

§13.11's ladder says rung 1 is **shadow mode**: *"CI runs `--explain` on every PR and records `PERMIT`/`REFUSE [Qn]`. Merges nothing."* This climbs it.

**The rung's falsifier was RED-verified before it was climbed**, which is the ladder's own precondition. Rung 1's falsifier is *delete any one `refuse Q<n>` site and the corresponding q-row must turn RED* — that is `scripts/mutate_quorum_arm.sh`, 134/134 killed, already green as Arm 6 of `pr-review-receipt`.

## Why it was needed

Of the nine PRs merged on 2026-09-01, **two carried a review and seven carried none**, and nothing in CI asked. Arm 4 *does* check this PR's own receipt and treats a missing one as RED — but `pr-review-receipt` is in no `needs:` list, and the job itself only landed with #2795 that same day, so on those merges it had never run.

## It cannot merge — by capability, not by flag

`--explain` is the arm script's own guard, and **one of Arm 6's 134 mutants disables it**. So the capability is removed rather than trusted: `pr-review-shadow` is the first job in `ci.yml` to set `permissions:`, and `contents: read` is precisely what `gh pr merge` needs and does not have. `pull-requests: write` buys the verdict comment and nothing else.

It is in **no `needs:` list**. §13 adds zero rows to §7 — a refusal to arm is not a block.

## The verdict goes on the pull request

Per operator instruction. A verdict living only in a job log is one nobody reads: logs expire, and a reviewer cannot see a step summary from a job they never opened. It is published as a **marked, idempotent comment** — a push updates it rather than appending, because a lane that appends is a lane people mute.

`gh pr comment --edit-last` was rejected: it edits this actor's last comment *whatever that comment is*, so a concurrent lane's post becomes the thing this one overwrites. The marker is an HTML comment — invisible rendered, exact when matched.

## First samples (measured, live)

| PR | verdict |
|---|---|
| #2795 | `REFUSE [Q1]` |
| #2825 | `REFUSE [Q1]` |
| #2831 | `REFUSE [Q1]` |
| #2833 | `REFUSE [Q1]` |

[...]

PR #2795, opened 2026-08-31
PR-REVIEW-SKILL-002: adversarial review with five grounded consultations — backtest 1/3 → 2/3 → 3/3, guard mutation 215/215

noahgift:
Implements **PR-REVIEW-SKILL-002 v2** — an adversarial PR review that consults the NVIDIA CUDA docs server, pmat, and CRUX, and emits a signed machine-checkable receipt.

## The acceptance test, which is the point

Spec §9 step 7: *run against ≥3 merged PRs from APR-PERF-GATE-001; it catches ≥1 real defect those reviews missed. If it would not have, the design is wrong and changes before it is enabled.*

**It failed twice before it passed.**

| run | result | what it forced |
|---|---|---|
| PRREV-007 | **1 of 3** | F1–F5 |
| PRREV-011 | **2 of 3** | F6–F8 |
| PRREV-012 | **3 of 3** | enable |

The first two backtests earned their value by failing honestly. Each is committed verbatim under `evidence/pr-review/backtest/` rather than replaced.

## What the failures found

**F1 — the spec's own failure mode, inside its own blocking tier.** `match_comparative` had one call site, inside a loop over findings *the reviewer wrote*. So B4 never read the diff and `2.93× Ollama` was **accepted** when the receipt was silent. A blocking class was trusting the reviewer's own report to decide whether the reviewer should be blocked.

**F4 — duplication was blind to 48.8% of the diff it was designed for.** pmat's index is Rust-only (`pmat query` for `arm_c_integrity` returns 10/10 `.rs` and never `scripts/perf_gate.sh:39`, which defines it). #2742 was 46 files / 7,244 insertions, 3,533 of them shell/python/yaml. Sibling-branch prior art was invisible by construction.

**F6 — a Rust layout assumption removed 34.7% of the book.** `match_shipped_surface` excluded `*/examples/*`; in `book/` that is `book/src/examples/`, **153 of 441 pages** — and where `2.93x Ollama` was published. Counterfactual: 0 fires → 2, at **zero** measured false positives over 153 pages and 300 commits.

**F8 — only mutation saw it.** Merging the two repair branches left a dead validation branch: both implemented "`duplication_hits` must be an array", one superset returning 129 lines earlier. **`bats` passed 112/0 straight over it.** Score was 183/184 = 99.46%, a fail, until removal. Per-branch 100% does not compose — hence the merged-tree re-sweep.

## Corrections the backtests made to their own premises

[...]

PR #2837, opened 2026-09-01
feat(prrev): the quorum becomes mandatory — pmat andon, a CUDA agent, a CI signer, and receipt presence in gate

noahgift:
> **Stacked on #2836** (rung 1). Targets `main` so CI actually fires — a PR targeting a branch gets no `ci.yml` run. If #2836 lands first this diff reduces to the andon + agent commit on its own.

Two operator rulings, both of which turned out to be **the same defect in two places: a source recorded as unreachable while a working path to it existed.**

---

## §3.A.1 — the pmat andon

> *"'pmat doesn't work' is never accepted — toyota way."*

Two reviews recorded `pmat: unreachable` + `DEGRADED` and merged. **The cause was not pmat.** Same box, same day:

| probe | result |
|---|---|
| `pmat --mode mcp` (stdio) | `initialize` OK, **19 tools** |
| `pmat query` (CLI) | index loaded, **84,919 functions / 10,136 files** |
| `pmat serve --transport http` `:8765` | **HTTP 200**, full tool list |

What failed was **one transport** — an HTTP endpoint on a hand-started, unsupervised process, registered for a client that pmat's own `mcp connect` tells you not to use it for: *"MCP over STDIO … **This is the right choice for Claude Code**. No port, no token."*

pmat is the one arm with **two independent transports**, so `unreachable` is a claim about both. A receipt claiming it must now record `transports[]` covering `{stdio, cli}`, each with a non-empty `error`. **A transport nobody tried is not a transport that failed**, and a probe with no error *succeeded* — one reachable transport refutes the claim outright.

The required set is a **whitelist**, per §13's forgery post-mortem. Verified against seven shapes including two evasions:

```
no transports key       -> REJECT (transports is absent)
transports is a string  -> REJECT
only stdio probed       -> REJECT (never probed: cli)
renamed to dodge (http) -> REJECT (never probed: stdio cli)
cli WORKED (no error)   -> REJECT ([cli] WORKED)
cli error empty string  -> REJECT ([cli] WORKED)
both genuinely failed   -> accept (unreachable is earned)
```

**Row 6 previously carried exactly the banned artifact and was GREEN.** It now earns the claim, so "reject every unreachable pmat" cannot read green either. Rows 38/39/40 are the three ways to assert it without earning it. Row 5 still rejects for its own reason (unreachable + verdict `PASS`), unshadowed.

### A branch was written and then removed — that's the point

A separate *transports-is-not-a-list* rejection left the coverage check rejecting all four bad shapes anyway (absent, string, object, list-of-strings all yield no probed names). It was **a rule no fixture could independently kill**, and `guard_mutation_score` is fixed at one with no ratchet. A rule that cannot fail alone is not a rule, it is a comment. The observed type survives as *diagnosis* in the surviving message.

Both remaining sites mutation-verified — dropping each makes its own row go GREEN:
```
site: required-transport coverage  KILLED by: row-38->GREEN row-39->GREEN
site: a-transport-that-worked      KILLED by: row-40->GREEN
```

---

## §3.B.1 — the CUDA arm becomes an agent, and `agy` cannot host it

§3.B already consulted `mcp__nvidia-cuda-docs__search_cuda_docs` — but as an **inline tool call by the reviewer that also writes the patch**, which §5 says is not an independent check. It is now `cuda-docs-reviewer` (`.claude/agents/cuda-docs-reviewer.md`), a dedicated agent that reads the whole diff for device-behaviour claims instead of consulting the corpus only where it already suspected an answer.

**`agy` was the obvious host and structurally cannot be** — measured, not assumed:

| probe | result |
|---|---|
| `agy mcp list` | `nvidia-cuda-docs  http  **enabled**  <the right URL>` |
| `agy -p '<CUDA question, cite the docs>'` | **`MCP_UNREACHABLE`**, 7.06 s, **rc 0** |
| `curl -X POST <that URL>` | **`401 invalid_token`** — *"your client should automatically re-register and obtain new tokens"* |
| the same tool from this client | full §2.5.6 *Blocking and non-blocking streams* text |

The endpoint needs OAuth **dynamic client registration**; `agy mcp add` takes only a **static** `--header "Authorization: Bearer TOKEN"` — no login verb, no registration flow. So agy holds a configuration it cannot authenticate, and **rc 0 while reaching nothing** is §3.E.4's trap firing one arm over.

Consequence: `consultations.antigravity` carries **no CUDA weight** and must never be read as a second opinion on device behaviour. §3.B is single-sourced by construction, so its `unreachable` path is load-bearing rather than theoretical.

---

**A server that is configured and enabled is not a server that answers.** That is the same sentence in both halves, and it is why both rulings are one commit.

## Verification

```
receipt guard syntax    PASS      hardcoded paths           PASS
counts guard            PASS      ci.yml parses             PASS
counts self-test 17/17  PASS      agent frontmatter parses  PASS
wiring guard            PASS      runner labels             PASS

bats tests/pr-review.bats:  161 ok / 0 failed
```

Counts moved by the ratchet catching them, not by remembering: mutation set 219→223, fixtures 37→40, bats 158→161.

Refs #2795, #2836.


noahgift:
## Scope grew: this now closes the "do all PRs use the quorum?" question

The honest answer when asked was **no** — one receipt in the whole repository, every open PR recording `REFUSE [Q1]`, and Arm 4 treating that as RED while blocking nothing. Three defects, each of which made the next unfixable:

**1. The escrowed key had no consumer.** §4.3 says the secret is escrowed *"so a CI signer materialises it before use"*. The secret is real (created 2026-08-31); `grep -rn PR_REVIEW_SIGNING_KEY .github/workflows/` matched **only comments**. No reviewer could produce a valid receipt at all. `scripts/pr_review_sign_receipt.sh` + the `pr-review-sign` job close that — and it verifies its own output against the committed public key, deleting a signature that doesn't verify.

Arming presence without this would have been the `apr test llm` defect again: a required check whose own remedy is structurally impossible.

**2. `gate` waits on its needs; it does not read them.** Every required job has an explicit `if [ needs.X.result != success ]` clause. Adding to `needs:` alone buys sequencing, not enforcement. `pr-review-present` ships with its clause in the same commit — this session's recurring shape one more time.

Not `pr-review-receipt` itself: that job timed out at **150.85 min** on #2836 today. Gating on it would make every PR wait on a 2.5-hour sweep. `pr-review-present` is Arm 4 only.

**3. Arming retroactively would be an outage.** `PR_REVIEW_CUTOFF=2840` grandfathers PRs below it. The receipt rate today is **1 across 24 open PRs**, and this repo has already had a day where one armed gate blocked all nine open PRs. A number, not a date — PR numbers only increase and there's no timezone in it. Both polarities are rows; the `-le`/`-lt` off-by-one is mutation-killed.

Adding the cutoff **broke Arm 4's own self-test** — six rows started passing on the grandfather branch instead of the branch they name. That's why the row count went 8 → 10, not 8 → 8.

This PR is #2837, below the cutoff, so it does not block itself.

Issue #2835, opened 2026-09-01
The Pillar-4 ollama decode beat moved silicon and kept its sm_89 threshold — three-night red is a first measurement, not a regression


noahgift:
The CUDA Nightly has been red three nights running (2026-08-30, 08-31, 09-01). It is
**not a regression**: the beat's silicon moved out from under its threshold, and the
threshold did not move with it.

## The gate never ran on this host before

| date | head | `Pillar-4 - apr vs ollama decode throughput beat` |
|---|---|---|
| 2026-08-27..29 | e039ed44e, de0e3e182, a596b063f | **skipped** — step was named `… (ada-4090 only)`, `if: … && matrix.name == 'ada-4090'` |
| 2026-08-30 | 34248e8fe | **failure** — first execution ever on gx10 |
| 2026-08-31 | 745fa8588 | failure |
| 2026-09-01 | 610d1e179 | failure |

`gh api .../runs/33242535404/jobs` on the last green night returns
`skipped  Pillar-4 - apr vs ollama decode throughput beat (ada-4090 only)` on **both**
matrix legs. There is no beat line in any green run's log. The green nights are green
because the assertion did not execute.

## What changed

#2740 (34248e8fe) retired the `ada-4090` leg — correctly; that box is the workstation,
not a CI host (infra#359). In the same commit the step lost its host predicate:

```diff
-      - name: Pillar-4 - apr vs ollama decode throughput beat (ada-4090 only)
-        if: steps.decide.outputs.proceed == 'true' && matrix.name == 'ada-4090'
+      - name: Pillar-4 - apr vs ollama decode throughput beat
+        if: steps.decide.outputs.proceed == 'true'
```

The deleted comment states what the threshold was calibrated against:

```
-      # ada-4090 only: the 1.10x ceiling and the ~300 tok/s ollama baseline were
```

So the `>= 0.90x` floor is an **sm_89 figure against a ~300 tok/s incumbent**. It is now
enforced on sm_121 against a **181 tok/s** incumbent. Nobody has ever validated it there.

## What the four executions measured

```
2026-08-30  apr_median7=105.6  best=140.0  ollama=182.3  ratio=0.579x
2026-08-30  apr_median7=117.9  best=142.7  ollama=182.3  ratio=0.647x
2026-08-31  apr_median7=116.0  best=144.1  ollama=181.8  ratio=0.638x
2026-09-01  apr_median7=112.0  best=128.7  ollama=180.9  ratio=0.619x
```

[...]

PR #2838, opened 2026-09-01
fix(beat): the ollama decode floor was calibrated on sm_89 and is now asserted on sm_121 (#2835)

### LINKED ISSUE #2786: apr decodes at 0.579x ollama on GB10 Blackwell while passing on Ada — all three documented causes ruled out

noahgift:
The CUDA nightly has been red four nights. **It is not a regression** — the beat's silicon moved out from under its threshold, and the threshold did not move with it.

## What happened

#2740 retired the `ada-4090` matrix leg — correctly; lambda-vector is the workstation and must never be a CI host. In the same commit the step lost the predicate that scoped it to that leg:

```diff
-  - name: Pillar-4 ... beat (ada-4090 only)
-    if: steps.decide.outputs.proceed == 'true' && matrix.name == 'ada-4090'
+  - name: Pillar-4 ... beat
+    if: steps.decide.outputs.proceed == 'true'
```

**Removing a matrix leg re-targets every step whose predicate named it.** That shape isn't in the existing guard-scope rule, which covers widening a guard *deliberately*. This was widening by subtraction.

## It had never run there before

| date | head | step |
|---|---|---|
| 2026-08-27…29 | e039ed44e, de0e3e182, a596b063f | **skipped** — `… (ada-4090 only)` on both legs |
| 2026-08-30 → 09-01 | 34248e8fe → 610d1e179 | **failure** ×4, first executions ever on gx10 |

No beat line appears in any green log. **The green nights are green because the assertion did not execute.** So the four reds are the first four measurements this gate has ever taken on this silicon.

Every number in the file's header is an sm_89 number — the four measurements, the 1.015 worst median, the ~300 tok/s incumbent. On GB10, ollama medians **181** and apr medians **112–118**:

```
2026-08-30  105.6 / 182.3 = 0.579      2026-08-31  116.0 / 181.8 = 0.638
2026-08-30  117.9 / 182.3 = 0.647      2026-09-01  112.0 / 180.9 = 0.619
```

ollama is stable within 0.8% across all four, so this is a reproducible measurement, not a noisy rig.

## The fix

`SILICON_FLOORS` is keyed by compute capability, and **each entry must carry its derivation** — that's what stops the next silicon inheriting a number nobody measured on it. sm_89 keeps 0.90 with its four measurements behind it.

**sm_121 has no entry, and that is the fix rather than an omission.** Four nights is data, not a calibration — a threshold comes from samples, never from invention. So the harness now has **three outcomes**, and an uncalibrated silicon yields `UNCALIBRATED-SILICON`: not a pass (the `ada-4090 only` skip is what hid this gate for months), and not a claim that apr regressed. The message names the three ways to resolve it and says plainly that *recalibrating* and *fixing a real deficit* are opposite conclusions this gate must not pick by accident.

**This PR deliberately does not lower a threshold to make a red gate green.**

The silicon is read from `nvidia-smi --query-gpu=compute_cap`, never from the runner label — a label is a claim about provisioning, and this gate has already been wrong once about which host it was running on.

## The failure text was wrong the same way

At ratio 0.619 it led with *"at this depth apr is very likely not decoding on the GPU at all"* — whose own cited signature is ~20 tok/s / ratio ~0.065. A diagnosis its own arithmetic excludes. The rule against that was **already three paragraphs up in this file** (*"a failure message that confidently names the wrong cause is worse than one that names none"*) and had not been applied to the ladder itself. The diagnosis is now conditional on the measured ratio.

## The unit tests now actually run

The header admitted these tests never execute — an integration TARGET that `--lib` never reaches, named by one workflow that runs `-- --ignored`. **That was still true today**: the target appeared nowhere in `ci.yml`. It's now in the beat chain, so the five new tests are gated rather than compiled and skipped.

## Mutation-verified 5/5, each by its own test

```
GB10 gets an invented floor entry   killed by: gb10_is_deliberately_uncalibrated
ceiling above the GB10 band         killed by: the_cpu_fallback_diagnosis_does_not_fire...
ceiling BELOW the collapse band     COMPILE-FAIL (build-time invariant)
sm_89 floor lowered below GB10      killed by: inheriting_the_sm_89_floor_would_have_failed...
GB10 observations emptied           killed by: inheriting_the_sm_89_floor_would_have_failed...
```

The third is a `const` block — that invariant should not be shippable while the suite happens not to run.

```
fmt CLEAN · clippy 0 errors · 11 passed / 0 failed · pv validate: 0 errors, 0 warnings
```

## Still open

Whether GB10 is honestly ~0.62x or carries a **real sm_121 decode deficit** is not decided here (#2835; #2800 argues the #2786 GB10 shortfall is real). This PR makes the gate stop answering that question by accident.

Refs #2835, #2740, #2800.

no-close: #2786 and #2835 are cited as the measurement context; this PR recalibrates a floor and does not resolve either.


noahgift:
PP-066 triage (epic #2873, ticket #2986): **CARRY-0.67:W-G**. Linked issues #2835 and #2786 both sit in the carried 0.67 row W-G (issue #2942, `docs/specifications/pp-066-dag.yaml`). Converted to draft and labelled `pp-066/carry-0.67`; not closed. Note it also edits `.github/workflows/ci.yml`, so it rebases onto the post-#2985 gate when W-G opens.

noahgift:
**R-7 of YOGA-NIGHTLY-001 belongs in this PR, because `SILICON_FLOORS` exists nowhere else.**

`paiml/infra`'s YOGA-NIGHTLY-001 §7 asks for the floor table to be re-keyed, and the table is only in this branch — it is not on `main`, so the change cannot land in the yoga work (paiml/aprender#3060) or anywhere but here. Recording the measurement so it does not have to be retaken.

**The defect this PR fixes has a second instance it does not close.** The header here already says it out loud:

> THE FLOOR DID NOT MOVE WITH THE HOST (aprender#2835)

and the fix — `SILICON_FLOORS` keyed by `compute_cap`, absence yielding `UNCALIBRATED-SILICON` — is right for sm_121, which has no entry. It does not hold for the host that joined the fleet last week.

**Measured on yoga, 2026-09-09:**

```
NVIDIA GeForce RTX 4060 Laptop GPU, 8.9, 8188 MiB, 100.00 W
```

`8.9` is the key this table already has, whose single entry reads `arch: "sm_89 (RTX 4090)"`, `floor: 0.90`, derived from four measurements on lambda-vector — a **24 GB / ~450 W desktop 4090**. yoga is an **8 GB / 100 W mobile Max-Q part**. The refusal that exists for exactly this cannot fire, because **the key matches while the device does not**:

> Compute capability is a kernel-compatibility key, not a performance class.

**The change:** add a device discriminator to the lookup — `nvidia-smi --query-gpu=name` is already printed beside `compute_cap` at every call site, so it costs no new probe. Match on `(compute_cap, name)`; an uncalibrated card is then REFUSED rather than silently inheriting another machine's floor. Under the file's own rule ("Adding a silicon here requires the derivation, not just the number"), yoga gets no entry until someone measures it.

**Nothing is asserted on yoga in the meantime, and that is enforced rather than assumed** (paiml/aprender#3060):

- `scripts/perf-matrix.yaml` declares yoga `status: NA` with the reason and a decider, and its W1/W2 cells are NA — so it has no denominator cells and no clock.
- `cuda-nightly.yml`'s `ada-yoga` leg carries a VRAM ceiling and a zero-JIT assertion and **no throughput number**, with the reason in the job's own comment and in its step summary.

**Prediction, on the record, so the measurement can falsify it:** yoga's decode ratio will land **below** 0.90. Whichever way it goes it is data — refusing to take the measurement is how a 4090's floor gets inherited by silence.

Refs paiml/infra PMAT-272 (YOGA-NIGHTLY-001 §7, R-7, N-4), paiml/aprender#3060, #2835

https://claude.ai/code/session_01MSoNFCFvUg465kQBhw8kdL

Issue #2844, opened 2026-09-01
PARITY: the 10.34x gap was measured on a build with batching compiled out — re-baseline first


noahgift:
Execution plan for #2843. Spec: `docs/specifications/performance-parity-llama.cpp.md`.

**Goal:** `apr serve` reaches llama.cpp parity on the reference cell — lambda RTX 4090, Qwen2.5-Coder-7B Q4_K_M — in as few matrix runs as possible.

---

## ⚠ The 10.34x gap and its 6.07x/1.70x decomposition are WITHDRAWN

The number this plan was built on measured **a build with continuous batching compiled out.**

`evidence/parity-http/findings.json`: `apr_build: "0.64.0 (53062e7f3), --features cuda"`, dated
2026-08-24. At that commit `with_cuda_batch_tx` is behind `#[cfg(feature = "cuda-batch")]`, and
`crates/apr-cli/Cargo.toml:90` declares `cuda-batch = ["cuda"]` — **the implication runs the
wrong way**, so `--features cuda` compiled batching out. Fixed by `a18b1aced` 2026-08-25 20:21,
**27 hours later**, in a commit titled *"continuous batching was never compiled into any build a
user is told to make."*

The counter-measurement is committed: `evidence/perf-gate-001-w1-lambda/`, 2026-09-01, binary
`745fa8588`, same host and model, logs `CONTINUOUS BATCHING: max_batch=11` and aggregates
**99.9 -> 191.6 -> 353.9 -> 449.9**.

| band | ratio from the defective build | indicative ratio on main |
|---|---|---|
| c=1 | 0.534 | 0.591 |
| c=8 | 0.168 | 0.544 |
| c=16 | **0.097** | **0.401** |

**Indicative gap ~2.5x, not 10.34x** — `[U]`, the two artifacts run different workloads (128/128
single-prompt vs W1 512/128 over 1622 prompts), so this is indicative and not a ratio.

**Consequences.** Lever A's 6.07x had a build defect as its numerator. Lever B is discharged
separately: `crates/aprender-gpu/src/kernels/quantize/q4k/coalesced/` already ships ten
coalesced/DP4A GEMV variants, so §2.4's non-coalesced mechanism does not exist here — and above
m>=4 the batched path routes to cuBLAS GEMM and calls no GEMV at all. **Neither lever survives
as stated, so the ordering they implied is void.**

**The first deliverable is therefore one honest paired measurement, not an optimisation.**

## Steps

[...]

Issue #2843, opened 2026-09-01
Performance parity with llama.cpp — the single inference-performance spec (supersedes #2706)


noahgift:
Supersedes and closes **#2706** (APR-PERF-GATE-001, the receipt rule).

## Why #2706 is closed rather than continued

The epic ran 2026-08-27 → 2026-09-01. Its own status review (`docs/specifications/APR-PERF-GATE-001-status-review.md`, retained as the post-mortem) records:

- **2 of 8 matrix cells measured**, both on the final day, both CUDA, and **neither ratcheted into `perf-matrix.yaml`** — so by the epic's own instrument the matrix never moved
- **12 PRs repairing the instrument** against **3 producing a measurement**
- one defect class consumed **four sequential PRs**
- the enforcement machinery ran a full mutation sweep on **every** PR, contradicting its own §3.D trigger table; on 2026-09-01 two jobs were **cancelled at their 150-minute cap** having proved nothing
- **145 shell guards** under no complexity, coverage or TDG threshold

A v3.0 edit plan was proposed and reviewed. It did not survive: two of the "three live specs" it ruled on **do not exist**, the defect motivating its I-19 was **already fixed**, and its scope freeze rested on a c=1 decode ratio of `1.43×` that the repository's own pinned measurement puts at **0.587×**.

## What replaces it

`docs/specifications/performance-parity-llama.cpp.md` — written ground-up, and the **only** specification governing inference performance of models in this project.

It archives **13 documents, ~14,800 lines, across 4 repositories** (`aprender`, `realizar`, `qwen-coder-deploy`, `trueno`). Nothing is deleted; superseded material that still reads as current is the condition the document exists to end.

## The measurements it is built on

`scripts/llama_pin.toml`, 2026-08-25, lambda RTX 4090, Qwen2.5-Coder-7B Q4_K_M, comparator pinned `39173bcac`:

| band | agg ratio | dec ratio |
|---|---|---|
| c=1  | 0.534× | **0.587×** |
| c=8  | 0.169× | 1.352× |
| c=16 | **0.097×** | 1.554× |

**apr's aggregate is flat at ~110 tok/s from c=1 to c=16 — it does not batch, it serialises.** Per-user decode *rises* with concurrency only because each request gets the whole GPU in turn while llama.cpp shares it sixteen ways. A gate reading decode alone would call 0.097× a PASS. Hence **I-4: both metrics, every band.**

[...]

Issue #2841, opened 2026-09-01
APR-PERF-GATE-001: mini's cell declares compute_class: metal, and apr has no Metal inference path


noahgift:
## Summary

`scripts/perf-matrix.yaml` declares the `mini` host with `compute_class: metal`. **`apr` has no Metal inference path at all**, so that cell cannot be measured honestly today. Under APR-PERF-GATE-001 I-9 a cell is *spent* once run and may not be re-run to green, so a mislabeled measurement would be permanent.

This ticket records what is actually there, what enabling Metal would take, and the routes — including one this research initially missed and one it initially got wrong.

Found while attempting to start the `mini` W1 cell. **The cell was not run.**

## 1. There is no backend behind the flag

`crates/aprender-gpu/Cargo.toml:85` defines `metal = ["dep:manzana"]` (macOS-only). Four independent checks, each a command:

**Nothing enables it.** `aprender-serve` pins the dependency to CUDA:
```
crates/aprender-serve/Cargo.toml:52
  trueno-gpu = { ... package = "aprender-gpu", optional = true, features = ["cuda"] }

$ grep -rn 'aprender-gpu/metal\|"metal"' --include=Cargo.toml .    -> (empty)
```
There is no `metal` passthrough in the root facade (it has cuda, cuda-batch, wgpu, inference, training, ptx, full — not metal) and none in `apr-cli`.

**The trait it implements has no compute method.** `crates/aprender-gpu/src/backend/mod.rs`:
```rust
pub trait Backend: Send + Sync {
    fn name(&self) -> &str;
    fn is_available(&self) -> bool;
    fn device_count(&self) -> usize;
}
```
`MetalBackend` implements exactly these three, delegating to `manzana::metal::is_available()` and `MetalCompute::devices().len()`. It is a **device probe, not an inference backend**.

**The shaders exist and are never dispatched.** `backend/metal_shaders.rs` holds 13 source-text constants (ELEMENTWISE_ADD/MUL, SCALAR_MUL, DOT_PRODUCT, GEMM_NAIVE, GEMM_TILED, SOFTMAX, LAYERNORM, RELU, GELU, SILU, FUSED_ADD_RELU, COPY). The only code anywhere touching `MetalCompute` is `crates/aprender-gpu/tests/metal_backend_f101.rs`, which *enumerates devices*.

**`aprender-serve` has no Metal branch.** The only two matches for "metal" under `crates/aprender-serve/src/` are the words *"bare metal"* in comments.

[...]

PR #2845, opened 2026-09-01
spec(perf): one specification for inference performance parity with llama.cpp (#2843)

noahgift:
Closes the spec half of #2843. Supersedes #2706.

**One specification for inference performance**, written ground-up from measurements, archiving **13 documents / ~14,800 lines across 4 repositories**.

## The headline: the premise this was built on is false, and the spec says so

`evidence/parity-http/findings.json` records the parity run as `apr_build: "0.64.0 (53062e7f3), --features cuda"`. At that commit `with_cuda_batch_tx` sits behind `#[cfg(feature = "cuda-batch")]`, and `crates/apr-cli/Cargo.toml:90` declares:

```toml
cuda-batch = ["cuda"]        # the implication runs the WRONG way
```

So `--features cuda` compiled continuous batching **out**, and every request took the path `cuda_chat_backend.rs:149` labels `// Fallback: direct RwLock path (serialized)`. Fixed by `a18b1aced` at 2026-08-25 20:21 — **27 hours after the measurement** — in a commit titled *"continuous batching was never compiled into any build a user is told to make."*

The counter-evidence was committed in this tree the whole time. `evidence/perf-gate-001-w1-lambda/`, `745fa8588`, same host and model, logs `CONTINUOUS BATCHING: max_batch=11`:

| band | defective build | main | ratio then | ratio now |
|---|---|---|---|---|
| c=1 | 90.2 | 99.9 | 0.534 | 0.591 |
| c=8 | 109.6 | 353.9 | 0.168 | 0.544 |
| c=16 | 108.4 | 449.9 | **0.097** | **0.401** |

**Indicative gap ~2.5×, not 10.34×.** `[U]` — the artifacts run different workloads (128/128 single-prompt vs W1 512/128 over 1622 prompts), so it is indicative, not a ratio.

§2.1 is **WITHDRAWN in place**, with the evidence, rather than quietly deleted. This is exactly the failure the document exists to prevent — a number that never proved the mechanism engaged.

## What review changed

Five rounds — quorum, `agy /teamwork`, `pmat`, the NVIDIA CUDA docs MCP, `agy /grill-me`, `agy /plan`, and a Claude Code Plan agent. **Roughly a dozen of the document's own rules were changed by review rather than by me.** The three worth naming:

**The gate outlawed its own fix — twice.** Gating `dec_ratio ≥ 1.0` on every band demands apr *dominate* two metrics that trade against each other; the batching PR would necessarily lower per-user decode and be rejected. I fixed that with a non-regression floor — and the floor rejects it too:

```
apr decode c=16 today (serialising) : 110.6 tok/s
llama.cpp decode c=16 (batched)     :  71.2 tok/s
after batching:  ~71.2  =  a 35.6% DROP
```

The same defect in a new spelling, which is the shape #2707 is named for. Decode at c>1 is now REPORTING until aggregate parity, with the floor set from the **post-batching** baseline.

**The gate had no producer.** `test_llm_band.rs:231` returns `Unmeasured` unconditionally, guarded by a test named *"The producer must never be able to emit a comparator ratio"*, and `decode_tok_per_sec` is absent from every band of every receipt. §6.2 says so.

**Summing per-phase attribution was refused with a measurement**: over six receipts, Σ per-request samples ÷ band span is *exactly* the concurrency — 1.000, 4.000, 8.000, 15.530. That refutes summing overlapping spans, not averaging them, so averages and utilization are adopted instead.

## What is carried forward

The **invariant set** — the asset from #2706 — as I-1…I-14, plus four new ones each from a measured defect: **I-15** no boolean accelerator flag, **I-16** `compute_class` must be reachable by some build, **I-17** a parity claim names its band, **I-18** the measuring binary is built from an ancestor of HEAD.

**I-15 and I-16 are live defects today** (§6.1): the parity harness itself runs `apr serve run {model} --gpu ...`, and `perf-matrix.yaml` declares `mini` as `metal` with no Metal path (#2841).

## What it declines to do

No latency threshold (TTFT/ITL unmeasured, and *recorded* as unmeasured rather than defaulted) · no summing attribution identity · no iteration budget (a declared budget with a retrospective five-whys cannot decline work) · one gated cell, because breadth before depth is what produced eight empty cells.

## Archived, not deleted

aprender 2 (with 42 `roadmap.yaml` references repointed) · realizar 9 · qwen-coder-deploy 3 · trueno 1. Sibling repos are committed separately. `APR-PERF-GATE-001-status-review.md` is **retained** in `docs/specifications/` — it is the post-mortem justifying the supersession.

Execution plan: **#2844**.

Refs #2843, #2844, #2841, #2825, supersedes #2706


noahgift:
Superseded by #2851, which carries every commit of this branch as an ancestor plus the v3 instrument (receipt, JOIN, effective-config, conformance guard). Closing so the merge queue holds one PR for this work.

PR #2847, opened 2026-09-02
perf(prrev): cache the positive controls by the guard's own bytes — the sweep fits its budget again

noahgift:
## The sweep did not fit its budget

`scripts/pr_review_sweep_needed.sh` runs the guard's full positive-control set on every PR that
touches the review machinery. Two consecutive sweeps were killed by the job timeout at ~93
minutes. The controls are the expensive part: each one seeds a tree, runs the guard against it,
and asserts the guard turns RED — and the same control was re-run from scratch on every
invocation even when neither the guard nor the seed had changed by a byte.

## The fix

Cache each positive control's transcript, keyed on **the guard's own bytes** — `sha256` of
`check_pr_review_receipt.sh` plus the seed inputs — so a control is re-run exactly when
something it depends on changes, and is a cache hit otherwise.

Content-keyed, not time-keyed: an edit to the guard invalidates every control in the same
instant, which is the property that matters. A stale-by-mtime cache would have been the
"shadowed artifact" failure — edits look effective and change nothing.

**Sweep wall-clock: ~93 min (killed) → ~35 min (passes).**

## Why this is not a weakened check

The controls still run. `tests/pr-review.bats` gains rows asserting:
- a changed guard busts the cache (the control re-runs, and still turns RED)
- a changed seed busts the cache
- a cache hit reproduces the *same* transcript the live run produced

That last row is the one that matters: a cache that returned a different answer than the run it
replaces would be a guard that passes for the wrong reason. It is asserted, not assumed.

`ci.yml` gets the cache directory and nothing else.

Refs PRREV-023


noahgift:
PP-066 triage (epic #2873, ticket #2986): **MERGE-0.66:C0-1 → STOP:conflict** — one rebase attempt onto `origin/main` @ cdc0acb99 (`git rebase origin/main` in a detached worktree) stopped on: `.github/workflows/ci.yml tests/pr-review.bats`. Aborted (`git rebase --abort`), branch untouched, labelled `pp-066/merge,pp-066/conflict`. The pr-review sweep it caches was restructured by #2985 (cdc0acb99, base-owned pr-review-quorum.yml); the ci.yml hunk and the bats table need re-basing against that shape. Workflow-touching. Producer is never the gate: nothing armed.

noahgift:
PP-066 triage (epic #2873, ticket #2986) — **PRIOR-ART CLOSE** (operator decision, STEP 2b). The branch is retained as `prior-art/2847` and is never deleted; the intent is carried on the issue named below. Zero semantic merges were made.

`prior-art: prior-art/2847@07a1a4a71 · intent → #2993 (carry-0.67; sweep restructured by #2985) · salvage: tests/pr-review.bats · conflicts: .github/workflows/ci.yml tests/pr-review.bats`

PR #2848, opened 2026-09-02
fix(cgp): a binary on PATH is not a device on the bus — Coverage Nightly was red on an environment fact

noahgift:
## Coverage Nightly is red on an environment fact

```
thread 'profilers::system::tests::test_system_health_with_gpu' panicked at
  crates/aprender-cgp/src/profilers/system.rs:246:9:
  nvidia-smi exists but no health data
thread 'profilers::system::tests::test_vram_with_gpu' panicked at
  crates/aprender-cgp/src/profilers/system.rs:262:9:
  nvidia-smi exists but no VRAM data
test result: FAILED. 102 passed; 2 failed
```

Both tests gated on `which::which("nvidia-smi").is_ok()` and then asserted the collectors return
data. `intel-clean-room-16` has the NVIDIA userland installed and **no visible device**, so
nvidia-smi resolved, the tests concluded a GPU was present, `collect_system_health()` correctly
returned `None`, and the suite reported a **code defect for an environment fact**.

The collectors were right. The precondition was asking the wrong question: a binary on `PATH` is
not a device on the bus.

## The fix is a probe, not a skip

`gpus_reported()` runs `nvidia-smi --query-gpu=name --format=csv,noheader` and counts non-empty
lines — `0` when absent, `0` when present but the driver reports nothing, `N` when N devices
answer. Not-installed and installed-but-blind are the same fact here, and only one of them was
previously recognised.

## Both polarities, so neither half is vacuous

`#[ignore]` is banned in this repo and a one-sided skip is how a test quietly stops testing. The
single test becomes a pair, and exactly one is meaningful on any given box:

| test | precondition | runs on | asserts |
|---|---|---|---|
| `..._have_valid_data_when_a_gpu_is_reported` | ≥1 GPU reported | lambda, gx10, yoga | health + VRAM present and sane |
| `..._are_none_without_a_reporting_gpu` | 0 GPUs reported | **the clean room** | `None` — never a fabricated zero, never a panic |

The second is a real assertion where the failure was, not an early return. Each half skips
loudly (`eprintln!` naming which host condition sent it to the other), so a box that *should*
have a GPU and does not report one is visible rather than silently green.

## Mutation-verified, both halves

On a host with an RTX 4090:

| mutation | positive half | negative half |
|---|---|---|
| `collect_system_health()` forced to `None` | **FAILED** | passed |
| `gpus_reported()` forced to `0`, negative assertion inverted to `is_some()` | passed | **FAILED** |

10/10 `profilers::system` tests pass. `clippy -p aprender-cgp --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/aprender-cgp/src/profilers/system.rs
+++ b/crates/aprender-cgp/src/profilers/system.rs
@@ -236,74 +236,140 @@ mod tests {
         let _ = read_cpu_temperature();
     }
 
-    /// How many GPUs `nvidia-smi` actually REPORTS -- not whether the binary exists.
+    /// How many GPUs `nvidia-smi` actually REPORTS -- not whether the binary exists,
+    /// and never at the cost of hanging the suite.
     ///
-    /// `which::which("nvidia-smi").is_ok()` was the precondition for both GPU tests below,
-    /// and it is the wrong question. The intel clean-room runner has the NVIDIA userland
-    /// installed and no visible device, so `nvidia-smi` resolves, the tests decided a GPU
-    /// was present, the collectors correctly returned `None`, and Coverage Nightly failed
-    /// with "nvidia-smi exists but no health data" -- a red build reporting an ENVIRONMENT
-    /// fact as a code defect. A binary on `PATH` is not a device on the bus.
+    /// `which::which("nvidia-smi").is_ok()` was the precondition for both GPU tests, and it
+    /// is the wrong question. The intel clean-room runner has the NVIDIA userland installed
+    /// and no visible device, so `nvidia-smi` resolved, the tests decided a GPU was present,
+    /// the collectors correctly returned `None`, and Coverage Nightly failed with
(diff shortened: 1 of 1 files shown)


noahgift:
Pulled into **0.67.0** (from 0.68.0) by the train orchestrator, 2026-09-11 09:05Z: the nightly coverage run has been RED six days running and this PR fixes half of it — `profilers::system::tests::test_system_health_with_gpu` / `test_vram_with_gpu` failed on intel-clean-room-9 on 2026-09-10 (run 34451014241). The other half is the same defect class in aprender-compute (`backends::gpu::device::backward::tests::test_silu_backward_*`, `GpuDevice::new().expect("GPU device")`, run 34575134766 on yoga-build2) and is being fixed under PMAT-1105. Needs `gh pr update-branch` once #3089 lands (it carries the shallow-clone fixes every fresh runner needs).

PR #2849, opened 2026-09-02
fix(aprender-gpu): drop the manzana dependency and its dead Metal backend

noahgift:
Removes `manzana` from `aprender-gpu`, retiring [RUSTSEC-2026-0273](https://rustsec.org/advisories/RUSTSEC-2026-0273) from this workspace by deletion rather than suppression.

## Why remove rather than bump

Both published versions in the lockfile's range (0.1.0, 0.2.0) are **yanked**, and the advisory still carries `[versions] patched = []` — so no bump clears the gate. manzana 0.3.1 exists and is clean, but until [rustsec/advisory-db#3193](https://github.com/rustsec/advisory-db/pull/3193) merges, `cargo audit` reports *"No fixed upgrade is available!"* against it.

More to the point, upgrading buys nothing here. The two manzana calls behind the `metal` feature only ever **enumerated** devices. The operations that would make a Metal backend useful were fabricated upstream — `dispatch()` returned `Ok(())` having dispatched nothing, `compile_shader()` returned a handle built from a hash of the source string. manzana 0.3.x replaces both with `Err(Unimplemented)`. There is no published version on which this backend can do real work.

## The audit ignore named this exact condition

`.cargo/audit.toml` already specified when its own entry should go:

> REMOVE WHEN either the advisory gains a patched range and this workspace is on a version inside it, **or the manzana dependency is dropped from aprender-gpu entirely**.

So the 35-line ignore goes too. `cargo audit` now exits 0 with RUSTSEC-2026-0273 not suppressed but **absent** — manzana is gone from `Cargo.lock`, which is the only place it ever appeared unconditionally (cargo locks target- and feature-gated deps regardless of activation, which is why the ignore was needed at all).

## Nothing is lost

The `metal` feature was never enabled anywhere — not in `default`, not in another manifest, not in CI — so every line behind it was dead. `MetalBackend` now reports unavailable unconditionally, matching `VulkanBackend`, which has always been an honest placeholder. On macOS the `wgpu` feature already reaches Apple GPUs through wgpu's Metal backend and does execute.

## Deleted: a test suite that never asserted anything

`tests/metal_backend_f101.rs`, 359 lines, 10 `#[test]` fns. Nine opened with `if !metal_available() { return }`; the tenth asserted only `!available` off-macOS. `metal_available()` was hardwired to `false` whenever the feature was off — which was always. The suite reported 10 passing tests on every CI run and exercised nothing.

`metal_shaders` moves from behind the removed feature to unconditional compilation. It is MSL source strings with no manzana dependency, and **its 4 tests now actually run** — previously gated on `macos + metal`, so never built. Its doc example, which called the now-absent `MetalCompute::default_device()`, now says plainly that this crate has no dispatcher.

Net: **-472 lines, 10 tests that could not fail removed, 4 real tests gained.**

## Verification

```
cargo check -p aprender-gpu                clean
cargo test -p aprender-gpu --lib           444 passed / 0 failed / 0 ignored
cargo test -p aprender-contracts --lib     1474 passed / 0 failed
cargo audit                                exit 0; RUSTSEC-2026-0273 absent
cargo deny check advisories                ok
rustfmt (changed files)                    clean
```

The `useless use of vec!` clippy warnings are in `tests/rocm_backend_f111.rs` — untouched here and pre-existing.

**Not verified, stated rather than implied:** the macOS build. This toolchain has no std for `aarch64-apple-darwin`. The change removes platform-gated code rather than adding any, and no `cfg(target_os)` arm remains in the touched files.

## Note on trueno

`trueno-gpu/src/backend/mod.rs` in `paiml/trueno` is byte-identical to the file changed here and carries the same dependency. It needs no PR: that repo is **archived and read-only** (*"MOVED → paiml/aprender (crates/aprender-compute)"*), and nothing live path-depends on it.

https://claude.ai/code/session_01RnES3PSMkcR9tc2iP3sZQg


noahgift:
Tracking the release side in #2850: merging this is necessary but not sufficient. The broken artifact is the published `aprender-gpu 0.64.0`, which declares `manzana = "^0.2.0"` — a range whose only member is yanked — so `--features metal` fails to resolve on crates.io until the next release ships.

noahgift:
## Cross-vendor review complete — verdict `DEGRADED`, three advisory findings

Receipt at `evidence/pr-review/2849/9aa58d367…/`. Reviewer: **agy pinned to `gemini-3.1-pro-high`** — verified cross-vendor by the guard's own predicate (`--match-arm-e-same-family` returns 1 for it, 0 for the two Claude ids agy also offers, since agy is a harness and not a model). It returned `agreed_with_author: false`.

### Why DEGRADED rather than FINDINGS

Mechanical, not a judgement: `cargo mutants` tested **0 mutants** because `cargo test` fails in the *unmutated* tree — the doctest at `crates/aprender-gpu/src/lib.rs:12` imports `trueno_gpu::ptx`, a pre-rename path. §3.D says `attempted: 0` under `status: consulted` is a vacuous pass, so the arm is `unreachable`; §6 makes any unreachable consultation `DEGRADED`. **Reproduced identically on the merge-base `f21f437c2`** — pre-existing, not this PR's doing, and recorded as `measured` in the SARIF rather than narrated around.

### The three findings

| # | severity | finding | disposition |
|---|---|---|---|
| 1 | high | Removing the `metal` feature + `pub use manzana::metal::*` breaks semver for consumers of published 0.64.0 | **Already tracked** — #2850 says 0.65.0 is the honest number. Maps to §7 class B5, which no consultation emits today. |
| 2 | medium | Deleting the test file drops PMAT-006 / METAL-01..05 / FKR-011 tracking silently | **Fixed** — disposition recorded at #2841 (comment above). |
| 3 | low | `metal_shaders` now unconditional embeds MSL strings in non-Apple binaries | **Accepted, deliberate** — gating on `target_os = "macos"` would mean the 4 tests never run, since aprender CI has no macOS runner. Recorded rather than silently dismissed. |

Two of its factual claims were re-verified against the tree before being recorded, and one was worse than this PR originally claimed: agy quotes `let result = expected.clone();` — present at lines 60, 213 and 247. The tests that got past the skip guard asserted a value equals its own clone.

### Disclosure

`attestation_level: L1-self`, and the receipt carries an explicit `orchestration_disclosure`: the session that ran this skill also authored the diff, so it held the author's reasoning trace. The independence here is **agy's, not the orchestrator's**. Read the receipt with that bound in mind.

Not signed locally — `pr-review-sign` holds the secret. It validated `ACCEPT` under a throwaway keypair first, in a run where the guard's four positive controls all fired.

PR #2859, opened 2026-09-03
fix(release): publishing goes only through a preflight gate; the cascade no longer overrides a dirty tree (F-9, PMAT-745)

noahgift:
Ticket PMAT-745, the operator's release rule F-9: "publish only through a workflow gate; if the path is still `cargo publish --allow-dirty` with no gate, fix that as a ticket before publishing. Never --allow-dirty."

## The defect

`scripts/cascade-publish.sh:274` ran `cargo publish … --allow-dirty --locked` with no precondition of its own. A dirty tree, an untagged commit, a commit off main, or a release whose dogfood verdict was NO-GO could all reach an immutable registry.

## The change

`scripts/check_publish_preflight.sh`, five rules, each printing its own line, the verdict their AND: **R1** clean tree (tracked or untracked), **R2** version from `cargo metadata` (root manifest), **R3** tag `v<version>` at HEAD, **R4** HEAD an ancestor of `origin/main`, **R5** the newest `scripts/dogfood.sh` receipt says GO for this commit and version. Exit 2 when the box cannot answer (not a pass). The cascade calls it before its first upload in every uploading mode and refuses on non-zero; the dirty-tree override is gone. `/.dogfood/` is ignored so a dogfood run does not dirty the tree it certifies.

## Verification

- `--selftest`: 11/11 rows in a throwaway repository, both polarities of R1, R3, R4, R5.
- **Must-fire mutations, measured:** R1 dropped → `tracked_change_refuses` and `untracked_file_refuses` BROKE (9/11); GO check dropped → `dogfood_no_go_refuses` BROKE (10/11). Both restored.
- On this tree the gate REFUSES (R1 dirty, R3 no tag, R5 no receipt), as it must.
- ci.yml `guard-runner-labels` runs the case table (`check_guards_are_wired.sh` PASS); `FALSIFY-PUB-CLI-005` (`pv validate`: valid; its `test:` runs green); `check_pass_grep_anchored.sh`, ratchets, claim literals, roadmap citation PASS; bashrs 0 errors on both scripts.
- `cargo mutants --in-diff`: not applicable (no Rust in the diff); the guard's own case table plus the two measured mutations stand in.

## Owed separately

A GitHub publish workflow: no `CARGO_REGISTRY_TOKEN` secret exists for the fleet (only `PR_REVIEW_SIGNING_KEY_B64`), so the gate lives in the script the cascade and the drain already run. **[A]** recorded in `docs/audits/impl-PMAT-742-receipt.md`.

## Amended under review (seven commits; final head is the one the receipt names)

**The structural finding, mine.** R5 as first written required the newest dogfood receipt to say GO — and `scripts/dogfood.sh` is NO-GO before its own cascade *by construction*: `publish-dry-run` of a workspace root cannot resolve members that are not on the registry yet, and the declared `check_multiplatform_dogfood` gate reads receipts from `cargo install aprender --version X` on four hosts, which no host can run before X is published (0.64.0 shipped on a written NO-GO determination, `evidence/dogfood/0.64.0/VERDICT.md`). The same shape PMAT-744 had one gate over. Answer, recorded rather than bypassed: `dogfood.sh --phase pre-publish` marks those two rows `DEFER` with the obligation that discharges them, asks the crates.io **sparse index** directly whether the version is absent, and writes `phase` and `deferred[]` into the receipt; `DEFER` outside that phase is recorded as the FAIL it is. R5 accepts a pre-publish receipt only when every deferred row is in its whitelist (`publish-dry-run`, `declared:check_multiplatform_dogfood`). Measured on this tree: the pre-publish run engages both DEFER rows, `dogfood-gates` and `version-unpublished` green; the verdict is still NO-GO on six standing rows of main (bashrs 183 SEC/DET/IDEM errors, pv-contracts, 28 strict SATD, pmat-comply abort, transport-decl, and cli-surface — the last fixed here: the parser read clap's wrapped description lines as subcommands; 110/110 real ones answer `--help`).

**Six single-model cross-vendor reviews (agy, gemini-3.1-pro-high) and one three-lane quorum, every finding re-run by me before its disposition:**
- run 1: the cascade's config backup sat untracked inside the root crate after R1 had passed → mktemp outside the tree (measured both polarities); "no mutation tests" → contradicted, the ci.yml case table catches both (re-measured 9/11, 10/11).
- run 2: mktemp failure would overwrite the config with no backup → refusal, exit 2; the FALSIFY-PUB-CLI-005 grep tripped on my own comment → reworded; R3 `grep -qx` accepted tag `v1-2-3` for `1.2.3` → `-F` + row `tag_lookalike_refuses`; R2 had no negative row → `no_root_package_refuses`; Python traceback on empty `cargo metadata` → exit 1.
- run 3: `curl -f` turned a 404 (a never-published crate) into a FAIL → the sparse index, status read apart from the body; contract prediction said 11 rows → 16.
- run 4: HTTP 200 with a non-index body was read as "absent" → PASS (fail-open) → three exit codes, only "parsed and absent" passes (measured on an HTML page, an empty body, and the live index for 0.64.0/0.65.0).
- run 5: index path not lowercased → fixed; temp-dir whitelist "leaks under a custom TMPDIR" → accepted, not fixed (the SEC011 guard shape; a leaked scratch dir is the side to fail on).
- run 6: restore trap armed one line after the overwrite → armed before.
- quorum (2–1 do-not-implement-as-written): `cargo publish … | tail -6` read the verdict from text, never from cargo's status (two lanes independently) → status kept, "exit 0 without `Published`" deferred and named; multiplatform version by grep is empty for a member under `version.workspace = true` (measured) → cargo metadata, grep fallback, empty refuses; unquoted deferred-list split (fails closed, still) → `IFS=, read -ra`. Refuted: the "broken sed" index path (substring expansion, HTTP 200 measured), the library-facade `dogfood-use` claim (this crate builds `apr`), the ci.yml `set -e` claim (outside this diff; a pipeline's status is `head`'s).

**Verification on the final head:** `--selftest` 16/16. Must-fire mutations measured RED and restored: R1 dropped → `tracked_change_refuses` + `untracked_file_refuses` BROKE; GO conjunct dropped → `dogfood_no_go_refuses` BROKE; `PREPUBLISH_DEFERRABLE` shrunk → `prepublish_allowed_deferrals_pass` BROKE; phase check dropped → `deferral_outside_prepublish_refuses` BROKE; `-F` removed → `tag_lookalike_refuses` BROKE. `check_multiplatform_dogfood.sh` in both phases. `check_guards_are_wired`, `check_shell_lint_ratchet`, `check_pass_grep_anchored`, `check_verifier_pinning`, `check_dogfood_shim`, `check_no_timing_in_required`, `check_roadmap_completion_is_cited`: PASS. `bashrs lint`: 0 errors on all four scripts (`dogfood.sh` had 3).

**Not measured here:** the `DEFER`-outside-phase rule inside `dogfood.sh` has no unit fixture; it is exercised by the pre-publish run (recorded in the PMAT-742 impl receipt).


noahgift:
Re-queuing: the previous run was killed by the runner outage of 2026-09-03 (workspace-test and pr-review-receipt ended with zero failed steps and 'runner lost communication'; gate failed on their result). Close/reopen for a fresh run; head 389d707b5 unchanged. (PMAT-929)

PR #2862, opened 2026-09-04
fix(compute): shared_instance initializer no longer takes DEVICE_INIT_LOCK — ABBA deadlock with GpuDevice::new (PMAT-952)

noahgift:
## wgpu init deadlock: `shared_instance` initializer vs `GpuDevice::new` (PMAT-952)

**Found by the 0.65.0 release clean-room on `587ad0797`** (2026-09-04): `cargo test --workspace --lib` sat 20+ minutes in the aprender-compute lib tests with 49 threads in `futex_wait`. Two clean-rooms on the same code were green the same day; only the first initialization of a process can race.

**Mechanism (gdb, all threads):** `GpuDevice::new` holds `DEVICE_INIT_LOCK` and calls `shared_instance()`; the `shared_instance` `OnceLock` initializer took `DEVICE_INIT_LOCK` too. A thread entering `shared_instance()` through `is_available_async()` (no mutex) became the initializer, blocked on the mutex held by a `GpuDevice::new` thread, which was parked on the `OnceLock`. ABBA. Reachable in production whenever `is_available()` and `GpuDevice::new()` race on the wgpu path.

**Fix:** the initializer takes no mutex. The `OnceLock` already serializes the one-time ICD enumeration (PMAT-778's intent is kept; `DEVICE_INIT_LOCK` still serializes adapter/device creation).

| evidence | result |
|---|---|
| `cargo test -p aprender-compute --lib --features gpu pmat952` on the fix | 2 passed |
| must-fire mutation: mutex restored in the initializer | `test_pmat952_shared_instance_does_not_deadlock_against_device_init_lock` **FAILED**: probe child exit 101 after the 30 s timeout, `shared_instance() parked behind DEVICE_INIT_LOCK` |
| fix again | 2 passed |

The test runs its probe in a fresh child process (`current_exe --exact …`) because the `OnceLock` can only race while empty; an in-process test would pass vacuously after any other GPU test.

Hook precondition (`pmat analyze complexity --file … --max-cyclomatic 30 --max-cognitive 25`): no function over threshold. Roadmap block PMAT-952 in the same commit.


noahgift:

PR #2860, opened 2026-09-03
release gates 0.65.0: the six standing dogfood red rows and the clean-room doc red, fixed on main (PMAT-746..749)

noahgift:
Draft until the per-branch review quorums, the SATD pass, and the PR-REVIEW receipt land on it.

Batch of 19 fix branches off 68b059ca9 (merge commits, never rebased), each fixed by a worker and independently verified; every acceptance command re-run on the merged tree by the batch gate script (results in the PMAT-742 impl receipt):

- **bashrs (PMAT-747):** 184 SEC/DET/IDEM findings in 59 scripts fixed at their root (ten buckets + one leftover bucket). On the merged tree the dogfood row's own classifier reports **0 gating findings outside scripts/dogfood.sh**, whose last two are fixed on #2859 (the branch that owns that file).
- **Complexity (PMAT-746):** `serve/handlers.rs`, `eval/inference.rs`, `tokenize.rs` refactored under the pre-commit thresholds (cognitive ≤ 25, cyclomatic ≤ 30) with the crate's tests green; a CI-side ratchet `scripts/check_complexity_ratchet.sh` (704 baselined offenders, shrink-only, case table, wired into guard-runner-labels) so this debt can no longer land through the queue.
- **Doctests (PMAT-748):** the four apr-cli doc blocks fenced as text; `cargo test -p apr-cli --doc` green.
- **Contracts (PMAT-749):** `pv validate` accepts every file under contracts/ (kaizen records as a Kaizen kind with a validator measured against all 46 records, pv's binding manifests as their own kind, publish-manifests recognised as the artifact they are, one YAML syntax fix); `pv lint contracts --strict-test-binding` PASS; aprender-contracts tests green.
- **Transports (PMAT-749):** root `[package.metadata.transports]` with three e2e tests that spawn `CARGO_BIN_EXE_apr` (cli, mcp over stdio, http over a bound socket with a tracked fixture); all three green three times.
- **pmat-comply:** the comment dividers that tripped pmat 3.34.0's char-boundary panic are ASCII (comments only; upstream fix in paiml-mcp-agent-toolkit#1166 → one maintainer PR → 3.36.0). With the panic gone the row runs and hinges on **CB-200 (TDG grade gate): 609 functions below the declared `min_grade = "B"`** — ticketed as PMAT-768, not lowered.
- Ratchets re-seeded on the merged tree (complexity 8 rows out; shell-lint 876 → 13); three uncited speed literals the refactors had moved were removed rather than re-keyed.

[...]

PR #2863, opened 2026-09-04
fix(orchestrate): stdio MCP transport keeps a clean-exit server's response over a broken-pipe write (PMAT-953)

noahgift:
## stdio MCP transport: a clean-exit server's response survives a broken-pipe write (PMAT-953)

**Found by the 0.65.0 release clean-room, run 2 on `587ad0797`:** 88,537 passed, 4 failed, all in `mcp_client_tests.rs` (`test_discover_tools_no_result`, `_empty_response`, `_skips_empty_names`, `test_stdio_transport_result_no_content`), each with `write stdin: Broken pipe`. Their servers are `sh -c "echo '<json>'"`: they answer and exit 0 before the parent writes the request. Under load the child wins that race more often; the same tests were green in two clean-rooms earlier the same day.

**Mechanism:** PMAT-950 (#2860) held the write error until `wait_with_output`, then returned it whenever the exit was clean, discarding the response already on stdout. A server that ignores its stdin has still answered if stdout carries a response.

**Fix:** the write error is returned only when the child exited cleanly **and** wrote nothing.

| evidence | result |
|---|---|
| `cargo test -p aprender-orchestrate --lib -- test_stdio_transport test_discover_tools` on the fix | 20 passed |
| must-fire mutation: the stdout-empty guard reverted | `test_stdio_transport_clean_exit_without_reading_keeps_its_response` **FAILED**: `got: Err("write stdin: Broken pipe (os error 32)")` |
| fix again | 1 passed |

The test makes the race deterministic (child closes stdin, echoes a response, exits 0; 1 MiB request). Roadmap block PMAT-953 in the same commit. 0.65.0 was tagged at 587ad0797 with this defect by operator decision; this PR is the cherry-pick behind the tag.


noahgift:

PR #2864, opened 2026-09-04
fix(publish): sibling dev-dependencies path-only, preflight R6, drain keeps DEFER lines; 0.65.1 (PMAT-955, PMAT-954)

noahgift:
## 0.65.0 cascade stopped at 48/74: a version-level dev-dependency cycle (PMAT-955); the drain hid the reason (PMAT-954); 0.65.1

`cascade-drain.sh` reported `STUCK: no progress (48 -> 48)/74` with no reason: its log filter keeps only `PUBLISHED|FATAL|already|TIER|RETRY`, so every `DEFER` line (the only line that says why) was dropped. Running `cascade-publish.sh --only-tier 2` by hand:

```
→ aprender-core (0.64.0 → 0.65.0): DEFER (failed to select a version for the requirement `aprender-data = "^0.65.0"`)
```

`aprender-data`'s dev-dependency on `aprender-test-lib` went through the workspace alias, which carries `version = "0.65.0"`; cargo keeps a versioned dev-dependency in the published manifest and resolves it on the registry. `aprender-test-lib` needs `aprender-core`. No publish order breaks that cycle. Four more siblings had the same shape (`aprender-distribute`, `aprender-orchestrate`, `aprender-train` ×2). The memory rule (sibling dev-deps path-only, since the 0.60.0 cascade) had no guard.

| change | evidence |
|---|---|
| five sibling dev-dependencies path-only | `cargo metadata`: versioned sibling dev-deps `[]`; `cargo check -p … --tests` on the four crates: Finished |
| `check_publish_preflight.sh` **R6**: no sibling dev-dependency may carry a version; workspace fixture (root + members a, b) on both polarities | selftest **18/18**; must-fire mutation (R6 never sets `fails`) → `BROKE versioned_sibling_devdep_refuses`, 17/18; restored 18/18 |
| `cascade-drain.sh` keeps `DEFER` lines | filter now `PUBLISHED\|FATAL\|DEFER\|…`, tail 60 |
| workspace 0.65.0 → 0.65.1 | `bump-version.sh 0.65.1`; `--check` ok; facades' upstream pin 0.65.1 |
| lockfile registry-sibling ratchet | 7 → 1 collision (`--update`, shrink) |
| claim-literal baseline | two stale entries pruned (shrink). **No CHANGELOG entry**: the baseline is line-keyed and the set-aperture rule refuses the coordinate shift a changelog insert causes unless the guard changed (PMAT-956 filed); the 0.65.1 note goes in the GitHub release and the release receipt |

State on crates.io: 48 crates at 0.65.0, the root and 25 others at 0.64.0 with their 0.64 requirements intact (consistent). `v0.65.0` stays where it was cut. After this lands: perf-gate, clean-room, dogfood pre-publish, tag `v0.65.1`, cascade through R1–R6, post-publish dogfood.

Guards on this tree: `check_workspace_siblings_pathed.sh` PASS, `check_baseline_ratchets.sh` PASS, `check_roadmap_completion_is_cited.sh` PASS, `check_readme_claims.sh` 0 FAIL, bashrs 0 gating on the two scripts, `cargo fmt --check` ok, `cargo test -p aprender-contracts --lib` 1501 passed. Roadmap blocks PMAT-954, PMAT-955, PMAT-956.


noahgift:

PR #3025, opened 2026-09-06
docs(PP-066): session 2026-09-06 — SPEC-2.0 (v2.0, 18-row scope, claims ratchet, executable criteria), DAG rows/edges/amendments, decision receipts, roadmap mints, rescope quorum record

noahgift:
PP-066 orchestrator docs PR — session of 2026-09-06 (driver v3 → v5.1). Branch `agent/pp-066-spec` (orchestrator: the only branch that writes the DAG, the spec block, the roadmap and the README counts under G-11a's rule). Epic #2873.

**What lands (14 commits, every one verified by the guards it touches):**
- **SPEC-2.0** (#3023): spec v2.0 §4.1 — three claims, the 18-row 0.66 scope, 39 rows cut to 0.67 each naming the claim it protected (generated from the DAG's `cut_by`/`claim_protected`), the claims ratchet, the executable criteria table; `scripts/release_criteria.sh` (nine credited criteria C0 C4 C6 C7 C8 C9 C11 C13 C14, one exit-coded command each, C0 first through the analyser pin, never vacuous — 6-row case table; C1 C2 C3 C5 C10 C12 → 0.67); `scripts/run_clean_room.sh` (C8 via `../infra`, ENV exit 2 otherwise); the rescope quorum record `docs/audits/pp-066-rescope-quorum.md` (scope-holds-with-changes; nine claim removals assigned to R-7; C5 → 0.67 by unanimous Q3).
- **DAG** (102 rows, `check_dag_invariants.sh` exit 0, `render_dag.py --check` byte-identical): rows L0-1a/L0-1b (#2971), G-11 (#3012), G-11b (#3018), G-10b (#3013), G-10c (#3014), U-1 (#3015), S-0 (#3016), R-8 (#3019), SPEC-2.0 (#3023); R-0a/R-0b split folded from #3003; the v5 edges; amendments for every slack move; G-10 and the seven decision rows complete.
- **Receipts**: G-10 (PMAT-1059, #3011 merged) complete; seven decision receipts (D-2/3/5/8/9/10/11) each citing its #2873 comment.
- **Roadmap** (G-6 additive guard PASS after `roadmap_trim.py`): PMAT-1060..1069 minted by hand (pmat#1169), PMAT-1059 and the decision tickets completed with proof.
- `docs/audits/pp-066-status-2026-09-06.md` (from `scripts/pp066_state.sh`), `docs/audits/driver-kaizen.md` (15 lines), `docs/audits/impl-estimates.jsonl` (+1).

[...]

PR #3027, opened 2026-09-06
roadmap: dedupe twelve duplicated ids; PMAT-1060 README CPU-only claim vs wgpu default, for 0.65.3 (Refs PMAT-1060)

noahgift:
H4 of the BSE-001 convergence brief (repair once, commit): twelve epic-subtask stubs collided with full top-level items; the full item was kept and the stub dropped each time; `pmat work validate` now reports zero duplicates. PMAT-1060 files the 0.65.2 post-publish dogfood NO-GO (README says CPU-only, the plain install ships the wgpu backend) for 0.65.3; per spec v1.3 §3 the freeze is satisfied by the dogfood run and recorded.

Pmat-Ticket: PMAT-1060

### DIFF
--- a/docs/roadmaps/roadmap.yaml
+++ b/docs/roadmaps/roadmap.yaml
@@ -16422,3 +16422,28 @@ roadmap:
   - kind:code
   - pp-066
   notes: null
+- id: PMAT-1060
+  github_issue: null
+  item_type: bug
+  title: 'README claims cargo install aprender is CPU-only, but the plain install compiles the wgpu/Vulkan
+    backend by default (0.65.2 post-publish dogfood NO-GO) — fix for 0.65.3'
+  status: planned
+  priority: high
+  assigned_to: null
+  created: 2026-09-06T16:59:05Z
+  updated: 2026-09-06T16:59:05Z
+  spec: null
+  acceptance_criteria:
+  - 'Either README.md''s CPU-only claim is made true (crates/aprender-serve default features no longer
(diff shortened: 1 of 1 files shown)


noahgift:
FYI from the PP-066 driver session (instance A): #3030 carries the identical byte-for-byte dedup of these twelve ids (opened at 18:10Z without re-listing PRs — my I0 miss; both deletions merge cleanly in either order). Three things worth knowing here:

1. This PR's `guard-runner-labels` fails at G-6 ("a PR's roadmap diff only adds entries or edits lifecycle fields") because `subtasks` was not a lifecycle key. #3030 adds it (`scripts/lib/roadmap_diff.py` LIFECYCLE_KEYS + self-test row 3b), so once #3030 merges, a `gh pr update-branch 3027` should turn this green; until then the guard refuses it as written.
2. Root cause of the main red this repairs, measured: the fleet's PATH pmat moved 3.37.0 → 3.39.0 at 17:26–17:33Z (intel + lambda) and sovereign-ci's `roadmap-valid` runs bare PATH `pmat`; main @ 3792afa3d push run 34049248452 is RED on exactly these twelve ids. Tickets #3028 (aprender) and paiml/infra#468 (pin declaration vs hosts). #3030 also adds an in-repo, pin-independent duplicate-id guard so the runner's pmat version is never the only thing standing between main and a duplicate.
3. `PMAT-1060` minted here collides with `PMAT-1060` hand-minted on #3025 (R-0b). #3025 will renumber its own; nothing to do on this side.

Issue #493, opened 2026-03-14
profile CI mode: process::exit bypasses error handling


noahgift:
## Bug

In `validate.rs:419`, `std::process::exit(1)` is called directly instead of returning an error:

```rust
Ok(false) => {
    std::process::exit(1);
}
```

This bypasses the error reporting in main, prevents cleanup, and hardcodes exit code 1.

## Fix
Return `Err(CliError::ValidationFailed("CI assertions failed".into()))` instead.

Issue #361, opened 2026-02-28
apr-qa: 'Format not available for model file' in MVP conversion tests


noahgift:
## Problem

MVP playbook conversion tests fail with "Format SafeTensors/Apr/Gguf not available for model file" for multiple models. The tests expect model files at `output/workspace/{org}/{model}/{format}/` but the files aren't there.

**Affected models (from QA campaign):**
- openchat-3.5 (6x SafeTensors + 6x APR + 6x GGUF = 18 failures)
- deepseek-coder-v2-lite (same pattern)
- phi-3-medium-4k (same pattern)
- qwen3-30b-a3b (6x APR + 6x GGUF)

**Error pattern:**
```
Format SafeTensors not available for model file
Format Apr not available for model file
Format Gguf not available for model file
```

## Five Whys

1. Why "not available"? — Test looks for model file in workspace, not found
2. Why not in workspace? — Conversion step didn't produce output, or produced it in wrong location
3. Why wrong location? — Model was pulled to HF cache but workspace symlink/copy wasn't created
4. Root cause: **MVP workspace setup doesn't populate all format directories from cache**

## Impact

18 false failures per affected model. Combined with G0 config.json issue (aprender#360), models lose ~25 test points unnecessarily.

## References

- Discovered during model QA campaign (PMAT-034)
- Related: aprender#360 (config.json path resolution)

Issue #359, opened 2026-02-28
feat: apr eval --task classify for SSC checkpoint evaluation


noahgift:
Add classification evaluation path to `apr eval` for evaluating SSC (Shell Safety Classifier) checkpoints against JSONL test sets.

- Per-class precision/recall/F1 report
- Confusion matrix
- JSON output mode
- HF model card generation (--generate-card)

Depends on entrenar ClassifyEvalReport + evaluate_checkpoint().

Issue #352, opened 2026-02-27
apr pull uses 55 GB RAM for Qwen3-30B-A3B download


noahgift:
## Bug

`apr pull Qwen/Qwen3-30B-A3B` consumed **55.7 GB RSS** on a Lambda Labs machine (125 GB total). This swamps the machine and makes it unusable for other work.

## Evidence

```
$ ps aux | grep apr
noah 249728 44.6 44.5 148335476 58486596 pts/6 R+ 23:33 10:55 apr pull Qwen/Qwen3-30B-A3B
```

- **RSS**: 58,486,596 KB = **55.7 GB**
- **VIRT**: 148,335,476 KB = **141 GB**
- **%MEM**: 44.5% of 125 GB system RAM
- **Operation**: Just downloading/pulling model files from HuggingFace

## Expected behavior

`apr pull` should stream files to disk without buffering the entire model in memory. A download operation should use O(buffer_size) memory, not O(model_size).

For reference, `huggingface-cli download` for the same model uses ~50 MB RSS.

## Suspected cause

`apr pull` likely reads entire SafeTensor files into memory (for validation or conversion) instead of streaming them to disk. For a 30B-A3B model with ~16 GB of weight files, plus intermediate buffers, this balloons to 55+ GB.

## Impact

- **P1**: Makes Lambda Labs machine (125 GB) unusable during pulls of medium+ models
- Kills co-running QA jobs and finetune tasks
- On Intel (283 GB) this is tolerable but still wasteful

## Suggested fix

Stream SafeTensor shards directly to disk during pull. Defer any validation/conversion to a separate step that can process one shard at a time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

noahgift:
## Root Cause (Five-Whys)

`convert_safetensors_formats()` in `pull_remove_resolve_model.rs` eagerly ran `apr_import()` (SafeTensors→APR) + `apr_export()` (APR→GGUF) during every `apr pull`. These batch converters load the entire model into memory. For Qwen3-30B-A3B (~16 GB weights), this meant loading the model **twice** in succession, consuming ~55 GB RSS.

## Fix (2 files, net -44 lines)

1. **pull.rs**: Removed unused imports (`apr_import`, `apr_export`, `ExportFormat`, `ExportOptions`, `ImportOptions`, `ValidationConfig`)
[...]

PR #3060, opened 2026-09-09
YOGA-NIGHTLY-001 R-3/R-4/R-6: the sm_89 axis, and the assertion that lets it go red

noahgift:
Implements R-3, R-4 and R-6 of paiml/infra `docs/specifications/yoga-nightly-job.md` (paiml/infra#493). Draft — see *Not done* at the bottom.

infra has stood up **`yoga`**: RTX 4060 Laptop, AD107M, **sm_89**, 8188 MiB, CUDA 12.4, org-scoped in the `gpu-x86` group, which aprender is already in. This restores the x86_64 CUDA axis #2740 had to delete when its leg ran on lambda-labs (infra#359).

## The problem an sm_89 lane has

Post-mortem **Lesson 4**: the pre-warm bugs *existed on sm_89 too*, but JIT-on-demand **succeeded** there — sm_121's stricter behaviour is what turned them into hard failures.

> **An sm_89 pass/fail lane would have been GREEN through all seven defects of the Blackwell cascade.**

So the leg's value cannot be "did the tests pass".

## R-3 — the assertion that can go red

`cache_misses == 0` after pre-warm, asserted across a real forward pass. Both caches instrumented — the backward one is **not optional**, since cascade defects **#1810** and **#1813** were backward kernels.

Deliberately **not** the post-mortem's recommendation #3 (cross-architecture cache diffing): legitimate arch differences change cache keys — cuBLAS on sm_121 vs PTX GEMM on sm_89 **is** cascade defect #1804 — yoga and gx10 also differ in toolkit (12.4 vs 13.0), and a transcript nobody must read rots. This invariant is local: one machine, no confound, a verdict that fails.

**My first version was tautological, and the spec's own mutation test caught it before it shipped.** It did pre-warm → reset → *pre-warm again* → assert 0. With the key hardcoded the second pre-warm writes the same colliding key, finds it cached, and passes. It compared pre-warm against pre-warm — the same key construction on both sides. That is post-mortem **Lesson 5** in miniature. Replaced with a real `fused_residual_rmsnorm_forward` pass.

**Measured on real sm_89 hardware, both directions:**

| tree | result |
|---|---|
| correct | **PASS** |
| `warm!` hardcoded key reintroduced | **FAIL** — *"a real forward pass JIT-compiled 1 kernel(s)"* |

## R-4 — the `ada-yoga` leg

A separate job, not a matrix arm: `runs-on` is literal because infra#352's guard can only check labels it can **see**. Same reason `ada`/`yoga` go into `.github/actionlint.yaml` rather than the selector being hidden to dodge the lint.

Preflight **asks the box, never the label** — asserts `compute_cap 8.9` and `x86_64` before running anything. This gate has already been wrong once about which host it was on (#2835: an sm_89 floor asserted on GB10, red four nights).

**No QLoRA** (8 GB, mobile part — gx10 owns training). **No perf assertion**: `SILICON_FLOORS` is keyed on compute capability and yoga answers `8.9`, the same key as the 24 GB / 450 W desktop 4090 the floor came from, so `UNCALIBRATED-SILICON` **cannot fire**. **No yield-to-training predicate** — yoga has no competing workload.

## R-6 — an axis named `cuda` must invoke CUDA

**`--features cuda` appears ZERO times in `silicon-nightly.yml`**, while it declares `aarch64-cuda-sm121`, which holds gx10 for a 90-minute timeout to run `cargo test -p aprender-compute --lib` — a CPU suite. The ledger scores the axis covered; the coverage is of the CPU.

The new guard **fails on committed reality today** (`rc=1`), which is the strongest demonstration available that it discriminates. Four exit paths measured, including **NO-GO (rc=2)** when no cuda-named axis exists — "0 violations over 0 axes" is the shape it exists to refuse. One fixture per *form variant*: `--features cuda`, `--features "cuda,serve"`, `--all-features`, and a cuda-named axis with none.

## Not done

- **R-8** is already written in **#2793** — green for 10 days, 30 commits behind, just unmerged. I did not duplicate it.
- **R-2** (`pre_warm_keys ⊇ runtime_keys` as a CPU-only property test) needs key construction extracted out of `cfg(feature = "cuda")` — a real refactor, not attempted here. R-3 covers the same defect class on a GPU box.
- **R-5** (coverage guard asserting a RUN, not a runner) and **R-7** (re-key `SILICON_FLOORS`) are untouched; R-7 depends on unmerged #2838.

`cargo fmt` clean, `clippy --features cuda` 0 errors, `actionlint` clean repo-wide.

Refs paiml/infra PMAT-272, paiml/infra#493, paiml/infra#361

no-close: the paiml/infra refs (#352 #359 #361 #493) are issues in another repository, and the aprender refs are cited as prior work; none is closed here.


noahgift:
A second axis became promotable today, after this PR was written. Flagging it here so it is not lost when this lands.

`mini-m4` registered (`self-hosted,ARM64,macOS,apple-silicon,m4,mini`), and the coverage guard run against the live fleet now reports **two** PROMOTEs, not one:

```
  ok        x86_64-cpu           served by intel-clean-room-10
  ok        aarch64-cuda-sm121   served by gx10-blackwell
  PROMOTE   x86_64-cuda-sm89     marked pending:361, but yoga-gpu can serve it NOW
  deferred  aarch64-cuda-sm87    pending:361 — no runner yet
  PROMOTE   aarch64-macos        marked pending:361, but mini-m4 can serve it NOW
```

So silicon-nightly stays red on the macos half after this merges. Not a problem with this PR — the anti-rot rule is doing exactly what it says — just the next line in the queue.

I opened #3230 for both axes before checking which PRs already owned this file, and closed it: this one's "COVERAGE IS A RUN, NOT A RUNNER" rewrite is the better answer to the same question, and my version would have conflicted on the header. I will re-do the `aarch64-macos` half on top of the semantics here — a run inside a cadence window with a `job:` binding — once this lands, rather than stack it CI-dark underneath a queued PR.

Worth knowing for that follow-up: darwin/arm64 is not a formality. The first full workspace run on mini found four Linux-only assumptions that no Linux runner can reach by construction — a `/proc/meminfo` read, a `pid+nanos` temp filename that is not unique within one process, an ETXTBSY premise that is a Linux kernel guarantee and simply false on XNU, and a zero-TTL cache entry reporting itself fresh because `Instant::now() > expires_at` needs the clock to have moved. The last one is a production bug, not a test defect.

noahgift:
This PR sat at queue position 1 without merging because its merge group took
`guard-tree` red. Not a flake, and not unrelated to the PR:

```
FAIL  scripts/check_no_pipe_into_grep_q.sh [run]
      | FAIL: 84 sites, ceiling 83.
```

The one new site is this PR's own, `scripts/check_workflow_cargo_packages.sh:198`,
under that file's `set -uo pipefail` (line 46):

```bash
if printf '%s\n' "$MEMBERS" | grep -qxF -- "$tok"; then continue; fi
```

`grep -q` exits on its **first** match, `printf` takes SIGPIPE (141), `pipefail`
reports the pipeline as 141, and the `if` takes the **false** branch though grep
matched — so a `-p <crate>` token that *is* a workspace member would be printed as

```
VIOLATION <file>:<line>  `-p <tok>` names no workspace member
```

Size-dependent, which is the worst kind: ~78 members is about 1 KB and fits the
64 KiB pipe buffer, so `printf` finishes before `grep` exits and the guard stays
quiet today. It starts lying when the workspace grows past the buffer, or on a libc
that flushes differently — and it lies by **inventing** a violation, in a guard whose
entire job is to say when a package name is wrong.

Fixed in `0c6932fd5` with a here-string. Verified:

- the guard still measures the same tree — 17 workflow files, 81 static package
  references, 0 dynamic, 0 violations;
- on the **merged** tree (this branch merged with `origin/main`, which is what the
  merge group builds): `baseline 83  measured 83  PASS`.

`check_no_pipe_into_grep_q.sh` landed on main in #3228 after this branch last
updated, which is why the site was invisible here and only appeared in the group.
The ratchet did exactly what it is for.

Issue #3084, opened 2026-09-10
P0-3: a one-file YAML change costs two full workspace-test trips — 55-min serial quick tier + a FULL merge-queue rerun whenever main moved (67-E1/E2)


noahgift:
Epic: #3078 (release train 0.67.0) · spec: `docs/specifications/06x-release-schedule.md`

**P0-3 — A one-file YAML change costs two full workspace-test trips.** Spec rows **67-E1 / 67-E2** of `docs/specifications/06x-release-schedule.md` (§2 E); milestone 0.67.0. Five-whys handed over by the operator on 2026-09-10, completed from the measured runs:

- **Why did #3072 and #3074 take about two hours each?** CI ran the full-length `workspace-test` on the PR (run `34449608126`: 59 min, of which the step *"Quick tier: every test target that reads the tree (BSE-17)"* was **55 min**), then the merge queue ran again (`merge_group` goes FULL whenever main moved under the PR: `scripts/ci_test_tier.sh` — *"queue tree differs from PR head tree"*; the push-to-main FULL tier is 86 min, run `34448893051`).
- **Why is the quick tier 55 minutes?** `scripts/tree_reader_tests.txt` lists 41 targets across 26 crates and ci.yml chains **one `cargo nextest run -p <crate> …` per crate with `&&`** — 26 separate builds, no shared build graph, inside one 60-minute step (#3070).
- **Why does the queue re-run everything?** The tier decision has no `quick` outcome for `merge_group` when the tree differs — only `reuse` (identical tree) or `full`.
- **Why did nobody see the cost?** No budget existed for a PR job to exceed; 72 pull_request runs on 2026-09-08 alone, each ≈ 2.2 runner-hours, on a 16-runner pool — ≈ 41 % of the fleet on PR-time tests before reruns.
- **Why no budget?** The operator's rule (E) had not been written down as a number a gate reads.

[...]

Issue #3078, opened 2026-09-10
EPIC: release train 0.67.0 — 06x-release-schedule.md §3 (0.67.0)


noahgift:
**Release train 0.67.0** — the release that ships its own binaries, and a 20-minute PR

**Spec of record:** `docs/specifications/06x-release-schedule.md` (PMAT-1097; this epic is §3 "0.67.0" of that document, and §2 holds every row's acceptance command). **Contract:** `contracts/release-schedule-06x-v1.yaml`.

**Window (§1.1):** opens **2026-09-12T06:00Z** · must leave by **2026-09-13T06:00Z** · planned tag **2026-09-12T18:00Z**. `T` is the previous train's actual `publishedAt`; the train leaves on time — scope slips forward to the next epic with `slipped_from: 0.67.0`, dates do not slip. A red release gate holds the train at most 12 h.

Topic epic riding this train: #3062 (NVIDIA CUDA Rust — T0/O2/T3 landed; #3061 #3068 #3066 are 67-B1). P0 five-whys tickets filed under this epic: #3082 P0-1 (67-A1) · #3083 P0-2 (67-C2) · #3084 P0-3 (67-E1/E2).

**Decisions apply at their §6 defaults unless overridden in a comment here** (D-1 FULL tier only nightly + pre-publish dogfood · D-2 NVIDIA Rust default per layer · D-3 yoga advisory→required behind the runner-group restriction · D-4 gx10 takes arch-neutral guards · D-5 trains leave on time · D-6 parity hosts yoga + gx10 · D-7 recipe format · D-8 GH-480 rewriter fate · D-9 non-Linux binaries out). An override comment is copied into the spec's §11 with `decided_by` and the comment link.

## Scope (spec §3, as of `origin/main` @ `2c584a168`)

| Row | Item | Source | Host | State at HEAD |
|---|---|---|---|---|
| 67-A1 | **P0-1** four `apr` assets per tag, verify-apr-assets, dogfood + release-criteria rows | §2 A | yoga, gx10, intel | `v0.66.0`: 0 of 4 `[V]` |
| 67-A1b | backfill `v0.66.0` after #3074 | §2 A | gx10, yoga | #3074 open, `present` check red `[V]` |
| 67-C2 | **P0-2** self-hosted preflight + fleet toolset probe | §2 C | all self-hosted | `gh` missing on gx10 measured on run `34448908554` `[V]` |
[...]

Issue #3070, opened 2026-09-09
BSE-17 quick tier: a PR that touches only ci.yml/tree_reader_tests.txt runs 42 tree-reader targets serially (incl. aprender-train --lib, 7624 tests) and hits the step's 60-minute timeout under fleet load


noahgift:
## Observed
PR #3063 (files: `.github/workflows/ci.yml`, `scripts/tree_reader_tests.txt`), run 34369120088, job `workspace-test` 102563221825:

```
reason=pull_request: no workspace crate touched -> nothing to test; plus 42 tree-reader target(s) from scripts/tree_reader_tests.txt
tree-reader targets: 42
  +  cargo nextest run --profile ci -p apr-cli --lib --test beat_apr_sibling_cli_reach --test cli_commands --test pixel_regression
  +  cargo nextest run --profile ci -p aprender-cgp --lib
  ...
  +  cargo nextest run --profile ci -p aprender-gpu --lib
...
 Nextest run ID 5c48fdaa-… with nextest profile: ci
    Starting 7624 tests across 1 binary (14 tests skipped)          <- aprender-train --lib
        SLOW [>120.000s] aprender-train config::train::tests::train_yaml::test_train_from_yaml_transformer_with_config_file
        SLOW [>120.000s] aprender-train train::transformer_trainer::falsify_lora_tests::test_falsify_f_conv_003_cpu_lora_loss_decreases
##[error]The action 'Quick tier: every test target that reads the tree (BSE-17)' has timed out after 60 minutes.
```

The check-run annotation is the only surviving truth: `The action 'Quick tier: every test target that reads the tree (BSE-17)' has timed out after 60 minutes.` Zero failing tests. The same step passed on #3050 (also a ci.yml-touching PR) in the same hour, on a less loaded runner — the verdict is a function of fleet load, not of the diff (the box had 15/17 runners busy, shared across repos).

## Why it matters
`workspace-test` is a required check; a PR whose diff touches no crate gets the *largest* serial test set the tier owns (`&&`-chained, one `cargo nextest run -p` per crate, each with its own compile), and one of those members (`aprender-train --lib`) carries tests that run >120 s each. The 60-minute step budget cannot hold that under load, so a CI-only PR is the one most likely to time out.

## Asks
1. Bound the tree-reader tier: run the 42 targets through ONE `cargo nextest run` invocation (one compile, parallel test execution), or split the tier into a bounded quick leg and move the slow crates behind it, as BSE-17 already does for touched crates.
[...]

PR #3063, opened 2026-09-09
ci: un-dark aprender-gpu and aprender-cuda-edge in workspace-test (651 tests, 0.193s)

noahgift:
**T0** of [`docs/specifications/nvidia-cuda-rust-library-integration.md`](https://github.com/paiml/aprender/pull/3061), pulled forward into **0.66** because nothing else in that spec is measurable without it. Tracked by #3062 (0.67).

## What this is

Two crates were excluded from the **only required test check** for a CUDA/cuBLAS link their *default* features do not have. Both are `default = []` with no `build.rs`; the `cuda` feature is optional and off.

| Crate | Tests | Run time |
|---|---:|---:|
| `aprender-gpu` | 444 | 0.10s |
| `aprender-cuda-edge` | 207 | 0.03s |
| | **651** | **0.193s** |

Nothing else covered them. `.github/workflows/cuda-nightly.yml:242` is the only other CI reference to either crate and it runs a single `perf053` filter.

> **Correction (2026-09-09).** An earlier version of this description said this PR gives the 109k-LOC PTX generator required-check coverage. **It does not.** In `aprender-gpu`, `driver`, `kernels`, `memory` **and `ptx`** are all `#[cfg(feature = "cuda")]` (`lib.rs:138–151`). The 444 default-feature tests this PR un-darks are `monitor` (308), `testing` (72), `backend` (39), `error` (20), `graph` (4), plus 13 new `launch_budget` tests from #3064. The PTX generator's ~2,160 tests still run only under `--features cuda`, i.e. the gx10 nightly's single `perf053` filter. This PR is a first step, not the fix for that; the gap is tracked separately.

## Measured with the exact CI harness

Not `cargo test` — the command CI actually runs:

```
cargo nextest run --profile ci -p aprender-gpu -p aprender-cuda-edge --lib
  Starting 651 tests across 2 binaries
   Summary [0.193s] 651 tests run: 651 passed, 0 skipped
```

Re-run in a CI-like environment (`CUDA_VISIBLE_DEVICES=""`, empty `HOME`, scrubbed `XDG_CACHE_HOME`) — still green, so the pass does not depend on this box having a GPU or a warm `~/.cache/trueno`.

**Cost:** ~31s extra cold build (263 dep crates, measured in an empty target dir) against the job's `timeout-minutes: 150`.

`aprender-compute` stays excluded — it SIGSEGVs at exit and already runs as its own step below.

## Coupled state checked

- `scripts/ci_test_tier.sh` does not reference these crates
[...]

PR #3089, opened 2026-09-10
perf(ci): P0-3 — the quick tier is ONE build graph (55 min → ~6 min) and the merge queue mirrors the PR instead of paying an hour for a moved main (67-E1 + 67-E2, #3084, PMAT-1098)

noahgift:
## Rows 67-E1 + 67-E2 (P0-3, #3084) — spec `docs/specifications/06x-release-schedule.md` §2 E · epic #3078 · ticket PMAT-1098

**67-E1 — one build graph.** The quick-tier step expanded `scripts/tree_reader_tests.txt` into a `&&` chain of one `cargo nextest run -p <crate>` per crate: 26 cargo invocations, 26 dependency resolutions, serial runs; 55 min on a one-file YAML PR (run 34449608126) and a 60-min timeout under fleet load on #3063 (#3070). Now `scripts/ci_test_tier.sh --filterset` turns the same tokens into one nextest filterset (`(package(x) & kind(lib))`, `(package(x) & kind(bin))`, `binary_id(x::t)`; an unknown token is ENV, never a dropped target — 8 case rows both polarities) and the step runs ONE `cargo nextest run --profile ci --workspace --lib --tests --exclude … -E "<expr>"`. `timeout-minutes: 20` on both quick steps: over that, the quick tier has stopped being quick and fails loudly.

**Measured on lambda, warm target dir:** the same 39 test binaries as the chain; 65,263 tests run in 240 s, 351 s wall including the build. Per-binary counts differ from the chain by feature unification (aprender-core +199, aprender-orchestrate +36, aprender-test-lib −174): the one graph builds the tree-reader libs under `--workspace`-unified features — exactly what the FULL tier builds — so the quick tier now tests the same binaries the full tier does.

**67-E2 — the queue mirrors the PR.** `merge_group` on a moved main re-derives the PR's own diff from `HEAD^1..HEAD` on the queue ref and runs the tier the PR ran (the `reuse` rule is untouched); `push` to main runs quick on the push's diff; `schedule`/`workflow_dispatch` stay FULL (decision D-1: FULL lives in the nightly and the pre-publish dogfood). Over the cap → quick on the touched crates + one `cargo check --workspace --all-targets --locked` step (`check_workspace=1`, compile-level integration of every reverse dependent); a root-manifest diff (`Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`) stays FULL at PR, in the queue and on push — cited from `gate_touched_crates.sh`, not duplicated. A depth-2 fetch on merge_group/push so `HEAD^1` resolves on the shallow checkout — without it the change would look wired and be inert (fail-closed to full with a `::warning::`). 39/39 case rows, incl. a mutant row (a queue that always answers full loses the crates).

Orchestrator re-runs: case table 39/39 (E1 23 + E2), YAML parses, `check_workflow_path_filters`/`check_guards_are_wired` PASS, `make gate` 41 checks 0 failed. Two pre-existing rows encoded the replaced decision (`push -> full`, `aprender-core touched -> full`) and were rewritten, not deleted silently.

### DIFF
--- a/.github/workflows/binary-release.yml
+++ b/.github/workflows/binary-release.yml
@@ -236,6 +236,13 @@ jobs:
           # the same path. The work root is mirrored host<->container; $HOME is not.
           CACHE="$(cd "$GITHUB_WORKSPACE/../.." && pwd)/cache-apr-cuda/${{ matrix.target }}"
           mkdir -p "$CACHE/registry" "$CACHE/target"
+          # The bind-mount TARGET inside the workspace must exist BEFORE the container
+          # starts: when it does not, dockerd creates the mountpoint as root, and after
+          # the container exits the workspace carries an empty root-owned `target/` the
+          # runner (uid 1001) cannot write — run 34488955316 built the asset in 1m54s and
+          # then died on `mkdir: cannot create directory 'target/aarch64-unknown-linux-gnu':
+          # Permission denied`. Pre-creating it as the runner keeps the ownership.
+          mkdir -p "$GITHUB_WORKSPACE/target"
           $DOCKER run --rm \
             -v "$GITHUB_WORKSPACE:/workspace" \
             -v "$CACHE/registry:/usr/local/cargo/registry" \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1922,7 +1922,7 @@ jobs:
   # 3.32.0" is two steps above this one, and 20 machine-specific paths landed
   # through the gap while nothing re-read it.
   pr-review-receipt:
-    runs-on: [self-hosted, X64, Linux, build]   # BP-3 (#3100): any x86 box in the build pool (intel clean-room or yoga-eph)
+    runs-on: [self-hosted, X64, Linux, clean-room]   # NOT the build pool: a non-required review job must never hold a pool slot a required job needs (#3100)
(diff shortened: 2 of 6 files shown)


noahgift:
Folded #3108's two commits into this branch (`fecb358` and `ba69692`, cherry-picked with `-x`). This branch already carries #3107's `RUSTC_WRAPPER` fix, so with its own fast quick tier it is now one PR that can go green on any runner in the pool:

- **#3107's fix:** RUSTC_WRAPPER names sccache directly, because the dash wrapper dropped `CARGO_BIN_EXE_aprender-explain`. It is already on this branch.
- **This PR's own change:** the quick tier is one build graph. The serial tier ran past its 60-minute step limit on a loaded intel box: #3107's run 34535600563 passed every test, and the step still timed out.
- **#3108's fix:** `pp066_v16_defects.sh --v15-red` fetches its v1.5 commit on a shallow checkout, so a fresh yoga container passes guard-tree.

#3108 stays open as a hedge. Whichever lands first puts the pp066 fix on main, and the other becomes a no-op.

Checked locally before the push: `pp066_v16_defects.sh` and its `--v15-red` arm both return rc=0, and `cargo fmt --all -- --check`, `cargo deny check advisories` and `cargo test -p aprender-contracts --lib` are all green.

PR #3098, opened 2026-09-10
ci(release): pre-create the workspace target/ mountpoint before the bullseye CUDA build — the root-owned mountpoint killed rebuild 34488955316 at mkdir

noahgift:
Rebuild run 34488955316 (v0.66.0 CUDA assets, after #3086): the bullseye container built apr in 1m54s on gx10, then the host step died on `mkdir: cannot create directory 'target/aarch64-unknown-linux-gnu': Permission denied`. dockerd creates a missing bind-mount target as root; the workspace kept an empty root-owned `target/` the runner (uid 1001) cannot write. One line: `mkdir -p "$GITHUB_WORKSPACE/target"` before `docker run`. Verified by re-dispatching `binary-release.yml -f tag=v0.66.0` from this branch (the workflow definition comes from the branch; the source is checked out at the tag).

Pmat-Ticket: PMAT-1096


noahgift:

PR #3103, opened 2026-09-10
fix(zram-core,aarch64): NEON LZ4 match copy read unwritten output at offsets 16..63; orchestrate test hard-coded an x86_64 'native' target (gx10 CI P0, PMAT-1098)

noahgift:
Makes aprender's CI suite pass on aarch64 so gx10 can take CI jobs (operator P0 2026-09-10: intel + gx10 + yoga). Measured on gx10 with CI's exact nextest line: 82,075/82,078 pass; these are the three failures.

1. **NEON LZ4 decompressor (real aarch64 correctness bug):** the match copy used `wildcard_copy_neon` whenever `offset >= 16`, but its 32- and 64-byte copies load everything before storing, so for offsets 16..63 they read output bytes that were not written yet (a 16-byte repeating pattern decompressed to zeros after byte 32). Wide copies now run only for `offset >= 64` **and** with 64 bytes of room left in the page (the wildcard path overwrites past `match_len`); offsets 16..63 use exact 16-byte steps with a byte tail. New test `test_neon_roundtrip_offsets_around_copy_widths` covers pattern lengths 15..255 across every width boundary.
2. **orchestrate** `test_build_stage_execute_with_target` passed `x86_64-unknown-linux-gnu` as the 'native' target; it now reads the host triple from `rustc -vV`.
3. Four aarch64-only `as *const/mut` casts that x86 CI never lints now use `.cast::<T>()`.

Verified: `cargo check` + `cargo clippy -D warnings` for `--target aarch64-unknown-linux-gnu` clean; x86 `aprender-zram-core` 406/406, orchestrate test 1/1. Native gx10 run of the three tests follows in a comment.

Pmat-Ticket: PMAT-1098

### DIFF
--- a/crates/aprender-orchestrate/src/pipeline/tests.rs
+++ b/crates/aprender-orchestrate/src/pipeline/tests.rs
@@ -1325,8 +1325,17 @@ edition = "2021"
     .unwrap();
     std::fs::write(project_dir.join("src/main.rs"), "fn main() {}\n").unwrap();
 
-    // Use the native target so compilation should succeed
-    let stage = BuildStage::new(false, Some("x86_64-unknown-linux-gnu".to_string()), false);
+    // Use the HOST triple so compilation succeeds on any architecture. The literal
+    // "x86_64-unknown-linux-gnu" failed on aarch64 (gx10, 2026-09-10): building for a
+    // non-host triple needs a cross toolchain the box does not have.
+    let host = std::process::Command::new("rustc")
+        .arg("-vV")
+        .output()
+        .ok()
+        .and_then(|o| String::from_utf8(o.stdout).ok())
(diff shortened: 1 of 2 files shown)


noahgift:
Native aarch64 verification on gx10 (GB10) at 84f63be6, inside the disposable runner image (rustc 1.93.0, CI=true):
```
cargo test -p aprender-zram-core --lib                                           → 380 passed; 0 failed; 2 ignored   (was: test_neon_roundtrip_small_patterns + test_decompress_simd_medium_offsets FAILED)
cargo test -p aprender-orchestrate --lib -- pipeline::tests::test_build_stage_execute_with_target → 1 passed   (was: FAILED, x86_64 triple on an aarch64 host)
```
The new `test_neon_roundtrip_offsets_around_copy_widths` is in the 380. x86: zram-core 406/406, orchestrate test 1/1.

PR #3106, opened 2026-09-10
fix(serve,x86): AVX2 Q4_K×Q8_K kernels paired each block with the wrong scale — no CI box ever ran them

### LINKED ISSUE #3105: AVX2 Q4_K×Q8_K kernels apply each block scale to half of both blocks and drop 16 Q8 values — reached on every x86 box via the bsums FFN path
## What is wrong

Both AVX2 Q4_K x Q8_K dot kernels in `aprender-serve` compute the wrong value whenever a super-block's scales differ, which on real Q4_K weights is always.

1. `fused_q4k_q8k_dot_avx2` and `fused_q4k_q8k_dot_with_bsums_avx2` pair the nibble halves in the order from before the Q8 load fix. The low nibbles are block `is` and the high nibbles are block `is+1`, but the kernels add the first 128-bit half of each block together (and the second halves together) before applying `sc1` and `sc2`. Each scale ends up covering half of both blocks.
2. `fused_q4k_q8k_dot_avx2` also computes `q8[j+16..j+32]` as `_hsum_hi` and drops it, so block `is`'s min term misses 16 values.

## Why no CI box saw it

`fused_q4k_q8k_dot_simd` takes the best path the CPU has. Every x86 CI box had AVX-512 VNNI (the intel Xeon W-3245 and the lambda Threadripper 7960X), so the parity tests only ever judged the VNNI kernel. The defect surfaced on 2026-09-10, when yoga joined the build pool. Its Core Ultra 9 185H has AVX2 and AVX-VNNI but no AVX-512, and it failed `test_fused_q4k_q8k_dot_simd_*` with rel_err 2.6%.

## Where it is reached

- `fused_q4k_q8k_dot_with_bsums_simd` has no AVX-512 branch, so it runs the AVX2 kernel on every AVX2 box. `fused_q4k_q8k_ffn_up_gate_into`, the Q8K FFN up/gate projection called from `gguf/inference/forward/results.rs`, routes every row through it. Wherever the forward pass takes that path on x86, the defective kernel ran.
- `fused_q4k_q8k_dot_simd` reaches the plain AVX2 kernel only on CPUs without AVX-512 VNNI. That covers most laptops and desktops: Intel Alder, Raptor and Meteor Lake, and AMD Zen 2 and 3.

Not measured: how often the forward pass takes the Q8K FFN path, and what the defect does to real-model outputs. The fix PR does not claim a number for either.

noahgift:
Fixes #3105.

## The defects

Two AVX2 Q4_K x Q8_K kernels have been wrong since the Q8 load-layout fix. No CI box could see it, because every x86 CI box took the AVX-512 VNNI path.

- **Scale pairing.** This affects both `fused_q4k_q8k_dot_avx2` and `fused_q4k_q8k_dot_with_bsums_avx2`. Each block's product sum is now its own two 128-bit halves: `sum_lo_1 + sum_lo_2` for block `is` (sc1), and `sum_hi_1 + sum_hi_2` for block `is+1` (sc2). The kernels used to pair `sum_lo_1` with `sum_hi_1`.
- **Dropped Q8 half.** This affects `fused_q4k_q8k_dot_avx2` only. Block `is`'s Q8 sum now includes `q8[j+16..j+32]`, which the kernel computed as `_hsum_hi` and discarded.

The bsums dispatcher has no AVX-512 branch. The Q8K FFN up/gate projection, `fused_q4k_q8k_ffn_up_gate_into`, calls it for every row. So the pairing defect reached that path on every x86 CPU. The effect on real-model outputs is not measured, and this PR claims no number.

## The falsifier (FALSIFY-QDOT-006, `contracts/quantized-dot-product-v1.yaml` plus its `aprender/` copy)

Three tests call each kernel directly, not through a dispatcher. The fixture has super-blocks whose 8 scales and 8 mins all differ, and Q8 values that differ between the two halves of each block. Tolerance is 1e-3 relative.

| Test | Runs on |
|---|---|
| `test_fused_q4k_q8k_dot_avx2_path_matches_scalar_on_every_avx2_box` | every AVX2 box; asserts AVX2 is present rather than skipping |
| `test_fused_q4k_q8k_dot_with_bsums_matches_scalar_on_every_avx2_box` | every AVX2 box |
| `test_fused_q4k_q8k_dot_avx512vnni_path_matches_scalar` | AVX-512 VNNI boxes; prints SKIP elsewhere |

Measured on lambda (Threadripper 7960X):

| Tree | Result |
|---|---|
| Both fixes | 145 Q4K x Q8K and bsums tests pass, and the VNNI test ran (not skipped) |
| `requires.rs` fix reverted | avx2 n_sb=1: -3650.1 vs scalar -25161.5, RED |
| `bsum_precompute.rs` fix reverted | avx2+bsums n_sb=1: -3329.5 vs scalar -25161.5, RED |

The pre-commit gates pass: format, complexity, SATD and docs.

## Why now

This is the last code blocker for yoga in the build pool (operator P0, 2026-09-10: aprender jobs on intel, gx10 and yoga). yoga's CPU takes the AVX2 path, so every workspace-test there failed the existing parity tests.

### DIFF
--- a/contracts/aprender/quantized-dot-product-v1.yaml
+++ b/contracts/aprender/quantized-dot-product-v1.yaml
@@ -162,6 +162,8 @@ simd_dispatch:
     avx512_vnni: fused_q4k_dot_avx512_vnni
     q8k_scalar: fused_q4k_q8k_dot
     q8k_avx2: fused_q4k_q8k_dot_simd
+    q8k_avx2_kernel: fused_q4k_q8k_dot_avx2
+    q8k_bsums_avx2: fused_q4k_q8k_dot_with_bsums_avx2
     q8k_avx512_vnni: fused_q4k_q8k_dot_avx512vnni_v2
     q8k_4row_avx512_vnni: fused_q4k_q8k_dot_4rows_avx512vnni
   Q5_K:
@@ -235,6 +237,18 @@ falsification_tests:
   test: For each format in YAML, verify simd_dispatch has an entry with at least 'scalar'
     key
   if_fails: Format was added to registry but dispatch table not updated
+- id: FALSIFY-QDOT-006
+  rule: Every x86 Q4_K x Q8_K kernel matches scalar on every box that can run it, not only
+    on the box whose best CPU feature makes the dispatcher select it
+  prediction: fused_q4k_q8k_dot_avx2 and fused_q4k_q8k_dot_with_bsums_avx2 equal fused_q4k_q8k_dot
+    within 1e-3 relative on super-blocks whose 8 scales and 8 mins all differ, on any AVX2 box;
+    fused_q4k_q8k_dot_avx512vnni_v2 likewise on any AVX-512 VNNI box
+  test: cargo test -p aprender-serve --lib -- test_fused_q4k_q8k_dot_avx2_path_matches_scalar_on_every_avx2_box
+    test_fused_q4k_q8k_dot_with_bsums_matches_scalar_on_every_avx2_box test_fused_q4k_q8k_dot_avx512vnni_path_matches_scalar
+  if_fails: A kernel pairs a nibble half with the wrong block scale or drops part of a block's Q8 sum.
+    The dispatcher hid both AVX2 defects for as long as every x86 CI box had AVX-512 VNNI
+    (found 2026-09-10 when yoga, a Core Ultra 9 185H, joined the pool; reverting the fix gives
+    -3650.1 against scalar -25161.5)
 qa_gate:
   id: F-QDOT-001
   name: Quantized Dot Product Contract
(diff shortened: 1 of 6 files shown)


noahgift:
Measured on yoga (Intel Core Ultra 9 185H: AVX2 and AVX-VNNI, no AVX-512), the first CI box to take the AVX2 paths. The run used CI's workspace-test lines with `--no-fail-fast`, so nothing stopped at the first failure. It ran on this PR's tree before the last commit (adc6db306).

| Suite | Run | Passed | Failed |
|---|---|---|---|
| `cargo nextest run --profile ci --workspace --lib` (the three GPU and compute exclusions as in CI) | 82203 | 82202 | 1 |
| `cargo test -p aprender-compute --lib` | 3510 | 3510 | 0 |

The one failure was `quantize::tests::tests_28::test_simd_feature_detection`. It asserted a Threadripper's AVX-512 on every box, and e04bb355f replaces it with ISA-hierarchy checks plus a check that `detect_simd_backend` follows the CPU. No other test depends on an AVX-512 host.

PR #3109, opened 2026-09-10
fix(guard): spec-conformance fixtures are bytes — the zero-width key rendered only under a UTF-8 locale

noahgift:
## What broke
Main's `guard-tree` went RED on yoga-build3 (run 34529896596, job 103047791546):
`BROKE ledger_respend_zero_width_key  expected L1 got CLEAN`. The same tree passes 40/40 on intel.

## Why
The fixture spelled its zero-width space as `\u200b` and rendered it with `printf '%b'`. bash expands `\u`
only under a UTF-8 locale; under POSIX it keeps the literal text, so the key never matches `lambda` and the
respend row reads CLEAN. intel runs en_US.UTF-8; yoga's runner image ran POSIX (fixed on the runner side
since: `LANG=C.UTF-8` in the yoga and gx10 runner env).

## Fix
- The fixture spells the bytes (`\xe2\x80\x8b`), identical under any locale.
- `selftest()` pins `LC_ALL=C`, so a `\u` fixture fails on every box, not only on the odd runner.

## Evidence (a75e15460's tree, run locally)
| | C.UTF-8 | no locale |
|---|---|---|
| before | 40/40 | 39/40 (the BROKE row above) |
| after | 40/40 | 40/40 |
| mutation: `\u200b` restored, selftest pins C | 39/40 RED | — |

bashrs finding count unchanged. Pre-push: `cargo fmt --all -- --check`, `cargo test -p aprender-contracts --lib`,
`cargo deny check advisories` all green.

Milestone 0.67.0: runner parity for the three-host build pool (intel + yoga + gx10).


noahgift:

Issue #3091, opened 2026-09-10
CPU: Qwen3.5 / Qwen3.8 hybrid GGUFs are refused entirely — no fallback exists


alfredodeza:
## Summary

Unlike most GPU-ineligible architectures (e.g. Phi's LayerNorm gap, #3075), which fall
back to CPU inference with a warning, Qwen3.5/Qwen3.8-architecture GGUFs (hybrid Gated
DeltaNet / SSM layers, `general.architecture = qwen35`) are refused **completely** —
there is no CPU inference path at all today. Confirmed with two different downloads:

```
apr chat ~/Downloads/Qwen3.8-27B-UD-IQ2_XXS.gguf
apr chat ~/Downloads/Qwen3.5-4B-UD-Q4_K_XL.gguf
```

Both fail identically, even without `--gpu`:

```
[GGUF model parse failed: Format error: Architecture 'qwen35' uses SSM/Gated Delta Net
layers which are not yet supported for inference. Use a standard transformer model
(e.g., Qwen2.5, LLaMA, Mistral) or wait for SSM support in a future release., will use CPU]
You: hey
Assistant: [Error: Failed to create GGUF model: Format error: Architecture 'qwen35' ...]
```

The chat loop even prints "will use CPU" in the CUDA-init warning line, but then the
*CPU* load of the same GGUF fails with the exact same error — because the refusal is in
the shared GGUF transformer constructor (`QuantizedGGUFTransformer::from_gguf`), which
both the CPU and GPU code paths call. There is no degraded-but-working CPU mode; the
model simply cannot be loaded by this binary in any form.

## Root cause

`crates/aprender-serve/src/gguf/transformer.rs::unsupported_architecture_reason` is
called unconditionally inside `QuantizedGGUFTransformer::from_gguf` (GH-704), before
backend selection, and returns `Err(FormatError)` whenever any tensor name contains
`ssm_`/`ssm.`. This is upstream of the CPU/GPU fork — it blocks both.

## Ask

1. Implement a scalar/SIMD **CPU** Gated DeltaNet (linear attention) forward pass for
   GGUF-loaded hybrid layers — recurrent state + short causal conv1d + gated delta rule,
   per Qwen3.5/Qwen3-Next's architecture — analogous to what
   `crates/aprender-serve/src/gpu/scheduler/linear_attn.rs` already sketches for the GPU
   scheduler (SafeTensors path), but for the GGUF `QuantizedGGUFTransformer` CPU path.
[...]

PR #3110, opened 2026-09-11
fix(quantize): Q5_K readers decoded an invented layout — every Q5_K tensor on the CPU path and in apr import was wrong (PMAT-1101)

noahgift:
## Q5_K readers decoded an invented layout (PMAT-1101). P0 correctness, blocks #3091

**What was wrong.** ggml's `block_q5_K` stores 256 values in 176 bytes: `d`, `dmin`, 12 bytes of 6-bit scales and mins, `qh[32]`, then `qs[128]`. Sub-blocks `2c` and `2c+1` share the same 32 `qs` bytes, low nibbles first and then high nibbles. The fifth bit of value `l` in sub-block `s` is bit `s` of `qh[l]`. Three aprender readers each did something different:

| Reader | Used by | What it did |
|---|---|---|
| realizar `dequantize_q5_k` + `fused_q5k_dot` | every CPU Q5_K matmul (`apr run` / `serve` / `chat`) | read the two nibbles of one byte as neighbouring values, and took the fifth bit from `qh[4s + l/8]` |
| aprender-quant `dequantize_q5_k_to_f32` | `apr import` (GGUF → APR) | gave each sub-block its own 16 `qs` bytes |
| aprender-quant `quantize_q5_k` | quantizing to Q5_K | packed the same wrong way |

**Measured** on Qwen3.5-0.8B-Q4_K_M, where `attn_qkv` and `ssm_out` are Q5_K:
- On `blk.0.attn_qkv` (6144×1024), realizar and gguf-py (llama.cpp's own Python reader) disagree on **5,959,751 of 6,291,456** values (max |Δ| 0.59).
- Layer-0 QKV projection sum: realizar **+26.64**, llama.cpp **−17.06**, gguf-py −17.10. After the fix realizar gives **−17.0967**, and layers 0–2 of the #3091 op-by-op trace match llama.cpp.

**Why nothing caught it.** Every Q5_K test compared one aprender reader against another. realizar's dequantizer and fused dot shared one invented layout; aprender-quant's quantizer and dequantizer shared another.

**Fix**
- realizar: a single block walker, `for_each_q5k_value`, emits values in ggml order. `dequantize_q5_k` and `fused_q5k_dot` both go through it, so the two can no longer disagree.
- aprender-quant: the reader and the packer now use ggml's layout.

**FALSIFY-QDOT-007** (`contracts/quantized-dot-product-v1.yaml`, both copies) pins the readers to the 256 values gguf-py produced from a real super-block that llama.cpp quantized. The fixture is embedded in the tests.
- realizar `quantize::tests::q5k_ggml::*`: dequant is bit-exact; the fused dot with a one-hot activation returns each value exactly; the parallel matvec matches the gguf-py dot.
- aprender-quant `tests::test_q5k_ggml_*`: dequant is bit-exact; a quantize round-trip on data whose sub-blocks all differ stays within 0.15.

**Mutation check.** Each change below was reverted locally, and the tests went red every time:
- realizar nibble shift: RED
- realizar qh bit: RED
- aprender-quant reader: RED
- aprender-quant packer (sub-block halves swapped): RED

The packer mutant **survived** the existing ramp round-trip test, because a ramp gives every sub-block the same 5-bit codes. That is why the distinct-sub-block test exists.

**Local gates**
- Formatting and tests:
  - `cargo fmt --check` ✓
  - `cargo test -p aprender-quant --lib` ✓
  - `cargo test -p aprender-serve --lib -- quantize::`: 2223 passed
  - the four Q5_K integration targets ✓ (`quantize_coverage`, `property_quantize`, `gguf_quantize_coverage`, `quantize_property_tests`)
- Contracts and repo checks:
  - `pv validate`, both contract copies ✓
  - `cargo test -p aprender-contracts --lib` ✓
  - `check_baseline_ratchets.sh` ✓
  - `cargo deny check advisories` ✓
  - `guard_tree.sh --no-cargo` ✓, except `check_complexity_ratchet.sh`: the local pmat is not the pinned 3.39.0, so that failure is the instrument, not the code.
- Clippy:
  - `cargo clippy -p aprender-quant --all-targets -- -D warnings -A unused-variables` ✓
  - `cargo clippy -p aprender-serve --lib --tests` reports 5 errors that already exist on main, none in files this PR touches.

**Not in this PR:** the CUDA Q5_K kernels in `aprender-gpu` (`q5k/gemv.rs`, `q5k/gemm.rs`) carry two more wrong layouts. They are tracked in a separate issue and will get a GPU-side golden test.

Refs #3091

Pmat-Ticket: PMAT-1101


noahgift:

PR #3093, opened 2026-09-10
fix(pv): an empty corpus is refused with exit 2, never reported as PASS (PVL-1, PMAT-1099)

noahgift:
## PVL-1 (PMAT-1099): `pv` refuses to report over ZERO contracts

Spec: paiml/infra `docs/specifications/PVL-001-pv-lean-gate.md` §2 row EV-1. Ground truth: PV-LEAN-AUDIT-001 (2026-09-10 10:47Z).

### Facts re-verified at HEAD 43b8d8e8c (genchi genbutsu)
| cited fact | HEAD | note |
|---|---|---|
| `pv lint /nonexistent-path` → rc 0, `Result: PASS` over 0 contracts | VERIFIED | in-tree pv 0.66.0 built by `scripts/pv_bin.sh` [V] |
| `pv lint <empty dir>` → rc 0 | VERIFIED | [V] |
| `pv proof-status <empty dir>` → rc 0, `Proof Status (0 contracts)` | VERIFIED | [V] |
| `pv verify-pipeline <empty dir>` → "No contracts found", rc 0 | VERIFIED (not in the spec row; plan-grill finding, `verify_pipeline.rs:26-29`) | [V] |
| `pv lint|coverage|graph <single contract file>` → walked nothing, rc 0 | VERIFIED (validate gate `contracts: 0`; `Contracts: 0`; `Nodes: 0`) | [V] |
| `contract_walk::collect_contracts` used by graph, proof-status, lean-status, verify-pipeline; lint and coverage walk on their own | VERIFIED (`pmat query --literal "contract_walk::"`) | [V] |

### Change
- `contract_walk.rs`: `ZeroContracts { path, filter }` (Display `0 contracts under <path>`), `ZERO_CONTRACTS_EXIT = 2`, `collect_corpus()` (dir walked · single file = one-contract corpus · missing path = empty), `require_contracts()`, `exit_code_for()`.
- `lib.rs`: a `ZeroContracts` dispatch error exits **2** (nothing was measured — clap's usage-error code); every other error keeps exit 1 (measured, failed).
- `lint.rs`: `refuse_empty_corpus()` reads the gate **named** `validate` (count = parsed + parse errors) before any report is printed. A shape-only match was measured refusing every VALID corpus, because `duplicate-stems` reuses the `Validate` detail with zeros. `lint/gates.rs`: a single file is linted as a one-contract corpus (no cache beneath a file).
[...]

PR #3112, opened 2026-09-11
fix(compute,aarch64): gx10 could not run ci / lint — ARM-only clippy errors, and #2567's parallel Q4_K path was never called (PMAT-1102)

noahgift:
## gx10 could not run `ci / lint`: clippy errors only an ARM box sees (PMAT-1102)

The gx10-pool1 canary picked up #3089's `ci / lint` and failed it (job 103137079500). Lint is clean on x86. On aarch64, `cargo clippy --all-targets -- -D warnings -A unused-variables` fails, and so does the job's `-p aprender` fallback, because the failing crates sit in the facade's dependency closure. **pool1 is stopped until this lands** (pool2 not started), since every lint job it picked up would fail.

**What only fails on ARM**
- **Unreachable tails in 8 SIMD detectors.** Each ends `#[cfg(aarch64)] { return Neon; } Scalar`, so the tail is unreachable on ARM. Locations: aprender-compute ×3, present-terminal, the aprender-core demo, aprender-serve ×2 (one of them a test), trueno-ublk. The tails are now cfg'd blocks.
- **x86-only helpers compiled dead on ARM.** The Q6_K value extractor, the BLIS prefetch, NT-store and GEMV-tile thresholds, and the 32×6 packing import now carry their callers' `x86_64` cfg. The Q6_K chunk function is gated `x86_64 or test`.
- **Unused import.** rounding.rs imported the NEON backend and never used it.
- **#2567's parallel Q4_K path for ARM was never called.** #2567 made the non-x86 `matmul_q4k_f32_parallel` actually parallel (its doc records 1.21× on GB10), but its only call site sits inside the `x86_64` block. Every ARM machine still ran the serial kernel. The dispatcher now calls it on non-x86 above the same 8M-element threshold, and `test_q4k_parallel_matches_serial_on_every_arch` checks it against the serial kernel on every arch. The coverage module that exercised it only builds on x86_64.
- **compute.rs.** Two x86_64-only BLIS packers are dead on ARM. The file already carries 11 complexity violations, and the pre-commit gate refuses any edit to it, so `mod compute;` carries a non-x86 `expect(dead_code, reason = …)`. `expect` turns into an error as soon as the packers stop being dead. The proper fix is to decompose compute.rs and put the cfg on the functions.

**Verified locally, with CI's exact lint chain**
- aarch64 (cross, `--target aarch64-unknown-linux-gnu`): `cargo clippy --all-targets -- -D warnings -A unused-variables` is clean (it had 17 errors). The `-p aprender` fallback is clean too.
- x86_64: workspace `cargo clippy --all-targets -- -D warnings -A unused-variables` is clean.
- `cargo test -p aprender-compute --lib`: 3511 passed, plus the new test.
- fmt ✓, `check_baseline_ratchets.sh` ✓, `guard_tree.sh --no-cargo` ✓, except `check_complexity_ratchet.sh`, which fails because the local pmat is not the pinned 3.39.0 (instrument, not code).

**After merge:** restart gx10-pool1, then start pool2 (operator P0: aprender jobs run on intel + yoga + gx10).

Pmat-Ticket: PMAT-1102


noahgift:
Measured from the PVL-001 driver (#3093 waits on this PR: its `ci / lint` fails on gx10-pool2 with the aarch64 `aprender-compute` errors this PR fixes).

`ci / security` here failed at 10:58:19Z on **gx10-build** in step `roadmap-valid`: `pmat is not on PATH — an absent verifier is a NO-GO`. Read-only on gx10 at 11:2xZ:

| fact | value |
|---|---|
| host `/opt/ci-tools/pmat` | present, mtime 2026-09-10 17:44 (with `pv`, `forjar`, …) |
| gx10-build container binds | `/opt/ci-tools/pmat:/usr/local/bin/pmat:ro` |
| inside the container | `/usr/local/bin/pmat` = pmat 3.39.0; `.path` = `/home/runner/bin:/home/runner/.cargo/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:…` |
| containers | all pools recreated 11:15:0xZ by `run-ephemeral-docker-v3.sh` (ephemeral recycling) |

So the current gx10-build incarnation *can* see pmat, and the step's own `export PATH=/opt/fleet-bin/bin:$HOME/.cargo/bin:/usr/local/cargo/bin:$PATH` keeps `/usr/local/bin`. I could not establish what the 10:58 incarnation saw (no log of its binds). Not re-running it from here: a rerun is not a pass until the cause is named. If you know what changed between 10:58 and 11:15, that is the cause.

noahgift:

Issue #3111, opened 2026-09-11
CUDA Q5_K gemv/gemm read a non-ggml block layout (qh bit order; gemm also qh/qs offsets and scale unpacking)


noahgift:
Found while fixing the CPU readers in #3110 (PMAT-1101). I read the CUDA Q5_K kernels in `crates/aprender-gpu/src/kernels/quantize/q5k/` against ggml's `block_q5_K` (`d`, `dmin`, `scales[12]`, `qh[32]` at offset 16, `qs[128]` at offset 48; value `l` of sub-block `s` takes its fifth bit from bit `s` of `qh[l]`; sub-blocks `2c`/`2c+1` share `qs[32c..32c+32]`, low nibbles then high). Neither kernel matches.

**`gemv.rs` (`Q5KGemvKernel`, the M=1 decode path; `WeightQuantType::Q5K => GemvKernel::Q5K`)**
- `qs` offset and nibble selection match ggml.
- The fifth bit is read from `qh[val_idx / 8] >> (val_idx % 8)`, a sequential bitmask. ggml puts it at `qh[val_idx % 32] >> (val_idx / 32)`, so it differs for most values.

**`gemm.rs` (`Q5KKernel`)**
- Assumes `qs` sits at 16..144 and `qh` at 144..176. ggml has `qh` at 16..48 and `qs` at 48..176.
- Takes sub-block `s`'s `qs` bytes from `16 + s*16 + lane/2`, and the bit from `(s*32 + lane)` as a sequential bitmask.
- Unpacks the scales as a 12-bit stride (`bit_offset = s*12`) and normalises by `1/63`. That is not ggml's `get_scale_min_k4`.

**Not measured yet.** This comes from reading the code. The plan is to reuse the gguf-py golden super-block from FALSIFY-QDOT-009 (#3110) as a GPU-side test: gemv against the gguf-py dot, and gemm against the reference matmul. It will run on lambda (sm_89) and gx10 (sm_121), then the kernels get fixed.

Impact, if confirmed: `--gpu` on any model with Q5_K tensors (every Q5_K_M/Q5_K_S file, plus the Q5_K tensors inside Q4_K_M files) computes wrong projections on CUDA, and the CUDA release assets ship that path.

noahgift:
[...]

PR #3113, opened 2026-09-11
fix(cuda): the Q5_K GEMV read each value's fifth bit from a sequential bitmask — every Q5_K tensor decoded wrong on the GPU (#3111)

noahgift:
## CUDA Q5_K GEMV read each value's fifth bit from the wrong place (#3111, PMAT-1104). P0

ggml's `block_q5_K` keeps the fifth bit of value `l` in sub-block `s` at bit `s` of `qh[l]`, i.e. byte `val_idx % 32`, bit `val_idx / 32`. `Q5KGemvKernel`, the M=1 decode path that `WeightQuantType::Q5K` dispatches, read byte `val_idx / 8`, bit `val_idx % 8`, as if the bits were a sequential bitmask.

**Measured on an RTX 4090 (lambda).** Run on the FALSIFY-QDOT-007 golden super-block (#3110), which is real bytes from a llama.cpp-quantized GGUF:
- the shipped indexing gives a ramp dot of **2.659**; gguf-py gives **0.622**
- with the fix, the one-hot and ramp tests both pass (2/2, no skips, on the GPU)

The only existing GPU Q5_K test asserted `q5k_gemv(...).is_ok()` on synthetic weights, so nothing could see the wrong values.

**FALSIFY-QDOT-008** (`cuda::executor::tests_q5k_ggml`, needs an NVIDIA GPU):
- a one-hot activation must return each of the 256 gguf-py values
- a ramp must reproduce gguf-py's dot

The golden block becomes a shared `include!()` fixture (`quantize/tests/q5k_ggml_fixture.rs`) used by the CPU and GPU tests. Contract entries are in both copies of `quantized-dot-product-v1.yaml`.

**Stacked on #3110.** This branch is #3110's plus this fix; the diff shrinks to 7 files once #3110 merges.

**Not in this PR:** `Q5KKernel` (the GEMM) has its own wrong layout (qh/qs offsets, a 12-bit scale unpack). Inference never dispatches it; only `q5k_matvec` in the layer-trace path and tests reach it. It stays on #3111.

Refs #3111

Pmat-Ticket: PMAT-1104


noahgift:

PR #3099, opened 2026-09-10
fix(gguf): Qwen3.5/Qwen3.8 hybrid GGUFs — the refusal names BOTH backends and the CLI stops promising a CPU fallback (#3091 ask 2, #3090 related; PMAT-1098)

noahgift:
## Qwen3.5 / Qwen3.8 hybrid GGUFs (Alfredo's #3091 ask 2; #3090 related) — PMAT-1098

**Before:** `apr chat --gpu` on a `qwen35` GGUF printed `[GGUF CUDA init failed: … will use CPU]` and then failed again on the CPU load with the same format error — the message promised a fallback that does not exist.

**After (this PR):**
- `unsupported_architecture_reason` says **NEITHER the CPU nor the GPU backend implements Gated DeltaNet/SSM layers yet**, names the triggering tensor (`blk.0.ssm_a`) and the tracking issues #3090 (GPU) / #3091 (CPU).
- The chat/run CLI prints a format-level architecture refusal once, without "will use CPU", makes no second load attempt and exits non-zero.
- `apr qa` reports it as a structured **Capability Match** failure (no crash); the `qwen3_5` row in `arch_constraints_fallback.rs` + `contracts/model-families/qwen3_5.yaml` describe the real hybrid (Gated DeltaNet linear layers + gated full attention every `full_attention_interval`).

**Measured (lambda, `~/models/Qwen3.5-0.8B-Q4_K_M.gguf`, HEAD of this branch):**
```
cargo test -p aprender-serve --lib -- unsupported_architecture   → 3 passed
cargo test -p apr-cli --lib -- chat_load                          → 1 passed
apr run <gguf> --prompt hi --max-tokens 1                         → rc=8, output contains "neither", no "will use CPU"
apr qa <gguf>                                                     → ✗ FAIL Capability Match … NEITHER the CPU nor the GPU backend implements …
cargo clippy -p aprender-serve -p apr-cli --lib -- -D warnings; cargo fmt --check → 0 / 0
make gate                                                         → rc=0
```
Gap left open (ask 4, half): `apr validate` exits 0 and stays silent on the capability; tracked with the CPU implementation (#3091 ask 1, branch `PMAT-1098-qwen35-cpu-deltanet`, in progress).

Receipt stays `status: partial` per G-11. Produced by an agy goal lane; the lane committed in the shared checkout (isolation violation recorded in the PMAT-1098 receipt), acceptance re-run by the orchestrator.

Pmat-Ticket: PMAT-1098

[...]

Issue #3090, opened 2026-09-10
GPU: implement Gated DeltaNet / SSM inference for Qwen3.5 / Qwen3.8 hybrid GGUFs


alfredodeza:
## Summary

Any GGUF whose tensors include hybrid linear-attention (Gated DeltaNet / SSM) layers
is refused before inference ever starts — including on GPU. Confirmed with two
different downloads that both report the same internal architecture string:

```
apr chat --gpu ~/Downloads/Qwen3.5-4B-UD-Q4_K_XL.gguf
apr chat --gpu ~/Downloads/Qwen3.8-27B-UD-IQ2_XXS.gguf
```

Both fail identically:

```
[GGUF model parse failed: Format error: Architecture 'qwen35' uses SSM/Gated Delta Net
layers which are not yet supported for inference. Use a standard transformer model
(e.g., Qwen2.5, LLaMA, Mistral) or wait for SSM support in a future release., will use CPU]
```

Note: the 27B file is named "Qwen3.8" by its uploader but reports `general.architecture
= qwen35` internally — same code path, same refusal, same gap. Worth confirming in triage
whether "Qwen3.8" GGUFs in the wild are literally Qwen3.5-architecture checkpoints under a
different marketing name, or whether the GGUF writer just didn't update the arch string —
either way, this repo's refusal logic treats them identically today.

## Root cause

`crates/aprender-serve/src/gguf/transformer.rs::unsupported_architecture_reason` scans
tensor names for `ssm_`/`ssm.` and refuses unconditionally (GH-704) before any backend
(CPU or GPU) is selected:

```rust
let has_ssm = tensor_names.into_iter().any(|name| name.contains("ssm_") || name.contains("ssm."));
if has_ssm {
    return Some(format!("Architecture '{architecture}' uses SSM/Gated Delta Net layers ..."));
}
```

This is a hard, format-level refusal — not the `RequiredOp`/`gpu_supported_ops()`
capability gate used for e.g. Phi's LayerNorm gap (#3075), which at least falls back to
CPU. There is currently no forward-pass implementation for Gated DeltaNet in the GGUF
inference path at all.

## What exists already (starting point, not a working path)

There is real scaffolding for Gated DeltaNet, but it's wired to the **SafeTensors/APR**
scheduler path, not GGUF, and appears untested end-to-end:

- `crates/aprender-serve/src/gpu/scheduler/linear_attn.rs` — `LinearAttnWeights`,
[...]

Issue #3038, opened 2026-09-07
feat: Qwen3.5 (qwen35) SSM/Gated Delta Net inference support


alfredodeza:
## Summary

`apr run --gpu` (and presumably CPU) on a Qwen3.5-family GGUF refuses inference with a clear, intentional error rather than a crash — but the underlying capability (SSM / Gated Delta Net hybrid-attention inference) is still unimplemented. Filing this as a tracked low-priority feature request for actual support, since the current behavior is "fail fast with a good message," not "works."

## Repro

```
apr run --gpu ~/Downloads/Qwen3.8-27B-UD-IQ2_XXS.gguf
=== APR Run ===

Source: /home/alfredo/Downloads/Qwen3.8-27B-UD-IQ2_XXS.gguf
error: Inference failed: Format error: Architecture 'qwen35' uses SSM/Gated Delta Net layers which are not yet supported for inference. Use a standard transformer model (e.g., Qwen2.5, LLaMA, Mistral) or wait for SSM support in a future release.
```

Note on the filename: the GGUF is named `Qwen3.8-27B-UD-IQ2_XXS.gguf` (Unsloth "UD" dynamic-quant naming, `IQ2_XXS`), which doesn't match any published Qwen version number I'm aware of (3, 3.5, 3-Next...). Regardless of what the download is labeled, `apr`'s own GGUF metadata read reports `general.architecture = "qwen35"`, so the routing decision is based on the file's actual metadata, not the filename — worth double-checking against the specific file if anyone picks this up, but I don't think the architecture string itself is wrong.

## Is the error message correct?

As far as I can verify from the source, yes — this looks intentional and accurate, not a bug in the message itself:

- `crates/apr-cli/src/commands/inference_output.rs` has a test (`tests_2403::inference_context_appears_exactly_once`) that pins this exact string: `"Format error: Architecture 'qwen35' uses SSM/Gated Delta Net layers ..."`. That test was added for #2403, which fixed a *doubled* `"Inference failed:"` prefix — the message content itself predates and survives that fix.
[...]

PR #3114, opened 2026-09-11
feat(qwen35): Qwen3.5 / Qwen3.8 hybrid GGUFs run on the CPU — apr run + apr chat (#3091)

### LINKED ISSUE #3303: CPU parity reference for qwen35: build it from the GPU methodology, not from the PR it measures
Child of the 0.68 Qwen 3.5 epic **#3038**. Operator ruling 2026-09-15: *"A CPU parity
reference is required: build it from the existing GPU methodology (same thresholds shape,
CPU host, llama.cpp `[X]`), not from the PR it measures. Evidence introduced by the PR
under test is not evidence."*

## Why this exists

PR #3114 currently supplies its own parity evidence —
`crates/aprender-serve/examples/qwen35_parity.rs` and
`evidence/parity/qwen35/lambda/parity.json`, **both absent from `main`**. A number that
cites only the change under test cannot judge it.

And there is no CPU-inference parity reference on `main` at all: the stack
(`docs/specifications/0.66-performance-parity-report.md`, `evidence/parity/thresholds.yaml`,
`evidence/parity/LEDGER.md`, `scripts/perf_gate.sh`) is GPU-hosted and dense-7B-scoped.
68-F1 is yoga + gx10 only.

## What already works, unchanged

**Correction to an earlier claim of mine:** the parity stack is GPU-only in its
*baselines and thresholds*, **not** in its *gate machinery*.

| piece | state on `origin/main` |
|---|---|
| host | `scripts/perf-matrix.yaml` declares **`intel`**: `accelerator: cpu`, `compute_class: cpu`, `comparator: llamacpp`, `ci_runner: self-hosted` — the only host with `compute_class: cpu` AND a runner (lambda/gx10/yoga are `cuda`; `mini` is `status: NA`, no Metal inference path, #2841) |
| `scripts/perf_gate.sh` | host-class-parameterised throughout (`--host`, `cell_status()`, `mx_host.get("compute_class")`, `reachable_by` at 237–244). 1,993 lines, **zero hardcoded GPU host** |
| threshold schema | `thresholds.yaml` `models: {}` is the declared extension point — *"per-model overrides land here with n and the command that measured them"*. A qwen35 CPU row is additive, no schema change |
| comparator pin | `scripts/llama_pin.toml` `build_commit = 39173bcac`, resolver `scripts/llama_bin.sh` refuses anything else |

## What must change

1. **A qwen35 CPU row under `models:` with its OWN basis.** Reusing `min_cosine: 0.98` is
   not available: its basis is *GPU-vs-CPU* cosine on qwen2.5-coder-7b/1.5b (known-good
   0.998607 / 0.998465, known-bad 0.950827 / 0.950611, n=5). That population does not
   cover CPU-vs-llama.cpp on a hybrid SSM model. **Long pole** — needs a known-good and a
   deliberately-broken qwen35 GGUF.
2. **Move `PARITY_GATE_COSINE_MIN` out of the CUDA module.** Defined at
   `crates/aprender-serve/src/gguf/cuda/mod.rs:873`, used at `mod.rs:140,148,159,180` and
   `mod_parity_gate.rs:90,96,123`. A CPU gate cannot read a constant living under
   `gguf/cuda/`. **The one unavoidable code change.**
3. **A new workload id.** `perf-matrix.yaml` declares `workloads: [W1, W2]` and
   `perf_gate.sh --workload` is a closed set. qwen35 is a different model from W1's, so
   this is a `perf_gate.sh` change, not just YAML. Note `intel`'s existing `D1` cell is
   `MEASURED` but its own comment says *"perf_gate.sh --workload accepts only W1|W2 … a
   RECORDED seed that no arm consumes yet; nothing here is ARMED."*
4. **A CPU-only llama.cpp build of `39173bcac` on intel.** The pin was recorded from the
   **CUDA** binary on lambda ("5 cuda/cublas libs linked, RTX 4090 sm_89"). Same commit,
   CPU build, recorded as the string that binary prints. **Do not bump the pin** — the
   file states a bump invalidates every ratio measured against `39173bcac`.
5. **Ledger + consumers.** A `RECORDED` row in `LEDGER.md` with `what_it_lacks[]` filled
   honestly (CONFORMANT needs PP-21/25/26/28/30, a 0.69 item per `06x:176`). Consumers:
   `parity_admission.rs`, `parity_per_op.rs`, `extended_commands.rs`,
   `tests/reg15_admission.rs`, `scripts/check_model_parity.sh`.

## Independence from #3114 — achievable, with one honest caveat

Items 1–4 touch `thresholds.yaml`, `perf-matrix.yaml`, `perf_gate.sh`, `llama_pin.toml`,
`LEDGER.md` and the constant's home — **disjoint from #3114's 10 files**.

The caveat: the reference can be *built* independently but cannot be *exercised on
qwen35* until a CPU qwen35 forward exists, which today lives only in #3114. Correct shape:
land the reference with the qwen35 cell `status: UNMEASURED` + owner + expiry — the same
pattern `intel`'s W1 already uses — so **the gate exists and is armed before the code it
will judge**.

## Falsifier

`scripts/perf_gate.sh --host intel --workload <new-id>` runs, reads the qwen35 row from
`thresholds.yaml`, and reports a verdict against a llama.cpp CPU `[X]` number — with the
cell moving UNMEASURED → RECORDED only when a measurement with `n` and its command exists.

## Open, needs a decider

A **deliberately-broken qwen35 GGUF** for the known-bad half of the basis. No such fixture
exists; the 0.98 basis used qwen2.5-coder-1.5b for this. Who produces it, and does it live
in `evidence/` or as a generated fixture?

noahgift:
## Qwen3.5 / Qwen3.8 hybrid GGUFs run on the CPU (#3091 ask 1, PMAT-1098). DRAFT: the 0.67 gate is not strictly green yet

`apr run` and `apr chat` on a `general.architecture = qwen35` GGUF now run a CPU Gated DeltaNet + gated-attention forward instead of refusing. The dense loader still refuses these files, so other entry points (`serve`, `bench`) refuse honestly rather than produce garbage. The GPU path is #3090 (0.68).

### Parity with llama.cpp master, the decided gate for 0.67
Measured with `examples/qwen35_parity.rs` on the 0.8B fixture; full record in `evidence/parity/qwen35/lambda/parity.json`.

| prompt | free-running greedy | teacher-forced | the disagreement |
|---|---|---|---|
| 0 (5 tokens) | diverges at step 0 | 27/28 | llama.cpp's token is realizar's **rank 2**, gap **0.145** logits |
| 1 (10 tokens) | identical, 23/23 | 23/23 | none |
| 2 (16 tokens) | diverges at step 13 | 16/17 | rank 2, gap **0.185** |

**Strict greedy equality is not green on 2 of 3 prompts.** Every divergence is a near-tie: llama.cpp's choice is realizar's second choice, within 0.19 logits. The likely cause is precision: llama.cpp's CPU matmuls quantize activations to Q8_K and its KV cache defaults to f16, while realizar's are f32. The example also runs a margin-aware gate (llama.cpp's token must be realizar's rank ≤ 2 with a gap ≤ 0.25), and that gate PASSES. The operator's 0.67 rule asks for greedy 32-token parity, so this stays a **draft**. Next step: match llama.cpp's activation and KV precision, then re-measure.

### What was wrong
Found op by op against llama-eval-callback's dump:
- **Forward pass, eight defects:**
  - per-head q/k L2 norm (it ran globally, twice)
  - no RoPE in the DeltaNet path
  - partial NEOX RoPE for attention (n_rot 64 of 256)
  - a stray L2 norm on the DeltaNet output
  - beta went through sigmoid twice
  - SiLU applied twice
  - the attention output gate was never applied
  - attention head width 256, not hidden/heads = 128
- **Q5_K weights.** The Q5_K readers decoded an invented layout. That fix is #3110, merged under this branch.
- **Refusal.** The lane had dropped the dense loader's refusal with no CPU path wired. #3099's check is restored.

### Verified (lambda)
- **Tests:** `cargo test -p aprender-serve --lib -- qwen35` passes 10 tests, including new unit tests for per-head L2, the sigmoid gate and partial NEOX RoPE. Reverting per-head L2 or the NEOX half turns them red.
- **Lint and build:** clippy `-D warnings` on aprender-serve and apr-cli passes. `cargo check -p apr-cli --features cuda` passes; the lane's base-model literal did not compile under CUDA and is fixed here.
- **`apr run`:** "What is 2+2? Answer in one word." gives `Four.`, and "The capital of France is" gives `The capital of France is **Paris**.`.
- **`apr chat`:** answers coherently and exits cleanly.
- **Dense regression:** qwen2.5-coder-0.5b gives `The capital of France is Paris.`.

**Depends on #3099 (honest refusal) and #3110 (Q5_K).** This branch carries both, so the diff shrinks once they merge.

**Merge note (#3331 squash):** the two conflicts — `crates/aprender-serve/benches/gguf_real.rs` and `crates/aprender-core/src/text/bpe/tests_encode_decode.rs` — were the SAME portability fix landed twice, independently, by #3331 and by this branch. `main`'s side was taken because it carries the fuller rationale (`models_root()`'s three-step resolution order and the sibling S1/S2 citation). No behaviour differs between the two versions and nothing of this branch's own work was dropped. Not a semantic decision.

Closes #3091

Pmat-Ticket: PMAT-1098

no-close: #3303 (the CPU parity reference) and #3208 (the Q8_K/f16-KV parity follow-up) are cited as prior work and as the measurement this PR rests on; neither is closed here. #3091 IS closed: this PR is the CPU support it asks for, and the parity row deliberately stays [U] with no threshold set.

ont-delta: reason QE2E-BND-002 — the bound F <= 2*P + O(seq_len*d*L) stated no constant, so no input could falsify it; it is now F(s) <= 2*P + 4*L_attn*d_kv*s, asserted as an equality on the slope. No obligation is added or removed.


noahgift:
## State refresh — the conflict is resolved, the diff shrank as predicted, and the draft rationale is now stale

### Mergeable again
The branch was **DIRTY** on eight files. Every conflicting hunk was this branch's own
Q5_K work as it looked *before* main renumbered the obligation (`FALSIFY-QDOT-009` →
`007`) and modernised the slice API (`chunks_exact` → `as_chunks::<N>().0.iter()`).
Main is strictly ahead: same reader fix under the final id, **plus** an eighth
obligation this branch predates — `FALSIFY-QDOT-008`, the CUDA Q5_K GEMV from
#3111/PMAT-1104. So main's side was taken on all eight; nothing of #3091's work is in
those files.

`roadmap_union.py` reports `main 872 + 0 own -> 872 entries; added=[]` — this branch
introduces no roadmap entry, so taking main's copy dropped nothing there either.

### The diff shrank exactly as this PR said it would

> *"Depends on #3099 (honest refusal) and #3110 (Q5_K). This branch carries both, so
> the diff shrinks once they merge."*

Both are closed and their work is on main. **31 files / 2571 insertions → 10 files /
1904 insertions.**

### Verified on the merged tree
```
cargo check -p aprender-serve --lib --locked            ok
cargo test  -p aprender-serve --lib -- qwen35           10 passed, 0 failed
cargo test  -p aprender-serve --lib -- quantize::     2223 passed, 0 failed
cargo test  -p aprender-quant --lib                     19 passed, 0 failed
```

### The reason this is a draft no longer applies as written

> *"The operator's 0.67 rule asks for greedy 32-token parity, so this stays a draft."*

**0.67.0 shipped and published on 2026-09-13 08:43Z** without it, and this PR is
milestoned **0.68.0**. The gate it is waiting on is a rule about a release that is over.

### Recommendation, with the evidence rather than as a question

**Land it for 0.68 on the margin-aware gate, and track exact greedy parity separately.**

1. **#3091's ask was not parity.** It is *"hybrid GGUFs are refused entirely — no
   fallback"*. A CPU forward that answers `Four.` and `…is **Paris**.` discharges that
   ask; refusing to ship it until a *different, stricter* property holds leaves the
   reported defect open in order to pursue an unreported one.
2. **The margin-aware gate passes** — llama.cpp's token is rank ≤ 2 within ≤ 0.25
   logits on every prompt; measured gaps 0.145 and 0.185. Teacher-forced is 27/28,
   23/23, 16/17.
3. **The divergence has a named, non-speculative cause**: llama.cpp quantizes CPU
   matmul activations to Q8_K and defaults its KV cache to f16; realizar's are f32. So
   this is a precision-matching task with a known shape, not an unknown defect — the
   right thing for its own issue with its own evidence, not a blocker on a feature that
   works.
4. **Nothing regresses.** The dense loader still refuses these files, so `serve` and
   `bench` refuse honestly instead of producing garbage, and qwen2.5-coder-0.5b is
   unaffected.

If 0.68 does want strict greedy equality as a release gate, that is an operator call
and this stays a draft until the precision work lands — but it should be a **decision
taken for 0.68**, not a rule inherited from a shipped release.

I have not undrafted it: the draft was a deliberate, documented choice, and changing it
is the author's or the operator's to make.

noahgift:
## Operator ruling 2026-09-15 — stays draft, and here is what unblocks it

**Your CI red was not your code.** `ci` and `workspace-test` — the two branch-protection required checks — were **SUCCESS** on run 34954974551. What failed was fleet-wide instrument drift:

```
guard-cargo  FAIL pmat pinned 3.40.0 found 3.40.1
guard-tree   cb200_baseline.txt recorded under pmat 3.40.0, runner has 3.40.1
gate         aggregates the two above
```

That is #3291, and it red-lined every open PR, not just this one. Fixed by **#3301**, merged 2026-09-15T11:09:40Z — `tools.toml` now pins 3.40.1 with the baselines restamped. Your re-run on `5b055c0ca` is against the corrected pin.

## Ruling: stays DRAFT until it carries discharged contract rows

Not because of CI. Because this PR adds **1,133 lines of new forward-pass code and zero contract rows**, while three qwen35 contracts sit un-discharged on `main`:

- `contracts/qwen35-hybrid-forward-v1.yaml`
- `contracts/gated-delta-net-v1.yaml`
- `contracts/qwen35-e2e-verification-v1.yaml`

Discharge them with RED→GREEN evidence — as commits here or as a PR ahead of this one. **If the code cannot discharge them, the code is not done.** The rows are the definition of Qwen 3.5 support, not a follow-up to it (CLAUDE.md Rule 7). Green required checks do not override that.

## The parity gate is the existing stack — and this PR cannot supply its own reference

Gate: `evidence/parity/thresholds.yaml` `min_cosine: 0.98` on logits, already on `main` with a measured basis, applied to the CPU path against **llama.cpp CPU as `[X]`**.

- **Greedy-token equality is not the gate.** It fails on legitimate quantization numerics — which is exactly why #3208 was carved out, and that carve-out is upheld.
- **`rank ≤ 2, gap ≤ 0.25` is not a gate either.** No basis; §3.3 forbids a threshold without one.
- Both become **report-only ledger fields**: greedy agreement and margin, per prompt. A threshold waits for **20 records**.

**And:** `examples/qwen35_parity.rs` + `evidence/parity/qwen35/lambda/parity.json` are introduced by this PR and absent from `main`. A number that cites only the change under test is not evidence. The CPU parity reference gets built from the existing GPU methodology — same `thresholds.yaml` shape, CPU host, llama.cpp `[X]` — and lands independently of the PR it judges.

Worth stating plainly: there is **no CPU-inference parity reference on `main` at all** today. The whole stack (`0.66-performance-parity-report.md`, `thresholds.yaml`, `LEDGER.md`, `perf_gate.sh`) is GPU-hosted and dense-7B-scoped; 68-F1 is yoga+gx10 only. Building that reference is part of the epic, tracked at #3038.

## Scope

CPU is **in** 0.68 — #3091 is the floor and this PR discharges it. GPU (#3090) ships in 0.68 if merged by the cut, else 0.69.

Ticket tree consolidated: #3038 is the epic, #3090/#3091/#3208 its children, all four now P1 (the parent had been P3 under P1 children).

Your draft rationale note from 2026-09-13 was right that the 0.67 greedy rule is stale — the reason to stay drafted is now the contracts, not the greedy gate.

noahgift:
## Contracts and parity evidence now in this PR — status for review

Pushed `29205600f`: the contract-test commits, with `origin/main` merged in cleanly. `cargo test -p aprender-serve --lib qwen35` passes 19/19 on lambda.

### Obligations (tests beside the forward, in `forward_qwen35_contract_tests.rs`, run by CI's `workspace-test --lib` with no wiring)

| contract | discharged | every body mutation-verified |
|---|---|---|
| `gated-delta-net-v1` | **5/5** | already on main via #3322 |
| `qwen35-hybrid-forward-v1` | **6/7** | 15 mutations, all red |
| `qwen35-e2e-verification-v1` | **3/7** | ORD-003, INV-006, CON-007 |

What remains is contract defects, not untested code:

- **QHF-BND-005**: the formal `∀l: ||h_l||_inf <= M for finite M` was vacuous. It is amended to explicit finiteness (branch stacked on #3316); a test body is in progress.
- **QE2E-ORD-003**: false below one K-quant super-block. Amended with an `n % 256 == 0` precondition; the test already draws whole blocks.
- **QE2E-INV-001**: a fact about one 9B checkpoint, not code. Its equation also lacked the untied lm_head term (fixed; `tied_embeddings: false` confirmed).
- **QE2E-BND-002** (O() with no constant), **QE2E-BND-005** (self-referential coverage) and **QE2E-MON-004** (a throughput claim): need `N/A` with a reason and owner. The schema has no N/A form (`AppliesTo`'s `#[serde(other)]` would parse `N/A` as an algorithm target), so `AppliesTo::NotApplicable` plus typed reason/owner is being added first.

API gap found by the tests: there is no per-sublayer entry point. `forward_attention` and `forward_deltanet` run their mixer and FFN together, so the tests isolate a sublayer by zeroing the other's output projection.

### Parity against llama.cpp — measured, same host

apr's CPU forward (this PR, `5b055c0ca`) against llama.cpp `d1d3c3396` raw float32 logits, all 78 positions of the corpus prompt, both on intel:

```
min cosine   0.995905  (position 4)
mean cosine  0.998989
below 0.98   0 of 78
argmax       4 mismatches (positions 2, 28, 36, 73) — all rank-2, on near-ties:
             the reference's own top-2 gap there is 0.43, 0.63, 0.013, 0.11
```

The reference is deterministic (n=5 byte-identical) and was falsified against llama-perplexity (0 of 9,436,312 words differ). The orchestrator re-ran the comparator independently and the numbers match. Named confounds: llama.cpp prefilled in one batch while apr went token by token, the subject is a program calling this forward rather than the `apr` binary, there is one prompt, and there is no known-bad Qwen3.5 model yet (so no threshold — the row stays `[U]`, fail-closed).

This PR stays a draft until its CI is green and QHF-BND-005 has a body. The draft ruling's stated reason, zero contracts, no longer holds.

Issue #3174, opened 2026-09-12
aprender-train-canary is dark: workspace-excluded, in no CI job, pinned to burn 0.21.0-pre.2 — there is no measured trueno-vs-Burn number


noahgift:
## The defect

`crates/aprender-train-canary` is the only thing in the tree that measures aprender against Burn — it benchmarks trueno's WGPU matmul against `burn::tensor::Tensor::matmul` across five LLM-shaped problem sizes and prints a ratio:

```rust
//! Performance canary: trueno WGPU vs Burn WGPU throughput
let ratio = burn_ms / trueno_ms;
eprintln!("  [{m}x{k}x{n}] trueno={trueno_ms:.1}ms ... burn={burn_ms:.1}ms ... ratio={ratio:.2}x");
```

**It never runs.** Three independent reasons, all verified 2026-09-12:

1. **Excluded from the workspace.** Root `Cargo.toml:114` lists it under `exclude`, and `Cargo.toml:47` records the reason (`sqlite3/burn link conflict`). It therefore has its own `[workspace]` stanza and is invisible to `cargo metadata`, `cargo check --workspace`, and every workspace-wide CI job.
2. **Not in any CI workflow.** No `.github/workflows/*.yml` names it.
3. **Pinned to a pre-release that has been superseded.** `crates/aprender-train-canary/Cargo.toml:8`:
   ```toml
   burn = { version = "0.21.0-pre.2", features = ["wgpu"] }
   ```
   Burn 0.21.0 released **2026-05-07** and 0.22.0-pre.3 on **2026-08-25**. The canary is pinned to a pre-release of a version that shipped four months ago.

## Consequence

**There is no current measured trueno-vs-Burn number.** The comparison is asserted by the existence of the file and by nothing else. Any claim about relative GPU matmul throughput against Burn is unsupported until this runs.

This is the dark-target class the repo already has rules about — benches are dark because CI builds `--examples` and never benches; a new `tests/*.rs` is dark until added to the one explicit line in `ci.yml`. This is the same failure with a third mechanism: workspace exclusion.

## Note on the comment that is already there

`Cargo.toml:143` says *"aprender-train-canary is intentionally NOT a member — see the exclusion"*. The exclusion is intentional; being **unrunnable** is not the same decision, and the two have been conflated. A crate can be excluded from the workspace and still have a CI job that builds and runs it in its own directory.

## Acceptance criteria (each must be able to FAIL)

[...]

Issue #3173, opened 2026-09-12
verify gate reads descriptive prose as a test citation — rsplit("::") ignores classify_binding, fails the build over a sentence


noahgift:
## The defect

The `verify` lint gate resolves a falsification test's binding like this — `crates/aprender-contracts/src/lint/gates_extended.rs:81`:

```rust
let raw = test_name.trim().trim_matches('"');
let name = raw.rsplit("::").next().unwrap_or(raw);
if name.starts_with("test_") || name.starts_with("prop_") {
    total_refs += 1;
    if !source_tests.contains(name) { /* PV-VER-001 error */ }
}
```

It takes the tail after the **last** `::` in the *entire* `test:` string and, if that tail begins with `test_` or `prop_`, treats it as a citation of a Rust test that must exist in `src/`.

That heuristic cannot distinguish a **citation** from a **description**. Any prose containing a `::`-qualified name is read as a binding, including prose that explicitly says the test does not exist yet.

## Reproduction

A `LIVE-PENDING` falsification body — the repo's own sanctioned state for a gate that cannot bind yet (`strict_test_binding.rs:119`) — that happens to name the test it will one day bind to:

```yaml
test: >-
  LIVE-PENDING - U and V are orthonormal. No test surface exists today because
  the capability is unimplemented. PROMOTE by authoring `svd::tests::prop_orthonormal_factors`
  in `aprender-compute`; until then the obligation is recorded, NOT satisfied.
```

`classify_binding` correctly returns `Unbindable` for this (it starts with `LIVE-PENDING`), so strict-test-binding ignores it. But the `verify` gate does **not** consult `classify_binding` at all — it applies its own `rsplit("::")` rule to the raw string and reports:

```
[ERROR] PV-VER-001 — Unfalsifiable: test `prop_orthonormal_factors` in `aprender-compute`;
until then the obligation is recorded, NOT satisfied.` not found in src/
```

The garbled test "name" in that message — it runs from `prop_orthonormal_factors` to the end of the sentence — is itself the tell that a prose string was parsed as an identifier.

Two contracts in aprender#3146 hit this. The workaround was to write module paths with `/` instead of `::`, which is a change to *documentation wording* to satisfy a *code* heuristic — the wrong direction.

## Why it is worth fixing rather than working around

[...]

Issue #3172, opened 2026-09-12
CRUX master registry drift: 250 stories vs 275 contracts — categories L and M have 24 contracts with no story rows


noahgift:
## The drift

`contracts/crux-competitive-research-ux-v1.yaml` is documented as the master registry — the source of truth from which sub-contracts are scaffolded, and the file `scripts/crux_scaffold_contracts.py` reads via `yq -o json .stories`.

It has drifted from the contracts on disk. Measured on `main`, 2026-09-12:

| | count |
|---|---|
| `stories:` rows in the master | **250** |
| `contracts/crux-[A-Z]-*.yaml` files | **275** |
| story categories present | `A B C D E F G H I J K` |
| **category L contracts** | **15, with 0 story rows** |
| **category M contracts** | **9, with 0 story rows** |

So 24 contracts exist with no entry in the registry that is supposed to enumerate them. L is "HF kernels-community integration" (competitor `hf-kernels-community`, 15 contracts) and M is "APR-QA Playbook Canonicalization" (competitor `apr-qa-playbook`, 9).

## Why it matters

- `coverage_intake` in the master declares `total: 250`. That figure is now a count of the registry, not of the corpus — the corpus is 275. Any report derived from `coverage_intake` under-reports by 24 contracts, and nothing validates the two against each other.
- The scaffolder's round-trip is broken for L and M: it emits contracts *from* stories, so stories that do not exist cannot regenerate the 24 contracts that do.
- It is the precedent an author naturally copies. Category N (aprender#3146) deliberately did **not** follow it — all 17 of its rows are registered and its `coverage_intake` counts are asserted against the story histogram — but that was a choice made after noticing the drift, not something the tooling enforced.

## Second defect, same area, fixed in #3169 — recorded so it is not re-introduced

`scripts/crux_scaffold_contracts.py`'s `CATEGORY_NAMES` map stopped at `K`. L and M had existed as contracts for months, so re-running the scaffolder over them would have emitted a bare letter as the category name. #3169 adds L, M and N. Nothing detected this in the interim, which is the same root cause: no gate compares the scaffolder's vocabulary to the corpus.

## Acceptance criteria (each must be able to FAIL)

- [ ] A test asserts a **bijection**: every `contracts/crux-<CAT>-<NN>-v1.yaml` has a story row with matching `id`, and every story row's `contract:` file exists. It must fail in **both** directions — an orphan contract and an orphan story are each an error.
[...]

Issue #3171, opened 2026-09-12
.claude/ holds 58 stray worktrees, 25 GB and 571,815 .rs files inside the repo root — and the ignore rule is local-only


noahgift:
## What happened

A tree-wide line count of the repository returned **222,601,319 lines across 581,978 `.rs` files**. The true figure is 3,947,294 lines across 10,274 files. The measurement was wrong by ~56x because `.claude/` inside the repository root holds a fleet of stale agent worktrees, each a full checkout plus its build artifacts.

Measured 2026-09-12 in `/home/noah/src/aprender`:

| | |
|---|---|
| `du -sh .claude/` | **25 GB** |
| worktrees under `.claude/worktrees/` | **58** |
| `.rs` files under `.claude/` | **571,815** |
| files under `.claude/` tracked by git | **6** |

## Why this is a defect and not just disk

1. **It silently corrupts any tree-wide tool that does not special-case it.** Anything walking the repo — a line counter, a custom lint, a `find`-based guard, a grep sweep, an embedding indexer — reads 571,815 extra `.rs` files as project source. The failure is silent and the result is plausible-looking, which is the worst combination. This is the guard-universe defect class the repo already has a rule about: a tool whose universe is built from the wrong side.
2. **The ignore rule is not shared.** `.claude/worktrees/` is excluded via `.git/info/exclude`, not `.gitignore`:
   ```
   .git/info/exclude:5:**/.claude/worktrees/
   ```
   `.git/info/exclude` is per-clone and never committed. Every other clone of this repo — CI runners included — has no such exclusion, so the protection exists only on this one machine.
3. **58 worktrees is an accumulation, not a working set.** Agent lanes register worktrees under the repo root and do not remove them.

## Related, already-known

Memory records that the delegate goal lane leaves a REGISTERED stray worktree under `repo_root` at `.claude/worktrees/lane-*/`. This ticket is that leak measured at scale, plus the observation that the containment (`.git/info/exclude`) is local-only.

## Acceptance criteria (each must be able to FAIL)

- [ ] The exclusion moves to a **tracked** file (`.gitignore`), so every clone inherits it. A test asserts `git check-ignore -v .claude/worktrees` resolves to `.gitignore`, **not** `.git/info/exclude` — checking only that it is ignored would pass today and prove nothing.
[...]

PR #3175, opened 2026-09-12
docs(roadmap): five defects the linfa/burn sweep tripped over, four of them in the tooling doing the sweeping

noahgift:
Files #3170–#3174. Roadmap rows only, no code changed.

None of these were the task — all five were hit while doing something else, which is the only reason they surfaced.

| # | Defect | Evidence |
|---|---|---|
| **#3170** P1 | `apr-agent` accepts any flag as a task slug | `apr-agent --help` created a worktree, branch and claim lockfile. The guard `case $slug in *[!a-zA-Z0-9._-]*)` rejects characters **outside** that class — `-` is inside it. Also: apr-agent is untracked, yet `.git/hooks/pre-commit` blocks every commit and directs you to it |
| **#3171** P1 | `.claude/` = 58 worktrees, 25 GB, 571,815 `.rs` files in the repo root | A tree-wide count returned **222,601,319** lines; the truth is 3,947,294 — wrong by 56×, and plausible-looking. The exclusion lives in `.git/info/exclude`, which is per-clone and protects one machine |
| **#3172** P2 | CRUX registry: 250 stories vs 275 contracts | Categories L (15) and M (9) have contracts and **no story rows**. `coverage_intake` says 250 |
| **#3173** P2 | verify gate reads prose as a test citation | `gates_extended.rs:81` `rsplit("::")` over the whole string; never consults `classify_binding`, which handles `LIVE-PENDING` explicitly. Fails a build over a sentence |
| **#3174** P2 | `aprender-train-canary` is dark | Workspace-excluded, in no workflow, pinned to `burn 0.21.0-pre.2` four months after 0.21.0 shipped. **No measured trueno-vs-Burn number exists** |

## ACs written to exclude the cheap fix

- **#3171** asserts `git check-ignore` resolves to `.gitignore`, *not* `.git/info/exclude` — asserting merely that it's ignored passes today.
- **#3172** requires the bijection to fail in **both** directions.
- **#3173** requires a genuine bare citation to still raise PV-VER-001, so a fix that just stops reporting cannot pass.
- **#3174** requires the canary **built, run, and its ratio compared** against a floor — or deleted and the claim withdrawn. Leaving it dark closes nothing.

### DIFF
--- a/docs/roadmaps/roadmap.yaml
+++ b/docs/roadmaps/roadmap.yaml
@@ -17099,3 +17099,98 @@ roadmap:
   labels:
   - kind:triage
   notes: null
+- id: PMAT-3170
+  github_issue: 3170
+  item_type: task
+  title: apr-agent accepts any flag as a task slug
+  status: planned
+  priority: high
+  assigned_to: null
+  created: 2026-09-12T12:28:29Z
+  updated: 2026-09-12T12:28:29Z
+  spec: null
+  acceptance_criteria:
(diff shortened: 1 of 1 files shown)


noahgift:
Dequeued by the 0.67.0 release train (APR-RELEASE-001 §3.4, heijunka: one aprender PR in CI at a time). This PR's merge-group run was sharing the fleet with the release bump #3145 while the bump's workspace-test was on intel.

Nothing is wrong with the PR: it is roadmap-only and additive (+95/-0). Auto-merge is re-armed on it right after v0.67.0 is tagged; it then ships in 0.68.0.

PR #3164, opened 2026-09-12
docs(release): APR-RELEASE-001 spec + PMAT-1098 receipt for session 1 — the 0.67.0 train

noahgift:
Draft until the train reaches its terminal step. Carries the operator's spec (untracked until now), the PMAT-1098 receipt-so-far (andon: K crossed at k=312), the R-17 routing row, and the two agy lane receipts under evidence/release/0.67.0/agy/. The §7 report and the docs/build-ledger train record are added when v0.67.0 is tagged or SKIPPED.

Refs #3078, PMAT-1098.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3170, opened 2026-09-12
apr-agent accepts any flag as a task slug — 'apr-agent --help' created a worktree, a branch and a claim lockfile


noahgift:
## What happened

Running `apr-agent --help` to read its usage did not print usage. It created a git worktree, a branch, and a claim lockfile, then tried to launch a Claude session:

```
✓ created worktree /home/noah/src/aprender-worktrees/--help on agent/--help (off origin/main)
→ claude in /home/noah/src/aprender-worktrees/--help   (slug=--help, branch=agent/--help, claim held)
Error: Input must be provided either through stdin or as a prompt argument when using --print
```

Residue left behind, all of which had to be cleaned by hand:

| Artifact | Path |
|---|---|
| worktree (registered with git) | `/home/noah/src/aprender-worktrees/--help` |
| branch | `agent/--help` |
| claim lockfile | `~/.config/apr-agent/claims/--help.lock` |

## Root cause — the slug charset admits every flag

`~/.local/bin/apr-agent`:

```bash
slug="${1:-}"
[ -n "$slug" ] || die "usage: apr-agent <task-slug> [-- <claude args...>]"
shift || true
[ "${1:-}" = "--" ] && shift || true
case "$slug" in
  *[!a-zA-Z0-9._-]*) die "slug must match [a-zA-Z0-9._-]  (got '$slug')";;
esac
```

The guard rejects characters **outside** `[a-zA-Z0-9._-]`. `-` is inside that class, so `--help` contains no rejected character and passes. So does `-h`, `--version`, `--dry-run`, and every other flag a user would try first. There is no `--help` handler and no leading-dash check.

This is the failure mode the repo already has a rule about — hand-rolled arg parsing that silently accepts what it should reject (cf. simular's `match args[1]` dropping `--seed`).

## Second, separable defect: the tool is not in the repo

`git ls-files | grep apr-agent` returns **0**. `apr-agent` lives only at `~/.local/bin/apr-agent`, and nothing under `scripts/` or in the `Makefile` references or installs it.

Yet `.git/hooks/pre-commit` refuses every commit in the main checkout and directs the user to it:

```
✗ apr-agent guard: refusing to commit in the MAIN aprender checkout.
  → Autonomous work:  apr-agent <task-slug>   (isolated worktree + claim)
```

[...]

Issue #3159, opened 2026-09-12
P2: Gaussian + Sparse random projection with the Johnson-Lindenstrauss bound


noahgift:
Parent: #3146 · Tier **P2 — completion** · Milestone **0.70.0**

## The gap

`RandomProjection` absent repo-wide (verified 2026-09-12). linfa-reduction ships random projections, marked Tested.

## Why it is worth filing

It is the cheapest item in this epic and it has a property none of the other reduction methods have: the Johnson-Lindenstrauss lemma gives a **distance-preservation guarantee that does not depend on the data**. Projecting to `O(log(n)/eps^2)` dimensions preserves all pairwise distances within `(1 +/- eps)` with high probability, with no fit pass over the data at all.

That is directly useful to work already in this repo: the neighbour index from T3 degrades in high dimension (every kd-tree does), and `crates/aprender-core/src/citl/pattern/mod.rs:25` already carries a comment describing its index as a "simplified linear scan for now". A random projection in front of a neighbour search is the standard fix.

## Scope

1. `GaussianRandomProjection`.
2. `SparseRandomProjection` (Achlioptas / Li et al. sparse matrix — faster and sparser).
3. `johnson_lindenstrauss_min_dim(n_samples, eps)` as a public helper.
4. `n_components='auto'` deriving the dimension from the JL bound.

## Acceptance criteria (each must be able to FAIL)

- [ ] **The JL guarantee is tested as a guarantee**: for 1000 random points in 5000 dimensions projected to `johnson_lindenstrauss_min_dim(1000, 0.1)`, the fraction of pairwise distances distorted by more than 10% is asserted below a documented failure probability. Checking a mean distortion instead of the tail would not test the lemma.
- [ ] Distortion tightens as eps tightens — asserted across at least 3 eps values, monotonically.
- [ ] `SparseRandomProjection` achieves the same distortion bound as the Gaussian version on the same fixture, with an asserted density below 1/3.
- [ ] Reproducibility: identical seed produces an identical projection matrix.
- [ ] The projection matrix is generated lazily / never materialized densely for the sparse variant — asserted by a memory or allocation-count check, since "sparse" that allocates dense is the defect here.
- [ ] Oracle parity vs `sklearn.random_projection` on `johnson_lindenstrauss_min_dim` — exact integer agreement.
- [ ] Composes with the T3 neighbour index: a test projects high-dimensional data then runs k-NN, asserting recall against exact brute-force search in the ORIGINAL space above a documented threshold.

[...]

Issue #3147, opened 2026-09-12
P0: SVD substrate in aprender-compute (Golub-Reinsch + randomized) — no SVD exists repo-wide


noahgift:
Parent: #3146 (Pareto gap-closure vs linfa/Burn) · Tier **P0 — substrate** · Milestone **0.68.0**

## The gap

There is no SVD anywhere in this repository. Verified 2026-09-12:

```
grep -rnE 'pub (struct|fn) .*(Svd|SVD|svd)' crates/aprender-compute/src/   -> 0 hits
grep -rlniE 'truncatedsvd|randomized_svd|halko' crates/aprender-core/src/  -> 0 hits
```

The only dense decomposition available is `SymmetricEigen` (`crates/aprender-compute/src/eigen/mod.rs:72`), which requires a symmetric input. Every algorithm that wants a singular value decomposition therefore either forms a covariance/Gram matrix first (numerically worse, see #NEXT_02) or does not exist.

## Why this is ranked first

It is the single missing primitive with the most downstream dependents. Four other tickets in this epic are blocked on it or degraded without it:

- PCA (currently takes the covariance route — the only measured sklearn **loss** in `docs/BEATS.md`, 18.6x slower)
- TruncatedSVD (does not exist)
- PLSRegression (does not exist)
- KernelPCA / Nystroem (do not exist)
- LSA over `text/` vectorizers (`TfidfVectorizer` exists with nothing to reduce it)
- Whitening in `preprocessing/`

Building one correct SVD closes the substrate for all of them. Building each of them separately does not.

## Scope

New module in `crates/aprender-compute/` (the `trueno` lib), alongside `eigen/`:

1. **Thin/economy SVD**, Golub-Reinsch (Householder bidiagonalization + implicit QR on the bidiagonal). Must handle both `m >= n` and `m < n` without the caller transposing.
2. **Randomized SVD** (Halko-Martinsson-Tropp): range finder with oversampling `p` and `q` power iterations, for `k << min(m,n)`. This is what makes large PCA tractable.
3. Deterministic **sign convention** (largest-magnitude component of each left singular vector is positive), so results are reproducible and diffable across runs and backends.

## Acceptance criteria (each must be able to FAIL)

[...]

PR #3200, opened 2026-09-13
feat(release): T-5 reconcile guards — check_pr_closes_issue.sh + check_reconcile.sh (APR-RELEASE-001 §6, kaizen)

noahgift:
APR-RELEASE-001 §6 **T-5 Reconcile** is a hard gate of every train (operator ruling 2026-09-13, kaizen). Measured on the 0.67.0 train: 308 open issues, 424 opened vs 124 closed in 30 d, 48 merged PRs of which 9 carried a closing reference, 880 remote branches of which 847 had no PR. Filing was the work product; closing was nobody's.

Two guards, each with a must-RED/must-GREEN self-test (mutation-checked: neutering the fail path turns the self-test red):

- `scripts/check_pr_closes_issue.sh` — a PR body that cites `#N` must use `Closes/Fixes/Resolves`, or carry `no-close: <reason>`. 9-case table.
- `scripts/check_reconcile.sh <tag> --prev-tag <prev> --json OUT` — the five predicates R1 fixed-but-open, R2 missing closing refs, R3 PR-less branches older than 14 d, R4 DIRTY PRs older than a train, R5 closure/arrival; exit 1 unless R1..R4 are 0, exit 2 on env. 17-case offline self-test via `RECONCILE_INPUT_DIR`.

First real receipt (v0.66.0→v0.67.0, after archiving 776 dead branches to `refs/archive/`): R1=0 R2=28 R3=0 R4=1 R5=22/61=0.36. Wiring R2 into ci.yml and `check_reconcile.sh` into the autopilot's `close` step follows in the next PR.

no-close: the spec amendment lives on PR #3164; this PR adds the guards only.


noahgift:
Ran this guard against the real tree before it merged — `check_reconcile.sh v0.67.0
--prev-tag v0.66.0` — and recorded the result as a T-5 ledger receipt in **#3212**. The
self-test is green (17 cases) and the live path works. Three findings about the
predicates, for whoever picks them up:

**R2 needs an epoch.** `check_pr_closes_issue.sh` is not on main yet, so every one of the
35 PRs in `v0.66.0..v0.67.0` merged before any such check existed — R2 = 35 of them, and
it cannot ever reach zero for that window. Worse, most of those citations are *correct*:
#3160 and #3175 **filed** their follow-ups (#3147–#3159, #3170–#3174) and cite them
without closing, which is what a backlog-opening PR should do. The `no-close:` line is
exactly the remedy for that shape, and those bodies could not carry it. Counting only PRs
merged after the guard was armed on main would make R2 a ratchet instead of a permanent
red.

**R3 conflates dead with stranded.** It named `feat/y2-calibrate` and
`feat/y1-7bgarbage`. Both carry unique unmerged work. The first holds `evidence/perf-060/`
— 791 insertions and the only measurement of mini's throughput that exists, which matters
now that mini is a full-time build host. Deleting it to zero the counter would have
destroyed it; it is landed as #3211 instead and R3 fell 2 → 1. Suggest splitting the
predicate: *no open PR and no unique commits* = dead; *no open PR with unique commits* =
stranded.

**R4 is volatile against main.** It read 0 at 11:44Z and 1 at 11:50Z, because #3164,
#3162 and #3201 merged as one queue group in between and made #3041 dirty. Against a
moving base it reports the queue's timing rather than the tree's health.

None of this blocks this PR — the guard measures what it says it measures. It is what the
first real run found.

PR #3205, opened 2026-09-13
ci(mini): macos-arm64 leg on mini-m4 (NEON tests + core check) and the darwin nightly comes back

noahgift:
## Why

The operator ruled 2026-09-13 that mini (Apple M4) is a **full-time aprender build host**. Measured state: the `mini-m4` runner registration had been auto-deleted by GitHub on 2026-09-12 for inactivity, and no aprender workflow could select a macOS runner anyway (every job names `Linux` or the `clean-room` docker pool). Re-registered at 10:07Z with labels `self-hosted,macOS,ARM64,apple-silicon,m4,mini` — online.

## What

- `ci.yml`: new `macos-arm64` job on `[self-hosted, macOS, ARM64, apple-silicon, mini]` — `cargo check -p aprender-core --lib` and `cargo test -p aprender-compute --lib` (the NEON SIMD path no Linux runner exercises). **Not in `gate`'s needs**: one 16 GB runner must never hold a merge. Timeout 40 `[U]`.
- `nightly.yml`: `build-darwin` restores the `aarch64-apple-darwin` apr archive natively on mini (no docker on macOS); `release` waits for both lanes and requires only the Linux `build`.
- `.github/actionlint.yaml`: labels `macOS`, `apple-silicon`, `m4`, `mini`.

## Evidence

Native on mini, main @ 3786870d9, 2026-09-13T10:11Z: `cargo test -p aprender-compute --lib -j6` rc=0; `cargo check -p aprender-core --lib` rc=0. Guards on the result: check_runner_labels, check_no_hosted_runners, check_workflow_env_defined, check_workflow_path_filters, check_hardcoded_paths, check_no_timing_in_required, check_perf_concurrency_groups all PASS; actionlint finding count unchanged from main (2 pre-existing SC2016 in the docker lanes).

Sibling: paiml/infra PR flips `machines/mini/README.md` from "cowork box, CI is a reluctant guest" to the operator's ruling and adds `mini` to the declared label set.

Closes #3204

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3204, opened 2026-09-13
mini (Apple M4) is a full-time aprender build host: macOS/arm64 CI leg + runner re-registration (operator 2026-09-13)


noahgift:
Operator ruling 2026-09-13 (verbatim): "always watching gx10, yoga, mini (you need this to be fulltime aprender build slave as well) optimization".

Measured 2026-09-13T10:05Z: the `mini-m4` runner registration had been auto-deleted by GitHub on 2026-09-12 (`Registration was not found or is not medium trust` … `Runner registrations are automatically deleted for runners that have not connected to the service recently`), and NO aprender workflow selects a macOS runner — every job names `Linux` or a `clean-room` docker pool, so mini could take nothing even when online.

Done at 10:07Z: re-registered via the infra recipe (`make -C machines/mini runner-register` shape) with labels `self-hosted,macOS,ARM64,apple-silicon,m4,mini` (opt-in box label `mini`, per the one-box rule) — online, idle.

This ticket: a `macos-arm64` job in ci.yml on mini (not a required check: one runner, 16 GB), the actionlint label allow-list, and the nightly aarch64-apple-darwin target that nightly.yml says "comes back when a box for them does". paiml/infra README policy flip is a sibling PR.

Pmat-Ticket: PMAT-1098

noahgift:
Milestone 0.70.0 → 0.70.1 (gate triage). 0.70.0 now holds only the operator's release gate: "FULL ontology spec and all aprender binary working" + "0.70 to crates.io" (ONT-001, every [[bin]] build+smoke, the publish cascade), plus P0s and operator-ruled rows. This issue isn't needed for any of those. Evidence: evidence/m070-gate-triage-1835.md. If that's wrong, move it back to 0.70.0 and say which gate it serves.

PR #3211, opened 2026-09-13
evidence(perf-060): land the stranded intel + mini c=1 calibration — T-5 called the branch dead, it holds the only measurement of what mini can do

noahgift:
no-close: this lands stranded measurement evidence; the PERF-060 work item it belongs to
is not resolved by publishing the calibration, and #2706 stays open.

## Why this is opening now, two weeks after it was measured

`scripts/check_reconcile.sh v0.67.0 --prev-tag v0.66.0` (APR-RELEASE-001 §6, T-5) reports
**R3 dead-branches = 2**, and names `feat/y2-calibrate` and `feat/y1-7bgarbage` — remote
branches with no open PR and a tip older than 14 days.

Checked before archiving either, and **neither is dead**: both carry unique unmerged
work. This one carries `evidence/perf-060/` — 791 insertions of calibration measured
2026-08-30 against `34248e8fe`, which exists nowhere else. Deleting it to make R3 zero
would have destroyed measurements to satisfy a counter.

So R3's predicate cannot distinguish **dead** from **stranded**, and the honest fix for a
stranded branch is to land it, not to delete it.

## What it measures

| host | prefill tok/s | decode tok/s | `t_req` |
|---|---|---|---|
| `intel` (Xeon W-3245, CPU) | 8.14 – 8.97 | 6.57 – 7.66 | 73.9 – 81.9 s |
| `mini` (M4, CPU) | 2.371 – 2.372 | 2.318 | 270.2 – 271.1 s |

And the finding that mattered: the runbook's estimates in
`docs/PERF-054-campaign-scope.md` §5 assumed intel 40/6 and mini 60/8 tok/s. Decode was
close on intel (pessimistic by ~1.2×); **prefill was wrong by ~4.8× on intel**. The error
is concentrated in prefill, which is where a c=1 campaign spends its time.

## Why it is worth more now than when it was written

mini became a **full-time aprender build host** on 2026-09-13 (operator ruling, #3204,
#3205). This is the only measurement of what mini can actually do — 2.37 prefill /
2.32 decode tok/s, roughly **3.4× slower than intel** on CPU inference — and it is the
evidence behind giving mini single-crate build work rather than inference lanes.

## Scope

Evidence only. **791 insertions, every one under `evidence/perf-060/`.** The file says so
itself in its first line: *"Calibration only. No cell was measured.
`scripts/perf-matrix.yaml` was not modified. No full band set was run."* No contract, no
gate, no threshold changes.

`feat/y1-7bgarbage` is the other branch R3 named. It carries a contract, an example and a
script, so it needs a real review rather than a publish, and is reported in the T-5
receipt instead of being swept in here.

### DIFF
--- a/evidence/perf-060/calibration.md
+++ b/evidence/perf-060/calibration.md
@@ -1,0 +1,587 @@
+# PERF-060 — c=1 calibration of `intel` and `mini`, the two non-CUDA hosts
+
+**Calibration only. No cell was measured. `scripts/perf-matrix.yaml` was not modified.
+No full band set was run.**
+
+Measured 2026-08-30 against commit-under-test **`34248e8fecf1e354f3426cd077d4c5b0536d6cc7`**
+(`origin/main` at the time of the run). Refs paiml/aprender#2706, PERF-060.
+
+---
+
+## 0. Answers to the three questions that were asked
+
+**1. The measured rates.** Per host, W1 shape (513-token prompt, 128 generated),
+c=1, `apr` serving the pinned 7B GGUF. Every replicate is listed in §7; these are
+the ranges, not medians.
+
+| host | prefill tok/s | decode tok/s | `t_req` |
+|---|---|---|---|
+| `intel` (Xeon W-3245, CPU) | **8.14 – 8.97** | **6.57 – 7.66** | **73.9 – 81.9 s** |
+| `mini` (M4, CPU — see §3) | **2.371 – 2.372** | **2.318** | **270.2 – 271.1 s** |
+
+**2. The runbook's estimates do not hold, and the error is concentrated in prefill.**
+`docs/PERF-054-campaign-scope.md` §5 (branch `feat/x3-campaign-plan`) assumed `intel` 40 / 6 and `mini`
+60 / 8 tok/s. Decode was close on `intel` (estimate 6, measured 6.6–7.7 — the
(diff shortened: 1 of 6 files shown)


noahgift:
`pr-review-sign` went RED here with **exit 127**, and it is worth naming because it is a
property of opening a PR from an old branch, not of this change:

```
bash: scripts/install_pr_review_tools.sh: No such file or directory
##[error]Process completed with exit code 127.
```

That job checks out `github.event.pull_request.head.ref` — the PR's **head tree** — while
running **main's** workflow file. `feat/y2-calibrate`'s tip is 2026-08-30 and predates the
script main's workflow calls, so main's steps ran against a tree that does not contain
them. Same shape as "a release event runs the workflow file AT THE TAG", inverted: here
the workflow is new and the tree is old.

`gh pr update-branch` fixes it — the branch now carries the script and the job is green.

Swept the other 23 open PRs for the same trap: **none is affected**, because every one of
them has had main merged in. This PR was the only one opened straight from a stale branch
without that step, which is exactly the case T5-F2 produced (a branch nobody had touched
in two weeks).

Also, for the record: the packer refused a roadmap union here
(`base=873 head=656 … deleted=217`). That was staleness, not content — this branch does
not touch `docs/roadmaps/roadmap.yaml` at all (`git diff --name-only
origin/main...origin/feat/y2-calibrate | grep -c roadmap` → 0). The update-branch cleared
that too.

PR #3214, opened 2026-09-13
fix(ci): the never-install policy had a detector nothing called — book.yml compiled bashrs 7.4.0 against a 7.0.1 pin, on every book build

### LINKED ISSUE #3213: check_tool_versions.sh --audit-workflow has no caller: book.yml and coverage-nightly.yml have been violating the never-install policy the whole time (bashrs 7.4.0 vs the 7.0.1 pin)
## The detector exists and nothing calls it

`ci.yml` has asserted **"pmat/bashrs must match the pinned fleet versions (never install)"**
since BSE-10a / PMAT-1066, and `scripts/check_tool_versions.sh` ships
`--audit-workflow FILE` to catch a workflow that installs instead of asserting.

```
basis=$ grep -rn "audit-workflow" .github/ scripts/
scripts/check_tool_versions.sh:30:#   bash scripts/check_tool_versions.sh --audit-workflow FILE  # scan a workflow
scripts/check_tool_versions.sh:208:if [ "${1:-}" = "--audit-workflow" ]; then
scripts/check_tool_versions.sh:210:        printf 'usage: %s --audit-workflow FILE\n' "$0" >&2
scripts/check_tool_versions.sh:218:    printf 'usage: %s [--self-test|--audit-workflow FILE]\n' "$0" >&2
```

Four hits, all of them the script's own usage strings. **No caller, ever.** Aimed at the
real tree for the first time it immediately found two violators:

| workflow | lines |
|---|---|
| `book.yml` | 2 — `cargo install bashrs --locked`, then `cargo install pmat … \|\| true` |
| `coverage-nightly.yml` | 1 — `command -v pmat \|\| cargo install pmat … \|\| true` |

## What it cost, measured

`book.yml` cold-compiled **bashrs from source on every book build**. It died twice on
2026-09-13:

```
error: could not parse/generate dep info at: /tmp/cargo-installlBzi1T/release/deps/bashrs-72081467d6f8a380.d
  No such file or directory (os error 2)
error: failed to compile `bashrs v7.4.0`
```

Same class as the cargo-audit cold build that evicted the merge queue. And the version is
the second defect: **7.4.0 installed, 7.0.1 pinned**. The book's shell gate and the
release's shell gate have been linting with different linters — which
`check_bashrs_gate.sh`'s own header calls *"the defect this fixes"* — and bashrs is known
to cross-contaminate between versions.

The pmat sites are worse than slow. `cargo install pmat … || true` is precisely the
fail-open shape `check_tool_versions.sh` was written to delete: a failed install carries
on with whatever is already on PATH, so a bad network day silently downgrades the gate
without failing the job that depends on it. `coverage-nightly.yml` had two fail-opens on
one line.

## Fixed in #3214

- All three sites call `bash scripts/check_tool_versions.sh`, which fails **closed**.
- `audit_all_workflows` globs `.github/workflows/*.yml` — a universe, not a hand-kept
  list — and runs in the guard's **bare** path, which `guard_tree.sh` already invokes. No
  ci.yml edit needed to arm it. 17 audited, 0 installing.
- The detector counted **comments**: a step explaining why it no longer installs stayed
  RED. It now skips lines whose first non-space character is `#`.
- Case table 8 → 14 rows with both polarities, mutation-verified.

noahgift:
Closes #3213

`ci.yml` has asserted **"pmat/bashrs must match the pinned fleet versions (never install)"**
since BSE-10a/PMAT-1066, and `check_tool_versions.sh` ships `--audit-workflow FILE` to
catch a workflow that installs instead. `grep -rn audit-workflow .github/ scripts/` finds
**only that script's own usage strings** — nothing has ever called it.

Aimed at the tree for the first time, it found two violators immediately.

## What it cost, measured

`book.yml` cold-compiled bashrs **from source on every book build**, and died twice on
2026-09-13:

```
error: could not parse/generate dep info at: /tmp/cargo-installlBzi1T/release/deps/bashrs-72081467d6f8a380.d
  No such file or directory (os error 2)
error: failed to compile `bashrs v7.4.0`
```

Same class as the cargo-audit cold build that evicted the merge queue. And the version is
the second defect: **7.4.0 installed against a 7.0.1 pin** — the book's shell gate and the
release's shell gate have been linting with *different linters*, which
`check_bashrs_gate.sh`'s own header calls "the defect this fixes", and bashrs is known to
cross-contaminate between versions.

The pmat sites are worse than slow: `cargo install pmat … || true` is exactly the
fail-open shape this guard exists to delete. `coverage-nightly.yml` had **two** fail-opens
on one line.

All three now call `bash scripts/check_tool_versions.sh`, which fails **closed**.

## The sweep, so it cannot come back

`audit_all_workflows` globs `.github/workflows/*.yml` — a universe, not a hand-kept list —
and runs in the guard's **bare** path, which `guard_tree.sh` already invokes. **No ci.yml
edit is needed to arm it.** 17 workflows audited, 0 installing.

## The detector could not tell prose from code

`grep -cE 'cargo install (pmat|bashrs)'` counted the **comment** explaining why a step no
longer installs — so a step documenting the fix stayed RED after the install was gone. The
wrong remedy is to reword the comment; a detector that cannot tell prose from code teaches
people to hide the prose. It now skips lines whose first non-space character is `#`.

Case table **8 → 14 rows**, both polarities, because this regex shipped with none:

| row | verdict |
|---|---|
| a real `cargo install bashrs` | RED |
| a COMMENT naming `cargo install pmat` | GREEN |
| `cargo install --path crates/apr-cli` | GREEN (our own binary) |
| `cargo install cargo-llvm-cov` | GREEN (not a pinned tool) |
| `check_tool_versions.sh \|\| true` | RED (fail-open invocation) |
| a sweep over zero workflows | RED (vacuity) |

Mutation-verified: reverting the comment skip turns row 10 RED; deleting the vacuity block
turns row 14 RED.

## One row did not work when it was written

Row 14 "passed" on `command not found`: `audit_all_workflows` was defined at line 255,
**below** the self-test block at 146, so the row never called the function it was testing.
A green row proving nothing, in the PR about detectors nobody calls. The definition moved
above the self-test, and only then did its mutant die.

## Verification

```
scripts/check_tool_versions.sh --self-test   SELF-TEST PASSED (14/14 rows)
scripts/check_tool_versions.sh (bare)        17 workflow(s) audited, 0 still installing
scripts/check_runner_labels.sh               every self-hosted job pins a discriminating label
scripts/check_no_hosted_runners.sh           OK: 17 workflow files, no GitHub-hosted image
actionlint book.yml coverage-nightly.yml     2 SC2016 info — identical count on main (pre-existing)
```


noahgift:
## Second commit: the same class, one level down

While verifying this PR's version pins I ran `check_shell_lint_ratchet.sh --update` and watched it **delete** the `# tool_version=` header from `shell_lint_baseline.txt`:

```diff
-# tool_version=bashrs 7.0.1
-8
+6
```

Then measured what notices. Nothing does:

| state of the file | `check_tool_versions.sh` says | verdict |
|---|---|---|
| header present | `3 baseline header(s) ... all agree with tools.toml` | PASS |
| header removed | `2 baseline header(s) ... all agree with tools.toml` | **PASS** |

A third of the universe vanished and the audit reported agreement over the remainder — the same shape as this PR's original finding (a detector that ran on a shrinking set without saying so).

**The harm is live.** `shell_lint_baseline.txt` records bashrs 7.0.1; this box runs 7.3.0. The ratchet prints `Improved: 8 -> 6. Run --update to record it.` — but the 8 came from 7.0.1 and the 6 from 7.3.0, and nothing in the repo can separate progress from instrument drift (bashrs counts have moved 13 → 53 on one tree across two releases). **So I did not record it.** It is not known to be an improvement.

Fixed here:
- `--update` stamps the **live** version, never the pinned one — re-baselining on a drifted box now lands a header that disagrees with `tools.toml`, and this guard says so. Stamping the pin would launder a number measured by the wrong instrument into one that looks correctly measured.
- `check_headers` gains a floor that is not a magic number: every baseline must declare its instrument, in either spelling the repo uses (`# tool_version=`, 16 files; or the stricter `count`/`pmat_version`/`basis` schema of `hardcoded_path_shipped_baseline.txt`). The summary now prints the denominator — `3 of 17` — so a shrinking universe is legible.

Three rows, mutation-verified, each killed by exactly its own row: drop the no-instrument check → row 9; drop the `pmat_version:` spelling → row 10; drop the zero-baselines arm → row 11. Self-test **17/17**.

`baseline_require_tool_version()` still has no caller outside its own test — that stays open as #3217, because wiring it can only be sequenced once the fleet pins and the headers agree.

PR #3245, opened 2026-09-14
fix(cbtop): a machine with no /proc was rendered as a machine with no RAM

### LINKED ISSUE #3236: cbtop renders 0 bytes of memory on macOS: read_meminfo's ENOENT is unwrap_or_default()'d into zeros
## `cbtop` reports 0 bytes of memory on macOS, silently

```
thread 'budget_and_collectors::f043_memory_collector_valid_metrics' panicked at
crates/aprender-cbtop/tests/falsification/budget_and_collectors.rs:101:5:
F043 FALSIFIED: Total memory is 0
```

Measured on mini-m4 (Apple M4, 16 GB) while sweeping the integration targets on darwin.

## The fail-open

```rust
pub fn collect(&mut self) -> MemoryMetrics {
    let metrics = self.read_meminfo().unwrap_or_default();   // <- swallows the error
    ...
}

fn read_meminfo(&self) -> Result<MemoryMetrics, std::io::Error> {
    let content = read_to_string("/proc/meminfo")?;          // <- ENOENT on darwin
```

There is no `/proc` on macOS, so `read_meminfo` returns `Err(ENOENT)` and `unwrap_or_default()` turns that into an **all-zero `MemoryMetrics`**. `cbtop` then renders `0` total, `0` available, `0` swap — presented exactly like a measurement, with nothing anywhere saying the read failed.

Reporting "0 bytes of memory" is worse than reporting nothing: a monitoring tool that cannot measure should say so. This is the same shape as the VRAM ledger's `is_alive` (#3205): an *unmeasurable* condition rendered as a definite, wrong value. There the default released live reservations; here it reports a machine with no RAM.

Note `crates/apr-cli/src/commands/cbtop_get_cpu_memory.rs` gets this right — it is `#[cfg(target_os = "linux")]` and falls back to a stated constant. The collector in `aprender-cbtop` does not.

## And the test that catches it is dark

```
$ grep -c "aprender-cbtop" .github/workflows/ci.yml
0
```

`workspace-test` runs `--workspace --lib`, so the crate's unit tests run; `tests/falsification/` is an integration target and is not in the explicit `--test` list. So F043 has never executed in CI on any platform — it only fires where `/proc` is absent, and no CI host lacks `/proc`.

## Two ways to fix, and they are not equivalent

1. **Make the failure visible.** `collect()` should not paper an I/O error into zeros — either propagate, or carry an explicit "unavailable" state the UI can render.
2. **Read memory portably.** darwin has `sysctl hw.memsize`; that makes `cbtop` actually work on macOS rather than merely honest about not working.

(1) is the correctness fix and stands on its own; (2) is a feature. I am filing rather than picking, because which one is wanted depends on whether `cbtop` is meant to support macOS at all — and `apr cbtop` is a shipped subcommand, so that is a product call.

Either way the falsification target should join CI's explicit `--test` list, or this stays invisible.

Found by: the darwin integration sweep behind #3205. Related dark-target finding: #3235.

### LINKED ISSUE #3239: 1,424 of 1,469 integration targets run only when a PR touches their own crate — four of tonight's defects trace to this

```
```

| found | actually was |
|---|---|

## How it was measured

```
```

```
```

## What is worth deciding

noahgift:
```
F043 FALSIFIED: Total memory is 0
```

Measured on **mini-m4** (Apple M4, 16 GB) while sweeping the integration targets on
darwin. The collector read `/proc/meminfo` unconditionally and swallowed the error:

```rust
let metrics = self.read_meminfo().unwrap_or_default();
```

There is no `/proc` on macOS, so that read is `Err(ENOENT)`, `unwrap_or_default()`
turned it into an all-zero `MemoryMetrics`, and `cbtop` rendered `0` total / `0`
available / `0` swap — presented exactly like a measurement, with nothing anywhere
saying the read had failed.

Reporting "0 bytes of memory" is worse than reporting nothing. It is the same shape as
the VRAM ledger's `is_alive` (#3205): an *unmeasurable* condition rendered as a
definite, wrong value. There the default released live reservations; here it reports a
machine with no RAM. The old `Default` was indistinguishable from a reading, so no
caller could have told the difference even if it wanted to.

## Two fixes, and they are not the same fix

1. **The fail-open.** `MemoryMetrics` carries `measured`, false on every failure path,
   and `collect()` uses `unwrap_or_else(MemoryMetrics::unavailable)` rather than a
   `Default` that looks like data. "0 because this host has no swap" and "0 because
   nothing was read" are now different values.

2. **The blindness.** darwin gets a real reader — `sysctl hw.memsize` + `hw.pagesize`,
   `vm_stat`, `sysctl vm.swapusage` — so `apr cbtop` measures the box instead of being
   honest about not measuring it. mini is a full-time aprender build host; a monitor
   that cannot see it is not much of a monitor. An unsupported platform still returns
   `Err` and renders unavailable: no invented constant.

The issue asked which of these was wanted and called it a product call. `apr cbtop` is
a shipped subcommand and mini is now a first-class host, so it is both — (1) is
correct unconditionally, (2) is what makes (1) not merely honest.

## The parsers are pure, which is the only reason the darwin rows ever run

No CI host lacks `/proc`, so a darwin-gated test is dark by construction.
`parse_vm_stat` and `parse_swapusage` take text, so their rows execute on Linux CI
against real output captured from mini over SSH:

```
sysctl -n hw.memsize   17179869184
sysctl -n hw.pagesize  16384
vm_stat                Pages free: 505274.  inactive: 143073.  speculative: 18831.
vm.swapusage           total = 1024.00M  used = 29.69M  free = 994.31M  (encrypted)
```

Eight unit rows; each mutation turns exactly its own red:

| mutation | rows red |
|---|---|
| `measured = total_kb > 0` → `= true` | `a_meminfo_without_memtotal_is_not_a_measurement` |
| `Default { measured: true }` | `unavailable_is_distinguishable_from_a_measurement` |
| page size hardcoded to 4096 | `vm_stat_pages_become_kb_…`, `the_page_size_is_load_bearing` |
| `available = free` (drop inactive+speculative) | `vm_stat_pages_become_kb_…` |

The page-size row earns its place: arm64 macOS uses 16 KiB pages and x86_64 uses 4 KiB,
so assuming 4096 under-reports an M-series box by **4×** — a wrong number rather than a
missing one, which is the whole complaint here.

**Deleted**: a unit test that asserted the *defect* — it required `total_kb == 0`
wherever `/proc/meminfo` was absent, encoding the zero as expected behaviour.

## Verified on the box

A `/tmp` worktree on mini (macOS 26.6.2, arm64), removed afterwards:

```
cargo test -p aprender-cbtop --lib bricks::collectors::memory    8 passed
cargo test -p aprender-cbtop --test falsification f043           1 passed
```

F043 passes on darwin for the right reason — a real reading — rather than being relaxed
to accommodate a zero. On this box: 8 unit + 36 falsification pass, `cargo fmt` clean,
`cargo clippy -D warnings` clean.

## And the target was dark

`workspace-test` is `--workspace --lib`, so `tests/` never ran, and `ci.yml`'s explicit
`--test` chain — the only thing that reaches an integration target — named no
`aprender-cbtop` target at all. **F043 has never executed in CI on any platform.** Added
to that chain, so the next regression is caught by CI rather than by hand on a laptop
months later. Part of #3239.

Not fixed here, filed as a note in the commit: `crates/apr-cli/src/commands/cbtop_get_cpu_memory.rs`
returns a hardcoded `64` on non-Linux, which reports 64 GB for mini's 16. Same class,
different file.

Closes #3236

no-close: #3239 stays OPEN. This wires ONE dark target of the 1,424 it counts; the
tiering defect it describes is untouched. #3205 is referenced only as the prior
instance of the same fail-open shape, not as work this PR does.


noahgift:

noahgift:

PR #3246, opened 2026-09-14
ci(guard): a sibling crate can arm a test target gated on a runner capability

### LINKED ISSUE #3242: A dev-dependency's feature arms a sibling's runner-gated test: the BSE-17 quick tier ran three browser falsifiers on a browser-less runner
## What happened

`workspace-test` failed on #3205, run [34803859272](https://github.com/paiml/aprender/actions/runs/34803859272), job on `gx10-pool1`,
step **"Quick tier: lib + integration tests of the selected crates (BSE-17)"**:

```
Summary [532.637s] 30284/39431 tests run: 30281 passed (5 slow), 3 failed, 75 skipped

aprender-test-lib::falsify_chromium_driver_is_real a_screenshot_is_real_png_pixels
aprender-test-lib::falsify_chromium_driver_is_real js_is_evaluated_by_a_real_engine
aprender-test-lib::falsify_chromium_driver_is_real navigation_and_metrics_reflect_the_live_page

  could not launch a real browser: Browser not found. Install Chromium or set CHROMIUM_PATH
```

Those three tests are gated `#![cfg(feature = "browser")]`, and `browser` is **not**
a default feature of `aprender-test-lib` (`default = ["tui", "media", "watch"]`).
The test file's own header says they should never have run there:

> These require Chrome/Chromium on PATH. They are NOT on ci.yml's beat list, because
> the clean-room image is not known to ship a browser; they belong on a Chrome-equipped
> runner the way the GPU falsifiers belong on a CUDA one. Running them without a
> browser FAILS — deliberately. There is no skip.

## Mechanism (measured, not inferred)

`crates/aprender-orchestrate/Cargo.toml`, `[dev-dependencies]` line 10:

```toml
jugar-probar = { path = "../aprender-test-lib", package = "aprender-test-lib", features = ["browser"] }
```

The BSE-17 quick tier runs **one** nextest invocation over the whole selection:

```
cargo nextest run --profile ci --lib --tests -p apr-cli -p aprender-cgp -p aprender-compute \
  -p aprender-mcp -p aprender-orchestrate -p aprender-test-lib -p aprender-train -p aprender-verify-ml
```

`--tests` builds dev-dependencies. Selecting `aprender-orchestrate` therefore enables
`aprender-test-lib/browser` — and because that is the same package instance in the same
build graph, `aprender-test-lib`'s **own** browser-gated integration target compiles and
runs too.

It needs both halves of the selection, which is why this had never fired before: a PR
must touch `aprender-test-lib` (so its test targets are built at all) **and** a crate
whose dev-dependency asks for `browser`. #3205 is the first to do both.

## Why this is a class, not an instance

`scripts/ci_test_tier.sh`'s header states the contract:

> Feature-gated suites (model-tests, setfit, …) belong to the full tier only; **the quick
> tier runs default features.**

That claim is false whenever a selected crate's **dev-dependency** enables a non-default
feature of another selected crate. Feature unification is a property of the *build graph*;
"default features" is a property of a *package*. The tier selects packages and then asserts
a package-level property over a graph-level fact.

`cargo tree` will not show this from the library side — the edge is in `[dev-dependencies]`,
so it only exists when something is built with `--tests`.

## Fix applied in #3205 (it cannot merge without it)

`browser-falsify = ["browser"]` in `aprender-test-lib`, and the target gates on the new
name. A cargo feature cannot express "this box has a browser"; a name a sibling cannot
reach by wanting `chromiumoxide` can. **No skip is introduced** — where `browser-falsify`
is on, the target still hard-fails without a browser.

## Open, and deliberately not done here

1. **Nothing runs these three tests anywhere.** No runner in the fleet has Chromium, and
   `.github/workflows/ci.yml` names no browser. The target has only ever been capable of
   failing. Putting Chromium on the clean-room image and arming a job is the real close-out
   — measure first, assert second (`tools.toml`'s own ordering rule).
2. **The general guard.** A `check_*.sh` that reads every workspace `[dev-dependencies]`
   edge and fails when one enables a feature that gates an integration target of the same
   workspace crate. That is the mechanical version of this ticket and would have caught it
   before #3205 did.
3. `crates/aprender-test-cli` has `default = ["browser"]` — worth checking whether the same
   trap exists there.

noahgift:
`crates/aprender-test-lib/tests/falsify_chromium_driver_is_real.rs` opens with
`#![cfg(feature = "browser")]`, and its own header states the containment:

> These require Chrome/Chromium on PATH. They are NOT on ci.yml's beat list, because
> the clean-room image is not known to ship a browser; they belong on a Chrome-equipped
> runner the way the GPU falsifiers belong on a CUDA one. Running them without a
> browser FAILS — deliberately. There is no skip.

A cargo feature could not hold that. `crates/aprender-orchestrate`'s `[dev-dependencies]`
line 10:

```toml
jugar-probar = { path = "../aprender-test-lib", package = "aprender-test-lib", features = ["browser"] }
```

The BSE-17 quick tier runs **one** nextest invocation with `--lib --tests` over the whole
selection, and `--tests` builds dev-dependencies. A PR touching both crates unified
`browser` on, compiled `aprender-test-lib`'s **own** gated target, and ran three browser
tests on a browser-less clean-room runner — 30,281 passed, 3 failed, on the operator's
priority PR (#3205, run 34803859272).

Feature unification is a property of the **build graph**. "Default features" is a property
of a **package**. `ci_test_tier.sh`'s header asserts the second over the first.

## Measured, not suspected

**33 such edges today, over 22 `(crate, feature)` gates.** The enabler reaches the feature
two different ways, and covering only one misses half the surface:

| spelling | example |
|---|---|
| `features = ["browser"]` on the dependency | `aprender-orchestrate` `[dev]` |
| a feature table entry `probar/browser` | `aprender-test-cli` `[normal]` |

`optional = true` does not exempt an edge either — it means "off until a feature turns it
on", which is exactly how `aprender-cgp` reaches `aprender-gpu/cuda`.

## A ratchet, not a hard fail

Most of the 33 are probably harmless: a crate the quick tier never selects, or a gate whose
capability every runner has. Failing on all 33 would be an assertion the tree has not
earned. So the baseline is the measured number, it only ever shrinks, and a **new** edge is
what fails — named:

```
NEW since the baseline:
  aprender-distribute        [dev   ] -> aprender-test-lib/gpu    via features=[] on the dependency
```

That is a real run: adding `"gpu"` to `aprender-distribute`'s dev-dep, then reverting it.

## The case table

Six rows over committed fixture manifests, no cargo. **R3 is the discrimination row** —
identical manifests, and the only difference is that the feature gates no test target, so
there is nothing to arm and nothing to report. **R4** holds that a crate arming its *own*
target is not a leak. **R5** holds that an optional dependency still counts. **E1** makes an
unreadable input ENV rather than an empty (and therefore passing) answer.

## Two placement decisions

- **The jq program is its own file** (`scripts/lib/feature_leak_edges.jq`). Inline, bashrs
  lints the guard as shell and reads jq's `$r[0]` / `$deps[]` as unbraced array expansions:
  nine `SC1087` errors on a program that is not shell. Silencing a false positive by
  rewriting the code is worse than the finding; removing it by putting the jq where jq
  belongs is just correct. `bashrs`: **0 errors**.
- **Wired in `guard-cargo`, not guard-tree.** `ci.yml` runs `guard_tree.sh --no-cargo`, so a
  cargo-reading guard placed there would never be dispatched — the class this repo calls "a
  facility with a self-test and no caller". `cargo metadata --no-deps` means no build.

Vacuity floor: fewer than 5 feature-gated test targets found means the **scan** is broken,
not that the tree is clean — and a broken scan reports zero edges, which reads exactly like
a pass.

Verified: `--self-test` 6/6; `check_guards_are_wired.sh` PASS (baseline did not grow);
`check_baseline_ratchets.sh` classifies the new baseline as `set`; the ratchet fires and
names a synthetic new edge.

Refs #3242

no-close: #3242 stays OPEN. This is the guard it asks for, but the ticket also owns the
un-run browser falsifiers (no runner in the fleet has Chromium) and the 33 existing edges,
none of which this PR examines.


noahgift:

No activity since 2026-09-17 13:42Z; **2 release trains have been cut since** (v0.68.2, v0.68.1). Age is the only input to this sweep — it is not a judgement on the work, and a train label such as `pp-066/carry-*` is not an exemption.

noahgift:

noahgift:

PR #3248, opened 2026-09-14
fix(apr-format): the byte-identity golden was asserting a format we deliberately left

### LINKED ISSUE #3235: apr-format golden_v2 F32 writer emits 516 bytes against a 1092-byte golden — and the target is dark in CI
## A byte-identical golden test has drifted by 2×, and nothing runs it

```
thread 'golden_v2_f32_writer_is_byte_identical' panicked at
crates/apr-format/tests/golden_fixtures.rs:119:5:
assertion `left == right` failed: v2 F32 byte length drifted from golden_v2.apr
  left: 516
 right: 1092
```

The v2 F32 writer emits **516** bytes; `tests/fixtures/golden_v2.apr` is **1092**. That is not a rounding difference — the written artifact is roughly half the golden.

## It is not a platform issue, and it is not the fixture

I found it while sweeping the integration targets on darwin, so the first thing to rule out was the platform:

```
linux   (lambda-vector)  test result: FAILED. 3 passed; 1 failed
darwin  (mini-m4)        test result: FAILED. 3 passed; 1 failed
```

Identical. And the fixture is byte-for-byte the same size on both checkouts (`1092`), so it is not a checkout/line-ending artifact either. The writer and the golden genuinely disagree, everywhere.

## Why nobody noticed

```
$ grep -c "golden_fixtures" .github/workflows/*.yml
0
```

`apr-format`'s `golden_fixtures` target is **dark**. CI runs `--workspace --lib` plus one explicit list of `--test` targets, and this one is not on that list — so a golden-file guard for the on-disk format has never executed in CI. Its three sibling rows in the same file pass; only the F32 length row is red.

This is the same class as the already-recorded "integration targets never run" and "benches are dark targets" findings, but it is worse in kind: a *golden* test exists precisely to catch silent format drift, and a golden test nothing runs is the one guard whose absence is invisible by construction.

## What needs deciding, and by whom

Either the writer regressed (and 516 bytes is a real format defect shipping today) or the golden is stale (and was never regenerated after a deliberate format change). Those have opposite fixes, and the answer is in the v2 F32 writer's history rather than in this test — so I am filing rather than guessing. Whichever way it resolves, the target should join CI's explicit `--test` list so it cannot go dark again.

Found by: widening the mini darwin leg to `--tests` (#3205). The leg's `--lib` surface is green — 81,689 tests, 73 binaries, 0 failures — so this only surfaces once the integration targets run at all.

### LINKED ISSUE #3239: 1,424 of 1,469 integration targets run only when a PR touches their own crate — four of tonight's defects trace to this

noahgift:
```
assertion `left == right` failed: v2 F32 byte length drifted from golden_v2.apr
  left: 516    right: 1092
```

Not a writer regression, and not a stale fixture to regenerate either. Measured by
parsing both artifacts rather than reasoning about them:

```
produced metadata  189 bytes,  9 keys
golden   metadata  715 bytes, 35 keys
ONLY IN GOLDEN: 26 keys — architecture, author, chat_format, hidden_size,
                num_layers, rope_theta, vocab_size, …   every one null
ONLY IN PRODUCED: none
DIFFERENT VALUES: none
```

Every key the two share holds an identical value. `git log -S skip_serializing_if` names
the change: **#2254**, `fix(finetune): apr finetune --merge produces a directly-runnable
.apr`, which added `#[serde(default, skip_serializing_if = "Option::is_none")]` to
`AprV2Metadata`. `golden_v2.apr` was captured in **#2236**, before it. The writer is
correct, the change was deliberate, and 516 bytes carries exactly what 1092 did.

## Regenerating `golden_v2.apr` would have been the easy fix and the wrong one

That file is the **pre-extraction oracle** — bytes written by the format code while it
still lived in `aprender-core` — and `golden_v2_loads_in_leaf` exists to prove the
extracted leaf still *reads* them. It passes today (3 passed, 1 failed: only the
byte-identity row was red). Overwrite it and that test proves nothing but that the leaf
reads its own output.

So the oracle **stays at 1092 bytes** and byte-identity gets a fixture of its own,
`golden_v2_current.apr` (516). Both properties are kept: old artifacts still parse, and
today's writer is still byte-pinned against silent drift.

A second golden is a licence to smuggle, so it does not stand alone.
`the_two_v2_goldens_carry_the_same_model` parses both and proves the 516-byte fixture is
the same model as the oracle — metadata values, tensor names, shapes, F32 payload.

## The failure now says which kind it is

A byte diff has two causes with opposite fixes. Reporting "bytes differ" makes every
reader re-derive that, which is how this sat for two months. Both branches proved by
mutation:

| mutation | what the test says |
|---|---|
| fixture := the 1092 oracle | `SERIALIZATION drift … both carry the same model … regenerate golden_v2_current.apr … golden_v2.apr must NOT be regenerated` |
| one weight `8.0` → `8.5` | `CONTENT drift … do NOT carry the same model. This is a writer defect, not a spelling change.` |
| one direction of `fields_agree` deleted | `same_v2_model_discriminates_a_real_difference` red |

The first reproduces #3235's original report verbatim and answers it.

`same_v2_model` is what the split rests on, so it gets rows of its own rather than only
what the committed pair happens to exercise — they differ **only** by null-spelled
fields, which drives one of its two symmetric directions. The new row drives both, plus a
real value difference and a real tensor difference.

Decomposed into `fields_agree` / `metadata_agrees` / `tensors_agree` /
`one_tensor_agrees` because the pre-commit gate refused the first draft at Cyclomatic 30
/ Cognitive 25 — correctly.

Verified: `clippy -D warnings` clean; apr-format 94 lib + 6 golden + 2 sovereignty pass.

Not fixed here: the target is still dark. `grep -c golden_fixtures .github/workflows/*.yml`
is **0**, which is why this ran for two months after #2254 without failing. The `--test`
chain is edited by #3245 this cycle and only one PR may hold that line, so the wiring goes
there.

Closes #3235

no-close: #3239 stays OPEN — this is one dark target of the 1,424 it counts, and the
tiering defect it describes is untouched.


noahgift:

noahgift:

PR #3250, opened 2026-09-14
fix(present-yaml): 34 example tests read a path nothing has lived at since APR-MONO

### LINKED ISSUE #3239: 1,424 of 1,469 integration targets run only when a PR touches their own crate — four of tonight's defects trace to this

noahgift:
```
ALD-002 should be valid: "Failed to read ../../examples/ald/data_table_virtualized.yaml:
No such file or directory (os error 2)"
```

Every row in `tests/examples.rs` (27) and `tests/prs_examples.rs` (7) failed, all on one cause.
The examples are in the **sibling** crate, `crates/aprender-present/examples/{ald,apr,charts,dashboards,edge_cases,prs}/`.

| file | what it did |
|---|---|
| `examples.rs` | `Path::new("../../examples/…")` — two levels up from `crates/aprender-present-yaml/` is the repository **root**, and CWD is whatever the runner chose. Both wrong. |
| `prs_examples.rs` | already anchored on `CARGO_MANIFEST_DIR`, then walked `.parent().parent().join("examples").join("prs")` — to the same repository root. The CWD-independence was right and the **destination** was wrong, which is why it reads as deliberate. |

`CARGO_MANIFEST_DIR` + `../aprender-present/examples` rather than a corrected `../`: it resolves
at **compile** time against the crate being built, so no working directory can break it and it
survives the next move. One helper per file, 27 + 3 call sites.

Proved: pointing the helper back at `../../examples` turns all 27 rows red again; restoring it
turns them green. `aprender-present-yaml` now **205 lib + 27 + 17 + 7** pass, `clippy -D warnings`
clean.

## How this was found — the more useful half

Nothing in this repo runs integration targets: `ci.yml`'s `workspace-test` is `--workspace --lib`
plus one explicit `--test` list, and `make coverage` — the only other full run — is **also
`--lib`**. So I ran the whole surface once:

```
cargo nextest run --workspace --lib --tests --no-fail-fast
  (GPU crates excluded, exactly as the full tier excludes them)

Summary [1141.178s] 98621 tests run: 98584 passed, 37 failed, 329 skipped
```

19 minutes, and ~16,000 tests across **654 binaries** that CI has never linked. The 37 failures
are five binaries, not a swamp:

```
aprender-present-yaml::examples            27   ← this PR
aprender-present-yaml::prs_examples         7   ← this PR
apr-format::golden_fixtures                 1   ← #3248
aprender-mcp::falsify_mcp_008               1   ← #3238
aprender-cgp::falsify doctor_speed_real     1   ← open
```

The cadence decision that follows from it is #3249.

## Not wired here

`ci.yml`'s `--test` chain is held by #3245 and #3248 this cycle and only one PR may edit that
line; a third would conflict with both. The wiring for these two targets goes in after they land
— and the wiring must land **with** a fix, never before it, or main goes red the moment it merges.

Refs #3239

no-close: #3239 stays OPEN — this fixes 34 of its 37 measured failures; the tiering decision it
describes is #3249's.


noahgift:

PR #3253, opened 2026-09-14
fix(cgp): six wall-clock assertions, three of them in the REQUIRED check

### LINKED ISSUE #3239: 1,424 of 1,469 integration targets run only when a PR touches their own crate — four of tonight's defects trace to this

noahgift:
```
FALSIFY-CGP-061 FAILED: doctor took 2665ms
```

Measured during a full `--lib --tests` sweep — the box was running **98,621 tests in parallel**.
2665 against a 2500 ms limit is a **6.6% miss** that says nothing about `cgp doctor` and everything
about the load. It is the **last of the 37** failures that sweep found, and the one blocking
#3249's hard-failing nightly.

Raising the limit is the "clever ratio rewrite" that once blocked all nine open PRs. The fix is to
stop asserting duration where duration cannot be measured.

## Fixed one, found six

The test that failed was the entry point, not the scope:

| site | limit | |
|---|---|---|
| `tests/falsify.rs:457` | doctor < 2500 ms | ← the one that failed |
| `tests/falsify.rs:583` | profile < 31 s | same file, simply not unlucky yet |
| `tests/integration.rs:39` | doctor < 30 s | |
| `src/doctor.rs:406` | checks < 2 s | **in `src/`** |
| `src/analysis/diff.rs:344` | < 100 ms | **in `src/`** |
| `src/metrics/export.rs:70` | < 100 ms | **in `src/`** |

Those last three are `#[cfg(test)]` units, so they run in `--workspace --lib` — **the required
`workspace-test`, on every PR and every merge group.** A 100 ms assertion on a shared runner is a
coin flip, and two of them have been sitting in a required check.

`scripts/check_no_timing_in_required.sh` exists for exactly this; its header says *"policy is what
failed the other eleven times."* But its registry covers `check_*.sh` **guards**. It has never been
able to see a Rust target, so nobody has ever counted these.

## `timing-gate`

Off by default, enabled by nothing — the same registry idea, for Rust. Arm it where duration is
measurable: a dedicated unloaded runner at release time. Not in CI's default set, and not in
#3249's nightly full surface.

**Gated, not deleted** — proved by mutation rather than asserted:

```
limit → 0, WITH --features timing-gate    FAILED (1 failed)   the assertion is live
limit → 0, WITHOUT the feature            ok                  compiled out
restored                                  ok
```

## A second defect in the same row

```rust
assert!(stdout.contains("RTX 4090") || stdout.contains("GPU"));
```

That asserts the **runner's hardware** — the class fixed in #3210 across four sibling rows — and
passes here only because this box happens to have a 4090. What CGP-061 owes is that `doctor`
reached a *verdict* about the device and printed it, so that is what it now asserts: non-empty
output, and a decision among GPU / CUDA / device / no GPU / not available.

Verified: `aprender-cgp` 121 lib + 3 falsify pass, `clippy -D warnings` clean.

## Not fixed here

The same regex over the workspace finds **~70 duration assertions in `crates/*/src/`** and 36 more
in `crates/*/tests/`, across 12 crates — `aprender-serve` 9, `aprender-core` 5, `aprender-test-lib`
3, and so on. This PR fixes the one crate that actually failed; the sweep is its own ticket.

Refs #3239

no-close: #3239 stays OPEN — this clears the last of its 37 measured failures, but the tiering
decision it describes is #3249's.


noahgift:

PR #3255, opened 2026-09-14
ci: run the test surface nothing else runs — hard-failing, no baseline

### LINKED ISSUE #3239: 1,424 of 1,469 integration targets run only when a PR touches their own crate — four of tonight's defects trace to this

noahgift:
> **DRAFT until its dependencies land.** This job hard-fails, so it must not merge before the
> five PRs that take the failing set to zero: **#3250** (34 rows), **#3248**, **#3238**, **#3253**
> (the last one), and **#3241** (which adds `scripts/cargo_step.sh`, used in step 1). Merged
> ahead of them it turns a nightly red on day one — the exact outcome the design argues against.

Implements the cadence in #3249.

## What it runs, and why nothing else does

`--tests` appears **exactly once** in every other workflow — `ci.yml`'s BSE-17 quick tier, scoped
to the crates a PR touched. `make coverage`, the only other whole-workspace run, is `--lib`.
`nightly.yml` runs no tests at all.

```
merge-queue full tier   82,217 tests /  73 binaries   (job 103868350604)
this job                98,621 tests / 727 binaries   (1141 s run, measured)
```

**654 test binaries have never been linked in CI**, let alone run. A binary that never links
cannot fail, cannot flake, and cannot appear in a report.

## No baseline. Hard fail on day one.

The first draft proposed a shrink-only ratchet. A 4-lane agy quorum refused it, and the objection
is mechanical rather than stylistic: **`--update` writes a file and needs a commit, and a scheduled
workflow has no PR and no author.** Every other ratchet here is maintained by a human inside a PR
*because it runs in one*. A cron's baseline would only ever be edited by someone who already
noticed the job was red — the exact population a ratchet exists to stop relying on.

So the 37 failures are fixed first and this ships with nothing to maintain.

## Not on gx10 — and the mechanism is measured

The build produces a **39 GB** target directory (808 binaries). gx10 is the smallest of the three
Linux hosts — 916 GB against intel's 3.6 TB and yoga's 935 GB — and it hit **zero bytes free** on
2026-09-14, dequeuing a PR from merge-queue position 1 (#3247).

`X64` is the exclusion: over 25 runs, **no job requesting X64 has ever landed on gx10** (it is
ARM64). No new label.

## Four refusals, each earned by a defect already paid for here

| step | why |
|---|---|
| `cargo_step.sh --self-test` **first** | a sweep of 727 binaries is exactly where an unclassified cargo death reads as 650 new defects |
| preflight ≥ 120 GB free | fail `ENV:` **before** the build, not half-way through — that is how #3247 presented |
| `--no-fail-fast` | plain `cargo test` stops at the first failing **binary**; seven rounds went into four tests |
| vacuity floor, 600 binaries | a selection that built nothing reports zero failures, which reads exactly like a pass |

It reclaims the 39 GB in-job under `always()`, because the host reaper's critical mode already
cannot reclaim under `keep=3` (#3247) — leaking there compounds a known-unfixed defect.

## The timeout is honest about being a guess

`timeout-minutes: 90`, deliberately generous and **not derived**. BSE-05 wants
`T = max(15, ceil(1.5·p99), p99+20)` over the job's own history and this job has none. The only
reading in hand is 21 min on a 48-core box with a warm sccache; a clean-room runner measured
**2.34× slower per test** on the same workload, extrapolating to ~51 min — an extrapolation, not a
measurement. Re-derive from three real runs and tighten.

## Also carries the selector-touch trigger

`ci_test_tier.sh`, `gate_touched_crates.sh`, `tree_reader_tests.txt`, `ci.yml`. A change to how the
tier chooses what to run is exactly when the dark set moves.

Verified: yaml parses; `check_no_hosted_runners`, `check_runner_labels`,
`check_perf_concurrency_groups`, `check_ci_reusable_workflow_pinned`, `check_no_timing_in_required`,
`check_cargo_install_private_root`, `check_hardcoded_paths`, `check_guards_are_wired` and
`check_tool_versions --audit-workflow` all pass on the new file.

Refs #3239

no-close: #3239 stays OPEN until this is undrafted and has run green three times — a workflow that
has never fired is a claim, not a gate.


noahgift:

noahgift:

PR #3260, opened 2026-09-14
ci(deep): the T-1 gate APR-RELEASE-001 requires and nobody wrote

### LINKED ISSUE #3239: 1,424 of 1,469 integration targets run only when a PR touches their own crate — four of tonight's defects trace to this

noahgift:
> **DRAFT until #3257 lands.** Clause 2 (`--no-default-features`) is red on main today — 43 errors,
> all in `aprender-distribute` (#3176). #3257 fixes it. Merged ahead of that, this gate is born red.

APR-RELEASE-001 §4 T-1: *"`ci / deep` green on cut sha: full tests, doctests, `--no-default-features`,
feature matrix, GPU, every `cargo run --example`"*.

**`ci / deep` has zero hits in every workflow here — and in the SHA-pinned upstream
`paiml/.github` `sovereign-ci.yml`.** The gate the release train asserts on does not exist. Its
named clauses run nowhere either:

```
cargo test --doc        0 hits in any workflow
cargo run --example     0 hits
feature matrix          0 hits
--no-default-features   1 hit — a Windows BUILD flag in nightly.yml
```

The 0.67.0 train satisfied T-1 by running pieces of this **by hand** from the autopilot script and
writing the numbers into the ledger. Nothing recorded them as a check, and nothing would have
stopped the train had they been skipped. That is not a criticism of the train — it is what a
missing gate looks like from the inside.

## The three clauses that are here

| clause | measured 2026-09-14 | cost |
|---|---|---|
| doctests | 978 pass, 0 fail, 428 ignored | 192 s |
| `--no-default-features` | 0 errors (**43** before #3257) | 26 s |
| examples **build** | 980 targets, 0 errors | 83 s |

**Five minutes for three clauses that have never run.** This is a wiring job rather than a fixing
job — and that is only true because #3257 landed the 43-error half an hour ago.

## What is deliberately not here

| | |
|---|---|
| full tests | #3255 owns it (98,621 tests / 727 binaries, ~21 min) |
| GPU | `cuda-nightly.yml` owns it |
| feature matrix | **undefined in the spec** — a clause with no definition cannot have an implementation |
| every `cargo run --example` | **written, measured, removed — see below** |

## The examples clause: I wrote it, measured it, and took it back out

The challenge was fair — slowness is what a nightly is *for*, and `.config/nextest.toml` already
answers **hanging** with `slow-timeout = { period = "60s", terminate-after = 20 }` so *"a genuine
hang dies at 20 minutes with a name"*. `check_examples_run` even sidesteps stdin with
`cargo run --example X -- --help`.

So I implemented it. Then ran it on a random sample of five:

```
HANG     aprender-core::grid_search_tuning
HANG     aprender-serve::imp_700_realworld_verification
exit101  aprender-serve::debug_q4k_gemv
ok       aprender-simulate::reproducibility
HANG     aprender-serve::bench_barrier_overhead
```

**Three of five ran past 60 s on `--help`.** These are not CLIs — they are compute demos that
ignore argv and do the work. `--help` does not make them exit, and *"hang"* is the wrong word:
they are **running**. A per-example timeout would report a working example as a failure, at 980
rows. (Five is a small sample; the ratio is not the claim, the *kind* is.)

Two shapes that could work are written into the workflow header for whoever decides: redefine as
*"starts and does not crash within N seconds"* with timeout treated as **pass** (~2 h at N=10 — so
#3255's full-surface lane, not this 5-minute one), or have examples **declare** runnability.

Clause 3 — every example **builds** — already catches the common rot, which is compile breakage.
That is the part that is cheap and unambiguous.

Separately filed as #3261: `check_examples_run`'s only callers are its own tests, and it fails
open — a non-zero runtime exit counts as a pass, so a crash passes.

## On the name — a real trade, not an oversight

A check called `ci / deep` can only come from a job named `deep` inside the reusable workflow this
repo calls as `ci` — an **org-wide** change to `paiml/.github`, blast radius across every consuming
repo. This job is named `deep` and emits the check `deep`.

Amending the **spec's string** is the cheap half of that trade; amending an org-wide reusable
workflow to match a string is the expensive half. **APR-RELEASE-001 §4 should read `deep`**, and
that amendment belongs with whoever owns the spec.

## Two properties it refuses to give up

- **Daily, not only on dispatch.** A gate that runs only at release time is discovered broken at
  the worst possible moment.
- **Vacuity floor** on the example count (800, under the measured 980). Each clause can pass by
  measuring nothing — a doctest run that compiled no crate, an examples build that found no
  example.

And every cargo invocation goes through `cargo_step.sh` (#3241): a T-1 lane that reports a full
disk as *"the release is not green"* is how a train gets held for a runner fault.

Verified: yaml parses; `check_no_hosted_runners`, `check_runner_labels`,
`check_perf_concurrency_groups`, `check_no_timing_in_required`, `check_guards_are_wired` and
`check_tool_versions --audit-workflow` all pass.

Refs #3239

no-close: #3239 stays OPEN — this implements three of T-1's clauses; the tiering decision it
describes is #3249's, and the two clauses above still need definitions.


noahgift:
## Measured: this lane cannot be exercised before it is merged

```
$ gh workflow run deep.yml --ref PMAT-1098-ci-deep-lane
HTTP 404: workflow deep.yml not found on the default branch
```

`workflow_dispatch` is only honoured for workflows that exist on the **default branch**,
and `deep.yml` has no `pull_request` trigger (deliberately — it is a deep lane). That
combination has a consequence worth stating before the next train, not after:

**`ci / deep` has never run, and cannot run until it is on `main`. Its first execution
would be the release cut it gates.**

§4 T-1 is `"ci / deep green on cut sha"`, and §4's rule for a red step is SKIPPED, no
partial promotion. So a lane born red does not fail loudly — it silently costs the train,
and 0.68.0 would fall back to 0.67.0's hand-run for the second consecutive release.

That is this repo's recurring shape: a gate whose first run is the thing it certifies.
The fix is sequencing, not code —

1. land #3277 (the queue is deadlocked on two PRs that each fix the other's red check),
2. land #3257 — clause 2 (`--no-default-features`) is the 43 errors in `aprender-distribute`,
3. land this,
4. **`gh workflow run deep.yml --ref main` and prove it green on `main`**,
5. only then treat T-1 as a gate rather than a ledger line.

Step 4 costs ~5 minutes (doctests 192 s + `--no-default-features` 26 s + examples build
83 s, per this PR's own table) and is the step that makes the difference between a gate
and a receipt.

Also still open from this PR's own description: **APR-RELEASE-001 §4 says `ci / deep`,
this job emits `deep`.** The spec amendment is the cheap half of that trade and should
land with the lane, or T-1's already-done test reads a check name that never appears.

Branch is 4 commits behind `main`; it will need an update before it can merge.

noahgift:

noahgift:

Issue #3179, opened 2026-09-12
aprender-zram-core's cuda feature has never compiled: unresolved trueno_gpu imports in library code, dark since before v0.66.0


noahgift:
## What

`aprender-zram-core`'s `cuda` feature does not compile. `crates/aprender-zram-core/examples/gpu_info.rs` declares `required-features = ["cuda"]`, so building it builds the library with `cuda`, and that fails:

```
error[E0432]: unresolved import `trueno_gpu::driver::PinnedBuffer`
   --> crates/aprender-zram-core/src/gpu/batch/mod.rs:241:82
241 |  use trueno_gpu::driver::{CudaContext, CudaModule, CudaStream, GpuBuffer, PinnedBuffer};
    |                                                                          ^^^^^^^^^^^^ no `PinnedBuffer` in `driver`

error[E0432]: unresolved imports `trueno_gpu::kernels::lz4::Lz4DecompressKernel`,
                                 `trueno_gpu::kernels::lz4::Lz4WarpShuffleKernel`
   --> crates/aprender-zram-core/src/gpu/batch/mod.rs:242:40
    |                                     no `Lz4WarpShuffleKernel` in `kernels::lz4`
```

The failing code is in the **library**, not the example. The example is only how it was reached.

## Not new

`crates/aprender-zram-core/src/gpu/batch/mod.rs` names `PinnedBuffer` five times at `v0.66.0` and five times at the 0.67.0 cut — unchanged. `v0.66.0` shipped with a `cuda` feature that does not build.

It stayed dark because nothing builds this crate with `cuda`: CI's `--examples` build does not enable it, and `workspace-test` excludes the GPU crates. The G3.EX sweep (APR-RELEASE-001 §4 T-2, #3121/#3122) builds each example *with its own `required-features`*, which is the first thing in this repository that ever asked for `aprender-zram-core/cuda`.

## Two defects, not one

1. The imports name items that `trueno_gpu` does not export (or no longer exports). Someone has to decide whether the intended API is `PinnedBuffer`/`Lz4WarpShuffleKernel` under different paths, or whether this code was written against an API that never landed.
2. A feature that does not compile is not a feature. Whatever the resolution, `cargo check -p aprender-zram-core --features cuda` belongs in the `ci / deep` lane (APR-RELEASE-001 §5 P1) so that "it builds" is measured on every tag rather than discovered by a sweep.

## Done

[...]

PR #3263, opened 2026-09-14
fix(features): 67 of the 100 red (crate, feature) pairs — 9 crates whose declarations never compiled

### LINKED ISSUE #3262: The (crate, feature) matrix has never been built: 100 of 430 selections do not compile, all of them unreachable from any default set
## Measurement

A full sweep of every `(crate, feature)` pair in the workspace — 430 pairs across 45 crates, each built as `cargo check -p <crate> --no-default-features --features <one feature>` — finished with **100 pairs RED**.

None of them are reachable from any crate's default set. That is the whole finding: a feature is declared in `Cargo.toml`, code is gated on it, and nothing in CI ever selects it, so the declaration and the code drifted apart with no gate to notice. `cargo check --workspace` stays green throughout because feature unification hands every crate whatever its siblings happened to enable.

Broken pairs by crate, as measured:

| crate | pairs | root causes |
|---|---:|---|
| aprender-train | 31 | `tui` declared in Cargo.toml, never applied in code; `pre_warm_realizador_gemm` re-export gated more loosely than its definition |
| aprender-distribute | 24 | `tokio` used unconditionally; **and** `num_cpus`, which only the `cpu` feature pulls |
| apr-cli | 19 | 8 sites where a definition is `#[cfg(feature = "inference")]` and its caller is not |
| aprender-orchestrate | 7 | `apr_serve` gated `native` but needs `realizar`+`libc`; rag TUI gated on the dep-feature, not the composite |
| aprender-simulate | 5 | CB-081 removed axum/jsonschema/z3 and left the gated code behind |
| aprender-serve | 2 | `cuda` supplies neither `gpu`, `server` nor `cli`, all of which its code uses |
| aprender-data | 2 | `cli` needs `local` + `shuffle` and declares neither |
| aprender-core | 1 | `showcase-profile` needs renacer, a DEV-dependency here (circular) |
| aprender | 1 | `wgpu` did not enable `entrenar/gpu` |
| aprender-zram-core / aprender-zram | 2 | `trueno_gpu::driver::PinnedBuffer` and the async pinned-copy API do not exist post-APR-MONO (#3179) |
| aprender-viz | 1 | `gpu-wasm` — `shared_instance`/`gpu_backends` gated `all(gpu, not(wasm32))` while their caller is not |
| aprender-test-cli | 1 | `ChatRequest` initializer missing `stream_options` |
| aprender-present-lib | 1 | `wgpu::RenderPassColorAttachment` initializer missing `depth_slice` |
| aprender-profile-core | 1 | `otlp = []` — an empty feature whose code needs `opentelemetry` |
| aprender-mcp | 1 | `pmcp-dispatcher` needs `native` |
| aprender-rag-cli | 1 | `EvalAction` missing the `Clone`/`Debug` its parent enum requires |

## Why it matters now

The night lane's feature-matrix job sweeps these 430 pairs. Born against this tree it would be red on 100 of them on its first run — a gate that is red before it has ever been green teaches people to ignore it.

## Remaining after PR #3263

PR #3263 fixes 9 crates. Four things are deliberately **not** in it:

1. **aprender-distribute (24 pairs)** — owned by open PR #3257, which gates on `feature = "tokio"`. That is necessary but not sufficient: `executor/cpu.rs` also uses `num_cpus`, which only the `cpu` feature pulls, so `checkpoint`, `microvm`, `remote`, `remote-tls`, `serverless`, `tokio` and `tui` stay red after it. Reported on that PR rather than duplicated here.
2. **aprender-zram-core / aprender-zram `cuda`** — #3179. `PinnedBuffer` plus `GpuBuffer::copy_{from,to}_pinned_async` are an entire host-pinned-memory API that does not exist in the in-tree `aprender-gpu`. Restoring it or rewriting the batch path on pageable memory is a driver-API decision, not a feature-declaration fix.
3. **aprender-present-lib `gpu`** — `wgpu::RenderPassColorAttachment` gained a `depth_slice` field; the initializer was never updated because nothing builds it. Mechanical, but it is a wgpu-version question, not a gate.
4. **aprender-test-cli `llm`** — `ChatRequest` gained `stream_options`; same shape.

Two features are declared but **cannot** be built by construction, and PR #3263 makes them say so with one `compile_error!` naming the exact edit that would enable them, instead of a cascade of unresolved imports:

- `aprender-simulate/z3-proofs` — CB-081 removed `z3`; unlike axum and jsonschema it is not already in `Cargo.lock` and needs the libz3 system library, so re-adding it is a real cost and a real decision.
- `aprender-core/showcase-profile` — `renacer` is a dev-dependency because renacer depends on aprender. Breaking that cycle is a prerequisite.

## Reproduce

```bash
cargo metadata --no-deps --format-version 1 \
  | python3 -c 'import json,sys;[print(p["name"],f) for p in json.load(sys.stdin)["packages"] for f in p["features"] if f!="default"]' \
  | while read -r c f; do
      cargo check -q -p "$c" --no-default-features --features "$f" 2>&1 | grep -qE 'error(\[|:)' && echo "$c $f"
    done
```

no-close: this is the umbrella measurement; it stays open until the four items above are green. PR #3263 closes none of it on its own.

noahgift:
## What

A full sweep of all 430 `(crate, feature)` pairs in the workspace found **100 red** — every one of them unreachable from any crate's default set, which is exactly why they were dark. Measurement and the full table: #3262.

This PR fixes **9 crates / 67 of those pairs**. Every one is the same class: a feature is declared, code is gated on it, and the declaration does not supply what that code needs. `cargo check --workspace` never noticed because feature unification hands each crate whatever its siblings enabled.

| crate | what was wrong |
|---|---|
| aprender-orchestrate | `apr_serve` gated `native` but calls `super::realizar` and `libc::prctl`, both from `inference`. Re-gated; its call site gets the cfg twin `build_fallback_driver` already had. Separately, the rag TUI gated on the implicit `presentar-terminal` dep-feature, which links presentar but not the `crossterm` it also uses — re-pointed at the `tui` composite. |
| apr-cli | `--no-default-features` — documented in CLAUDE.md as how you build without inference — had **never compiled**: 8 errors in 7 files, all "definition gated, caller not". Also `wgpu = ["inference"]` enabled neither entrenar nor `entrenar/gpu`, so `--features wgpu` and `--all-features` failed too. |
| aprender-train | `tui = ["dep:presentar-terminal"]` was declared in Cargo.toml and **never applied in code** — `grep -rn 'feature = "tui"' --include='*.rs'` returned nothing. 17 pairs on that one cause. Fixing it exposed a second: `pre_warm_realizador_gemm` re-exported under `realizar` while defined under `all(realizar, cuda)`. |
| aprender-simulate | CB-081 removed axum/jsonschema/z3 for Cargo.lock bloat and left the gated code. axum 0.7.9 and jsonschema 0.28.3 are **already in the lock**, so re-adding them costs zero new packages (verified: the lock diff is two edges, no new `[[package]]`). Plus three `contract_pre_iterator!` sites passing scalars to a slice precondition. |
| aprender-serve | `cuda` supplied none of `gpu`, `server`, `cli` — all of which the CUDA path uses. Only the default set, which happens to carry all three, ever built that code. |
| aprender-data | `cli` needs `local` (registry) and `shuffle` (`cmd_mix`); declared neither. |
| aprender-core | `showcase-profile` needs `renacer`, a DEV-dependency here because renacer depends on aprender. |
| aprender-compute / aprender-viz | `gpu-wasm` was broken on **every** target including wasm32, the one it exists for: `shared_instance`/`gpu_backends` were `all(gpu, not(wasm32))` while their caller carries no cfg. Also added the `Histogram::dimensions` setter — the fields existed, were defaulted, and were read by `to_framebuffer`; only the setter was missing, while every sibling widget has one. |
| aprender-mcp / aprender-profile-core / aprender-rag-cli | `pmcp-dispatcher` needs `native`; `otlp = []` was an empty feature whose code needs `opentelemetry` (its sibling aprender-profile declares it correctly — `-core` copied the name, not the dependency); `EvalAction` lacked the `Clone`/`Debug` its parent enum derives.

## Two features that cannot be built

`aprender-simulate/z3-proofs` and `aprender-core/showcase-profile` are declared but structurally unbuildable — z3 is not in the lock and needs libz3; renacer is circular. Each now emits **one** `compile_error!` naming the dependency and the exact edit that would enable it, with a private `__z3-linked` / `__renacer-linked` feature carrying the real gate so the refusal is the only diagnostic and aprender-core's existing stub still compiles.

## Complexity debt paid down

Touching `apr-cli/src/lib.rs` and two other files put five PRE-EXISTING violations in front of the pre-commit complexity gate, none of them in code this PR wrote. Paid down in the same commits, as the gate requires — all behaviour-preserving, each following a split the files already document:

- `dispatch_runtime_commands` 41 → under (the `apr run` arm extracted whole)
- `dispatch_diagnostic_commands` 30 → under (`apr trace` arm + save-tensor pre-dispatch extracted; the diff arm's doubled `and_then` nesting factored out)
- `help_producer_truth::resolve` 73 → under (token loop → a `Walk` state struct, one method per token kind)
- `explain_tensor` 28 → under (the `--json` half extracted)
- `oracle::rag::tui::run_loop` 32 → under (`sync_size` + `handle_key`)

Both extracted dispatch arms re-match behind a `let ... else` that refuses by name; neither panics.

## Verified

Per-selection `cargo check` on every pair this PR touches, plus each crate's default and the suites:

- apr-cli `--lib` 7242/7242 (including the three resolver behavioural tests covering the rewritten `resolve`); bare, `--all-features`, and 11 selections clean
- aprender-train `--lib` 7624/7624; 24 selections clean
- aprender-orchestrate `--lib` 6535/6535
- aprender-compute `--lib` 3511/3511; aprender-simulate `--lib` 3677/3677; aprender-viz + aprender-rag-cli `--lib` 605/605
- `Cargo.lock` gains zero new packages

## Not in this PR

- **aprender-distribute (24 pairs)** — owned by open PR #3257. Its `feature = "tokio"` gate is necessary but not sufficient: `executor/cpu.rs` also uses `num_cpus`, which only the `cpu` feature pulls, so 7 selections stay red after it. Reported there rather than duplicated here.
- **aprender-zram-core / aprender-zram `cuda`** — #3179. `PinnedBuffer` and the async pinned-copy API do not exist post-APR-MONO; restoring them is a driver-API decision.
- **aprender-present-lib `gpu`** and **aprender-test-cli `llm`** — upstream struct drift (`depth_slice`, `stream_options`), tracked in #3262.

no-close: #3262 is the umbrella measurement and stays open until the four items above are green; #3257 and #3179 are owned elsewhere.


noahgift:
Triage from the release babysit session (aprender-ba). I updated this armed branch with main (head 0f85f43b6) so its 09-15 checks stop being stale; the re-run shows two CODE reds and one ENV red:

1. **guard-tree, §6 R-2 body guard (CODE, body):** `no-close: #3262` is a hyphen-prefixed closing keyword — GitHub reads it as a real `close #3262`. Replace with `keep-open: #3262 <reason>` and **push** (the guard reads the frozen event payload; a body edit alone never re-runs it).
2. **guard-tree, dogfood coverage G2.1 freshness (CODE, tree):** this branch changes `crates/aprender-rag-cli/src/lib.rs`, which is cited by `docs/audits/surface_audit.csv`, without re-running the audit for those rows. Re-run the audit for the rag-cli rows and commit the csv.
3. **cuda-unit (ENV):** `driver::cublas_tests::cta64_vs_cta32_vs_cublas_fp16` hit `CUDA_ERROR_OUT_OF_MEMORY` on **yoga-eph** (job 105944360104, 18:23Z, the 8 GB 4060 Laptop). The same job passed on yoga-eph for #3498 (18:20Z) and #3533 (17:07Z), so this is a transient VRAM squeeze, not this diff. Rerun once the run finishes (`gh run rerun --job` refused while sibling jobs were still queued).

Auto-merge stays armed; it merges on its own once 1–2 are pushed.

noahgift:
Correction to item 3 above: **cuda-unit is not env.** The rerun (attempt 2, job 105953179488, yoga-eph, 19:22Z) fails identically — `driver::cublas_tests::cta64_vs_cta32_vs_cublas_fp16` at `crates/aprender-gpu/src/driver/cublas_tests.rs:1046`, `CUDA_ERROR_OUT_OF_MEMORY` after the 2048-row case — with the GPU otherwise idle (2 MiB of 8188 MiB in use, no other compute process). The same job passed on the same runner for #3498 (18:20Z) and #3533 (17:07Z), so it is the diff: this branch's feature-pair changes make that test allocate past the 4060 Laptop's 8 GB. Either the test needs a VRAM floor (skip-with-reason below N GB, like the other gpu tests) or the enabled feature set changed what it allocates. Owner's call; two occurrences with an idle card is no longer an anecdote.

noahgift:

noahgift:

Issue #2519, opened 2026-08-16
Three train-* binaries report confident results without doing the work — one is published to crates.io


noahgift:
Found while triaging binary surfaces with zero integration tests. These are not untested — they are **untestable as specified**, because they return fabricated data and exit 0.

Each claim below is a command that was actually run.

## 1. `aprender-train-inspect` — DELETE (published to crates.io, cascade tier 13)

`inspect.rs:115` carries its own confession:

> *"For real implementation, would parse the actual file / Here we return simulated data based on file size"* → `generate_mock_tensors()`

**Proof:** `info` on **5 KB of `/dev/urandom`** exits **0** and reports `Architecture: llama, Hidden Dim 768, Layers 1, Tensors 9`. A real 1-tensor safetensors file gets the **same 9 tensors**. `convert.rs:83` is "create a placeholder output".

Worth its own note: the `N-05` hardening in `architecture.rs` ("not hardcoded 4096") is **defeated** — it derives hidden-dim honestly from tensors that `inspect_model` fabricates. The hardening was applied one layer above the lie.

This one is the priority, because it ships.

## 2. `aprender-train-bench` — DELETE

`sweep.rs:169 simulate_training` is a closed-form curve with a comment reading *"Temperature ~4.0 is optimal"*.

**Proof:** run with no model, no data, no config → exits 0, prints `Optimal: temperature = 4.00` — the baked-in constant. `strategies.rs:80` likewise returns a per-variant literal table (`Combined → 0.71/0.831`). A hyperparameter sweep that never trains. Zero reverse-deps.

## 3. `aprender-train-shell` — DELETE

`commands.rs:232` *"Simulate model fetching"*. `estimate_params` string-matches `"7b"` in the model **ID**; `hidden_dim: 4096` is hardcoded; distill returns `"Training started... (simulated)"`.

**Proof:** `fetch does-not-exist/totally-fake-7b` exits **0** with `✓ Fetched … Parameters: 7.0B, Layers: 32`. Zero reverse-deps.

## 4. `aprender-qa-certify` — delete the `[[bin]]`, KEEP the lib

The binary reads `docs/certifications/models.csv`; **that directory does not exist**, and `README.md` has no `CERTIFICATION_TABLE_START` marker. Run in the repo root → exit 1, `No such file or directory`. Named in no workflow, Makefile, or doc.

[...]

PR #3264, opened 2026-09-14
ci(night): the two T-1 clauses too slow for a lane anyone waits on — examples RUN and the 430-pair feature matrix

noahgift:
## What

The two APR-RELEASE-001 T-1 clauses that are too slow for a lane anyone waits on. Operator ruling 2026-09-14 — *"wouldn't they be best done pre-release and/or night?"* — yes.

#3260 (`deep.yml`) carries the three cheap T-1 clauses: doctests (192 s), `--no-default-features` (26 s), examples BUILD (83 s). These two are hours:

| job | scope | measured |
|---|---|---|
| `examples-run` | 980 example targets | ~2 h |
| `feature-matrix` | 430 `(crate, feature)` pairs over 45 crates | ~1–2 h |

Runs at 22:40 UTC, clear of every other nightly (the earliest is cuda at 01:30), plus `workflow_dispatch` for pre-release.

## Timeout is a PASS, and that is the whole design

Measured on a random sample of five examples with `-- --help` and a 60 s cap:

```
HANG     aprender-core::grid_search_tuning
HANG     aprender-serve::imp_700_realworld_verification
exit101  aprender-serve::debug_q4k_gemv
ok       aprender-simulate::reproducibility
HANG     aprender-serve::bench_barrier_overhead
```

Three of five ran past it. They are not CLIs — they are compute **demos** that ignore argv and do the work, so "hang" is the wrong word: they are RUNNING. Treating a timeout as failure would report a working example as broken, 980 times.

So the job asserts what is actually owed: **every example starts and does not crash**. That catches the #2519 class (an example that panics immediately) without asserting a duration — which `check_no_timing_in_required.sh` exists to forbid.

## Known red, dated and sourced

The first full sweep (#3262) measured **100 of 430 pairs red**, every one unreachable from any default set. #3263 fixes 67. The remaining 9 are owned elsewhere:

- seven `aprender-distribute` selections → #3257. Its `feature = "tokio"` gate is right but not sufficient: `executor/cpu.rs` also uses `num_cpus`, which only the `cpu` feature pulls. Verified against that PR's head — all seven fail on `num_cpus`, and the other 17 selections are green.
- two `cuda` selections → #3179. `PinnedBuffer` plus `GpuBuffer::copy_{from,to}_pinned_async` do not exist post-APR-MONO.

Those 9 are **reported, not fatal**, so this lane can be GREEN on its first run rather than born red and taught to be ignored. A listed pair that **passes** is FATAL — the list may only shrink, and it shrinks by deleting a line. The matcher ships a 7-row case table, including the trap that `aprender-zram-core cuda` must not match the `aprender-zram cuda` row.

## Falsifiers

- Both jobs carry a **vacuity floor** (800 examples, 400 pairs). A discovery that finds almost nothing reports zero failures, which reads exactly like a pass.
- The matrix routes cargo deaths through `cargo_classify.sh`, so a saturated runner is reported as ENV rather than as a broken feature declaration.
- Both jobs open with `cargo_step.sh --self-test`.
- `$?` is read directly after each command, never through a pipe.
- All five workflow guards pass: reusable-workflow-pinned, no-timing-in-required, workflow-cargo-packages, workflow-env-defined, workflow-path-filters.

## Draft until #3263 lands

Held as a draft deliberately: the known-red list is written against the tree **after** #3263. Merged before it, `feature-matrix` would be red on 67 pairs the list does not name. Undraft once #3263 is in.

no-close: this adds the lane; #3262 is the umbrella measurement and stays open until the last 9 pairs are green, and #3257/#3179 are owned elsewhere.


noahgift:
Sequencing note before this is undrafted.

The known-red list in `night.yml` is dated 2026-09-14 and correct **as of that tree**. Since then:

- the seven `aprender-distribute` selections are fixed by #3257 (the tokio gate) **plus** #3266 (dropping `num_cpus` for `std::thread::available_parallelism`, decided 3/3 by quorum). Both are needed; neither alone turns them green.
- the two `cuda` selections are addressed by #3267, which does **not** make them green — it replaces the unresolved-import cascade with one named `compile_error!`. They stay red, so they stay on the list, and the reason now lives in the code.

The workflow treats a listed pair that PASSES as FATAL, by design — the list may only shrink. That means whoever undrafts this must **re-derive the list against the tree at that moment**, not trust the dated one: if #3257 and #3266 have landed, the seven distribute rows must be deleted or the first run goes red on "known-red list is stale".

Re-derive with:

```bash
cargo metadata --no-deps --format-version 1 \
  | python3 -c 'import json,sys;[print(p["name"],f) for p in json.load(sys.stdin)["packages"] for f in p["features"] if f!="default"]' \
  | while read -r c f; do
      cargo check -q -p "$c" --no-default-features --features "$f" 2>&1 | grep -qE 'error(\[|:)' && echo "$c $f"
    done
```

noahgift:

No activity since 2026-09-16 08:24Z; **3 release trains have been cut since** (v0.68.2, v0.68.1, v0.68.0). Age is the only input to this sweep — it is not a judgement on the work, and a train label such as `pp-066/carry-*` is not an exemption.

noahgift:

PR #3271, opened 2026-09-14
feat(release): APR-RELEASE-001 §5 P0·Instrument — `make build-report` reads the 1092 ledger records nothing read

noahgift:
## What

`APR-RELEASE-001` §5 P0·Instrument's *Done* test is "≥ 20 records; `make build-report` runs on a clean checkout". Measured on `main`: **1092 ledger records are committed** under `docs/build-ledger/` in exactly the P0·Instrument schema — 1084 of them job records — and **nothing in the tree read one**. So the first half passes 54× over and the second half had no mechanism, which is why §1's p95, §1's max-PRs-per-train and the whole of §7's `gate:` line were still `[U]` while the data sat in git.

This is the sccache class: recorded thousands of times, never compared.

`scripts/build_report.sh` reads them in **0.27 s**, byte-identical across runs.

## What it says — the numbers that were `[U]` until now

| host_class | n | p50 total_s | p95 total_s | p50 queue_s | p95 queue_s |
|---|---:|---:|---:|---:|---:|
| gx10 | 509 | 366 | 2282 | 137 | 1399 |
| intel | 454 | 666 | 4994 | 34 | 1066 |
| yoga | 121 | 340 | 3518 | 11 | 832 |

```
gate:    p95 ci/gate 20.1 | max PRs/train 120 | queue p95 intel 1066 yoga 832 gx10 1399
```

Two things fall out of that table that no occupancy average could show. **gx10 carries more jobs than intel** (509 vs 454) and has the *worst* queue p95 (1399 s vs 1066 s) — the box the packing rule treats as spare capacity is the busiest by count and the slowest to start. And **intel is the slowest per job** (p95 4994 s).

## The correction §1 needs

The first run of the report falsified the spec's own equation. §1 computes `max PRs/train ≈ 3 × 72 h / p95(ci / gate)`. The required checks on `main` are `gate` **and** `workspace-test`, and a PR merges when the **slowest** one is green:

| required check | n | p95 total_s |
|---|---:|---:|
| **workspace-test** | 147 | **6451** ← binds |
| ci / gate | 163 | 1205 |
| gate | 158 | 255 |

So §1 as written answers **645** and the honest answer is **120** — a 5.4× overstatement, on the one number §8 uses to decide whether to stop cutting trains ("measured max PRs/train < 10 `[A]`"). The report computes the bound over the required-check set, names which check binds, and prints the ratio by which §1's own form overstates. The `gate:` line keeps its specified shape; only the number is honest.

[...]

Issue #3269, opened 2026-09-14
ONT-001: aprender owns 16 untriaged ontology rows — one per train, ratcheted (APR-RELEASE-001 §11)


noahgift:
## Why this is filed

`ONT-001 v4.3` (`infra/docs/specifications/paiml-ontology.md`) §6 assigns **aprender** every ontology row but three. Measured on `main` @ `fa6e35f23`, 2026-09-14: **1 of 17 rows merged**, and **1 ticket** in this repo (#3222, closed, no milestone). Sixteen rows of upstream-assigned work had no ticket, no milestone and no owner here — untriaged, which is P0 by standing rule. This epic is the triage, not a promise to finish by the due date (see *Horizon*).

## Baseline — derive, don't quote

| ONT-001 requires | aprender has | derive with |
|---|---|---|
| contracts may carry `entity:` / `shape:` / `evidence:` | **0 / 0 / 0** of 1818 | `grep -rlE '^entity:' contracts --include='*.yaml' \| wc -l` |
| `pv census` (ONT-1) | absent | `pv --help` |
| `pv extract` → `contracts.nt` (R-18) | absent | `pv --help` |
| `crates/aprender-contracts/src/ontology/` (decision 4) | does not exist | `ls` |
| 16 entity types named, 14 extractors (7 in v1alpha1 scope) | **0** | §3.7 |
| every ONT row implemented (§6) | 1 of 17 — ONT-2a (#3224) | `git log --grep='ONT-' origin/main` |
| `pv kaizen` **is** the kaizen loop | code-only: bindings, call sites, E0/E1/E2 | `crates/aprender-contracts-cli/src/commands/kaizen.rs` |

Two of these are load-bearing:

1. **`pv kaizen` is blind to every surface that is not Rust.** The release train sweeps the `(crate, feature)` matrix, examples, `README.md`, `CLAUDE.md`, workflows, model files and CSVs — and the loop that is supposed to improve on each sweep cannot see one of them. ONT-001 R-17 is exactly this: one structure, any entity type.
[...]

Issue #2512, opened 2026-08-16
check_contract_test_binding.sh resolves 371 test references and runs in no CI workflow


noahgift:
Found while verifying a related claim. Small, cheap to fix, same family as #2496 and #2500.

## The guard works

On a clean checkout of `main`:

```
$ bash scripts/check_contract_test_binding.sh ; echo rc=$?
Running pv lint contracts --strict-test-binding ...
Resolved 371 test references; 27 dangling across 13 contract(s).
PASS: no contract cites more nonexistent tests than its baseline allows.
rc=0
```

It resolves **371 contract→test references**, holds 27 dangling ones at a shrink-only baseline, and is not vacuous.

## It runs nowhere

```
$ grep -c check_contract_test_binding .github/workflows/*.yml Makefile
.github/workflows/*.yml : 0   (all twelve)
Makefile                : 2
```

Makefile-only means `make tier3`, and **tier3 is not run in CI**. So a contract citing a test that does not exist can be added freely; the only thing that would catch it runs on whoever remembers to type `make tier3`.

That is the same shape as the `bashrs` gate covering 7 of 72 scripts (#2511) and the cgp targets that died for two months unobserved (#2496): the check is written, correct, and unwired.

## Correcting a claim I was given

I was told this guard "currently exits 1 on clean `main` with `VACUOUS: strict-test-binding gate was SKIPPED (contract validation failed); nothing was measured`."

**That is not reproducible on a clean tree** — it exits 0 and measures 371 references, as above. The vacuous branch fires when contract validation fails, which is what happens on a branch carrying a malformed contract. So the guard's fail-closed behaviour is working as designed; it was the *tree* that was broken, not the guard.

Worth recording because the two diagnoses lead opposite ways: "the guard is broken" invites rewriting a correct guard, while the real defect is that nothing runs it.

## Fix

Add it to `guard-runner-labels` in `ci.yml` (which is in `gate.needs`, so it blocks) beside the other contract guards. It shells out to `pv lint contracts --strict-test-binding`, so the cost is one `pv` invocation.

Worth checking at the same time whether the other Makefile-only guards are in the same position — `grep -c` across `.github/workflows/` for each `scripts/check_*.sh` would enumerate them in one pass.

noahgift:
Swept it, as suggested above. **This guard is one of four, and two of the four run in nothing at all — not CI, not even the Makefile.**

[...]

PR #3272, opened 2026-09-14
fix(guard): the roadmap guard counted agy lane scratch as a second copy of the tree (1107 phantom claims)

noahgift:
## What

`make gate` went red on a clean branch with **1107 "NEW unprovable claims"**, every one pointing at an innocent roadmap row. The claims were real rows — read twice.

`.claude/worktrees/` is agy's throwaway lane workspace: a full checkout, `roadmap.yaml` included, created per review lane and left behind when a lane dies. It is in `.git/info/exclude`, so nothing under it can ever reach the tree CI gates. But this guard's universe **deliberately** includes untracked files — that is the hole it was written to close (#2512: an untracked file is exactly how a new uncited claim arrives) — so it walked straight into the lane checkouts and counted every entry twice.

The arithmetic says it plainly:

| | before | after |
|---|---:|---:|
| roadmap files in the universe | 46 | **23** |
| completed claims | 2266 | **1133** |
| NEW (unbaselined, uncited) | **1107** | **0** |

Exactly half. The lane worktree was duplicating the entire universe.

## Why it is a one-line class of fix that was already known

`scripts/check_complexity_ratchet.sh:149` already prunes `*/.claude/worktrees/*`. That lesson was paid for once and never propagated; this guard is the second place it was needed. Of the scripts that mention lane scratch at all, only one actually excluded it — the rest name it in prose.

The prune is applied in `is_candidate()` rather than only in the `os.walk`, so the `git ls-files` half of the universe is covered by the same rule; the walk additionally skips descending into what was measured here as **66 registered lane worktrees**.

## This widens no hole

A path under `.claude/worktrees/` is in `.git/info/exclude` and cannot be committed, so it can never be part of the tree this guard gates. Excluding scratch from a guard over the tree is not a weakening — scanning it was the defect, because a guard that reds on files CI cannot see is a guard nobody can act on.

## RED first

A fixture roadmap planted at `.claude/worktrees/lane-deadbeef/docs/roadmaps/roadmap.yaml` with a completed-and-uncited entry must contribute 0 records. Without the prune the guard's own case table reads:

```
  FAIL  want 0          got 1          lane scratch contributes no records
```

The row ships with the fix, in the guard's existing `case table: the universe` block.

## Found by

Verifying #3271 (APR-RELEASE-001 §5 P0·Instrument). A `make gate` run there failed on four checks; three were pre-existing or time-dependent and one was mine. A **second** gate run after fixing mine surfaced this one, which the first run had not reached — the review quorum I had dispatched in between left the lane checkout that the guard then walked into.

Separately measured and **not** acted on here: 66 registered worktrees under `.claude/worktrees`, 1 prunable, 0 with a missing directory. They are live directories that may belong to other sessions, so pruning them is not this PR's business.

no-close: this is a guard-universe defect found while verifying #3271; it has no issue of its own and arrival already outruns closure 2:1, so it files nothing new.

ont-delta: none — the §11.1 delta for the guard surface is a `code` contract over the universe predicate, and it waits on ONT-1 (`pv census`), epic #3269.


noahgift:

PR #3274, opened 2026-09-14
chore(toolchain): ALWAYS LATEST — bashrs 7.4.1, pmat 3.40.0, and the three baselines that move with them

noahgift:
## What

Operator rule: **"ALWAYS LATEST … we always go latest crates.io on our stack."**

This moves aprender's assertion to the versions the fleet now runs, and re-records the three baselines that name those instruments.

| | before | after |
|---|---|---|
| `tools.toml [bashrs]` | 7.0.1 | **7.4.1** |
| `tools.toml [pmat]` | 3.39.0 | **3.40.0** |
| `scripts/shell_lint_baseline.txt` | 6 | **6** (header only) |
| `scripts/complexity_baseline.txt` | 688 rows | **681 rows** |
| `scripts/cb200_baseline.txt` + `.pmat-gates.toml [tdg]` | 609 | **602** |

## Order: converge → prove → assert

This file's own header records why that order is not negotiable — it was once moved to 7.0.4 ahead of the fleet and a `guard-cargo` job on yoga-build2 answered `FAIL bashrs pinned 7.0.4 found 7.0.1` twenty minutes later.

So the boxes moved **first**. Measured after the swap, all four:

```
intel  bashrs=7.4.1 pmat=3.40.0
gx10   bashrs=7.4.1 pmat=3.40.0
yoga   bashrs=7.4.1 pmat=3.40.0
mini   bashrs=7.4.1 pmat=3.40.0
```

The declarations landed first too, in paiml/infra#588 (29 pins across 6 machines, plus a guard that asserts them against crates.io nightly). Each binary was built to `/tmp` and swapped in with a **same-directory rename**, never `cargo install --force` into the shared root — `~/.cargo/bin` is shared by ~16 runners under one user, and a non-atomic replace is what produced `could not execute process … (never executed)` on this fleet twice.

Verified before pushing: `scripts/check_tool_versions.sh` reports **4/4 checks, 0 failed** on this branch under bashrs 7.4.1 + pmat 3.40.0.

## The baselines were re-measured, not re-stamped

- **shell_lint 6 → 6.** bashrs 7.4.1 counts the same 6 error lines over all 203 scripts as 7.0.1 did, so only the provenance header moves.
- **complexity 688 → 681.** Clean shrink: 7 rows left, **0 arrived**.
- **cb200 609 → 602.** `pmat comply` said so itself: *"the tree is 7 under the recorded baseline: lower `[tdg] baseline` to 602 in .pmat-gates.toml to bank it."*

**Both pmat numbers fell by exactly 7 and the tree did not change between the two readings** — same commit, different instrument. This is a re-measurement banked at a new floor, **not seven functions anybody fixed**, and the comments say so where the numbers live.

Worth knowing which way it cuts: among the 7 that left the complexity set is `decompress_neon_impl` at cognitive **130**. That function is still there and still that shape; pmat 3.40.0 simply no longer reports it. The ratchet now watches 681 rows instead of 688, and the seven it stopped watching did not improve.

no-close: this is the aprender half of the ALWAYS LATEST convergence; paiml/infra#588 carries the fleet declarations and the nightly guard.

ont-delta: none — the §11.1 delta for the toolchain surface is a `code` contract over the version assertion, and it waits on ONT-1 (`pv census`), epic #3269.


noahgift:
Superseded by #3277, which lands this together with the fix it was deadlocked against.

#3274 fails `guard-cargo` on cargo-deny advisories — what #3276 fixes.
#3276 fails `guard-cargo` on the pin mismatch and `guard-tree` on the baseline instrument mismatch — what #3274 fixes.

Neither can go green alone, so neither could ever enter the merge queue. #3277 carries both.

PR #3276, opened 2026-09-14
fix(security): RUSTSEC-2026-0285 — rustls 0.23.40 -> 0.23.45 (red-lines every PR until it lands)

noahgift:
## Why this is urgent

`cargo-deny` fetches the RustSec advisory-db **live**, so a newly published advisory red-lines **every PR and `main`** the moment it lands. This one landed mid-afternoon 2026-09-14 and turned `guard-cargo` red on a PR whose entire diff is four version strings.

```
error[vulnerability]: TLS 1.3 handshake messages incorrectly accepted across encryption level boundaries
rustls 0.23.40 — RUSTSEC-2026-0285
Solution: Upgrade to >=0.23.45
```

## What the advisory is

Rustls accepted TLS 1.3 handshake messages sent at the **wrong encryption level** when they followed a key-changing message in the same record — e.g. a plaintext `EncryptedExtensions` packed into the same record as the `ServerHello` was accepted. RFC 8446 §5.1 requires terminating with an `unexpected_message` alert.

The handshake transcript is still authenticated, so this is **not** handshake forgery or alteration. The practical effect is that a peer can send in plaintext what should have been encrypted and rustls does not reject the connection. Functionally the same bug as Go's [GO-2026-4340](https://pkg.go.dev/vuln/GO-2026-4340) (CVE-2025-61730).

## The change

`cargo update -p rustls@0.23.40`. The `@0.23.40` spec is **required** — the tree carries two rustls majors and a bare `-p rustls` is ambiguous:

```
error: There are multiple `rustls` packages in your project, and the specification `rustls` is ambiguous.
  rustls@0.21.12
  rustls@0.23.40
```

Four packages moved:

| crate | from | to |
|---|---|---|
| rustls | 0.23.40 | **0.23.45** |
| rustls-webpki | 0.103.13 | 0.103.15 |
| aws-lc-rs | 1.17.0 | 1.18.1 |
| aws-lc-sys | 0.41.0 | 0.45.0 |

`rustls 0.21.12` also remains in the lock. It is **not** what this advisory names, cargo-deny does not flag it, and it is deliberately left alone rather than swept up inside a security fix.

## Verified

- `cargo deny check advisories` → `advisories ok`
- `cargo check --workspace --locked` → clean

no-close: a live upstream advisory, not a tracked issue; arrival already outruns closure 2:1 so this files nothing new.

ont-delta: none — the §11.1 delta for the dependency surface is a `code` contract over the advisory gate, and it waits on ONT-1 (`pv census`), epic #3269.


noahgift:
Superseded by #3277, which lands this together with the fix it was deadlocked against.

#3274 fails `guard-cargo` on cargo-deny advisories — what #3276 fixes.
#3276 fails `guard-cargo` on the pin mismatch and `guard-tree` on the baseline instrument mismatch — what #3274 fixes.

Neither can go green alone, so neither could ever enter the merge queue. #3277 carries both.

PR #3273, opened 2026-09-14
APEX-2a: deterministic render — svg_identical from SVG bytes on X64+ARM64; NEW viz manifest.rs/text.rs/render_determinism.rs (EV-2b breaks as ticks), ci determinism+compare jobs, libm-ban-live.sh; EDITS .clippy.toml (libm bans), breaks.rs powi->sq, svg/lib/scale/plots

noahgift:
Part of #3233. APEX-001 **EV-2a**. Ends at `STOP(PUBLISH: aprender-viz <version>)` — publishing is operator-only and is **not** done here.

## The five rules

| | |
|---|---|
| **1** | `text::to_path(text, font_bytes, size_px) -> kurbo::BezPath` via skrifa, behind `text-path`. The font is **bytes the caller pins** — no lookup, no fallback, no `fontdb::load_system_fonts`. A machine's installed font set is an undeclared input. |
| **2** | `SvgEncoder::text_as_path` emits a `<path>`, so the shapes are in the file rather than a request that the viewer find a font. The existing `text()` is untouched. |
| **3** | PNG compression, filter and adaptive-filter pinned explicitly, in **one** `pin()` helper both entry points call so `write_to_file` and `to_bytes` cannot drift. No `tIME`, no `tEXt`/`iTXt`/`zTXt`. |
| **4** | `manifest::{digest_bytes, Descriptor, Manifest{insert, canonical, root, verify}}` moved from `rmedia-types` **with its tests**, including the FALSIFY-LOCK-001 test that the digest is over raw bytes and not a canonicalisation. |
| **5** | 17 transcendental call sites routed through pure-Rust `libm`; coordinates quantised to `COORD_GRID = 1e-3`. |

## Four defects in the spec row, each measured

**1 — the probe names a file that cannot enforce the rule.** EV-2a asserts a `disallowed-methods` list in the **repository root** `.clippy.toml`. Clippy reads exactly one config, the nearest, and `crates/aprender-viz/.clippy.toml` already exists. A ban planted at the root produced **no lint here** while firing for crates that have no config of their own. A root list satisfies the probe and enforces nothing. The list is in the crate config.

**2 — the spec's ban list catches zero call sites.** The row names `f64::ln`, `f64::log10`, … Every transcendental in this crate is `f32`. Measured: f64-only → **0** violations; adding the f32 twins → **13**. Both are listed here.

**3 — three of five plausible path spellings are silent no-ops.**

```
f64::log10                  -> bans
[...]

PR #3134, opened 2026-09-12
ci(R2a): junit + sccache-metrics + llvm-cov artifacts for Arbiter, and fix the DEAD Coverage Nightly (#3133)

noahgift:
Row R2a of ARBITER-001 (paiml/infra `docs/specifications/arbiter.md` §9.4, infra#532). Makes aprender's `main` CI upload the three artifacts Arbiter's `s-practice` sensor reads, and fixes the coverage lane that has been DEAD for two weeks.

## What changed
- **`scripts/ci_measured.sh`** — a bashrs-clean wrapper: `ci_measured.sh <name> <cmd…>` runs the command, then (whatever its exit code, and forwarding TERM/INT/HUP so a step-timeout kill still collects) moves `target/nextest/ci/junit.xml` to `target/ci-artifacts/junit-<name>.xml` and writes `sccache --show-stats --stats-format json` to `target/ci-artifacts/sccache-<name>.json`. A 26-case `--self-test` covers rc 0, rc 7, no-junit, no-sccache, `bash -c` chains, a TERM'd child (exit 143, files still written), a bad name, and a failing sccache. `bashrs lint`: 0 errors; the shell-lint ratchet improved 8→6.
- **`workspace-test`** — the six nextest/`cargo test` invocations are wrapped, and two `upload-artifact@v7` steps publish `junit` (per-invocation xml + a `tier` marker) and `sccache-metrics`. Every push run to `main` now carries both.
- **`coverage-nightly.yml`** — after `make coverage` a `cargo llvm-cov report --json --summary-only` re-emits the coverage it just measured (no second test run) and uploads it as `llvm-cov`.
- **The DEAD Coverage Nightly, fixed at the root.** It has failed every night since 2026-08-29: `test_silu_backward_at_zero` and `test_silu_backward_length_mismatch` in `crates/aprender-compute/src/backends/gpu/device/backward.rs` called `GpuDevice::new().expect(..)` on the bare-metal clean-room runner (no Vulkan adapter). All nine tests in that module now obtain their device through a `device_or_skip()` helper gated on `GpuDevice::is_available()` — the crate's own convention (`backends/gpu/batch/tests.rs`) — so they RUN where a device exists (the CI container, and this workstation's 4090) and SKIP where none does, instead of the seven that only survived by the recipe's brittle `--skip gpu_` name filter. The infra Nightly Audit's dead-man's switch reported `aprender/Coverage Nightly is DEAD (no-healthy 12d)`.

## Verification
[...]

PR #3280, opened 2026-09-14
feat(ont): §11.2's five counters become measurable — and an anchor with no consumer is refused, not counted

noahgift:
§11.2 gives five counters, a direction each, and records "Today" as `0/0/0/0/[U]`. Those
numbers lived in a spec table. Nothing computed them, `git ls-files contracts/lint-baseline.json`
was empty, and `grep -c '^ont-ratchet' Makefile` was `0`. **A ratchet whose counters are typed
into prose cannot turn.**

## Measured, on the real tree

```
consumer_present           false
contracts_total             1837      <- §11.0 said 1818; main has grown since fa6e35f23
entity_types_registered        0
extractors_implemented         0
contracts_anchored             0
contracts_shaped               0
unanchored_but_bindable      297      <- §11.2 had this as [U]
```

## The gate, and why this is more than a counter

```
$ printf 'entity: kernel\n' | cat - contracts/<any>.yaml > /tmp/a.yaml
$ pv validate /tmp/a.yaml
0 error(s), 0 warning(s)   Contract is valid.
```

`pv validate` **accepts `entity:` by ignoring it.** There is no `pv census`, no `pv extract`,
no `crates/aprender-contracts/src/ontology/`, and nothing in either contracts crate reads the
key — the single grep hit was `identity:` in a test message.

So anchoring the corpus today would write **1837 rows that every gate calls valid and no code
reads**, while `ont.contracts_anchored` climbed from 0 to 1837. A counter measuring its own
decoration, and it would have looked like the fastest progress in the epic.

ONT-001 **R-2 — "zero is a decline, never an accept"** — applies to the instrument as much as
to a verdict. `contracts_anchored` may not RISE while `consumer_present` is false; the guard
refuses it and names ONT-1 as the fix. The probe is **derived from `pv --help`**, never
declared, so it flips on its own the day census lands.

Case table: 10 rows. The deciding one is mutation-verified — replacing the consumer gate with
`if false` turns the table RED.

## Three defects found writing it, each kept as a comment where it bit

- `set -euo pipefail` + a grep matching **nothing** = pipeline exit 1 = silent death. Every
  counter here starts at zero, so the **first real measurement was the one that died**:
  `--write` produced no file, no error, exit 1, and only `bash -x` showed it stopping one line
  after `anchored=0`.
- a `RETURN` trap runs *after* bash destroys the function's locals — cleanup died on an unbound
  variable after the rows had already passed.
- `field()` stripped the wrong characters and compared `"7"` against `"7}}"`.

## And a defect in a neighbouring guard

`check_no_pipe_into_grep_q.sh`'s `HAZARD` regex matched the **second bar of `||`**, so
`cmd || grep -q PAT file` — a logical OR with no producer to kill and no SIGPIPE to misread —
was flagged as a pipe. Two pre-existing sites (`dogfood.sh:1477` greps a FILE,
`qwen-story.sh:281` greps a HERE-STRING), plus every new one written since.

The regex now requires a single bar and ships four rows in **both** directions: `||` not
flagged, here-string not flagged, a real pipe still flagged, and a real pipe *later* on a line
that also contains `||` still flagged. Shrink-only baseline restamped **83 → 81**.

## Wiring — no workflow edit

`guard_universe()` is `git ls-files 'scripts/check_*.sh'`, tracked-only, so the guard was
invisible to the dispatcher until it was **staged** — the free-pass shape that has cost this
repo four guards. Staged, `guard_tree.sh --dry-run` reports `run: scripts/check_ont_ratchet.sh`
and `check_guards_are_wired.sh` exits 0.

`guard_tree --no-cargo`: **55 checks** (was 53), 2 failed — both the pmat 3.39.0 baselines this
branch inherits from main, which #3277 fixes. bashrs 7.4.1: **0 errors** on both touched guards.

ont-delta: resolves scripts/check_ont_ratchet.sh — §11.2 was a table of numbers nothing
computed; it is now a measurement with a refusal (form 4).

no-close: §11.2 is a spec section rather than a filed issue; the ONT rows are tracked in epic
#3269 and this lands the ratchet those rows will move.


noahgift:
## The remaining backlog this guard reveals, measured — deliberately not swept here

Fixing the regex and the two harness rows dropped the ceiling 83 → 79. The rest of what the
guard now sees, triaged by **position**, because position is what decides the blast radius:

```
raw sites (all .sh)                      91
armed (file actually enables pipefail)   79
  in scripts/tests/                      12
  inside a row()/$want case-table helper  8   <- the dangerous shape
  in a --self-test block                  1
```

The eight row-shaped files, all with `pipefail` armed:

```
scripts/check_hardcoded_paths.sh               1 site
scripts/check_release_assets.sh                1
scripts/check_roadmap_sorted.sh                1
scripts/ci_gpu_touched.sh                      5
scripts/ci_test_tier.sh                        6
scripts/tests/baseline_tool_version_test.sh    2
scripts/tests/check_tool_versions_test.sh      1
scripts/tests/ci_gpu_touched_test.sh           2
```

Why these eight and not the other 71: a `row()` helper compares `$rc` to `$want` **and** greps
the captured output. When the grep matches, `printf` takes SIGPIPE, the pipeline returns 141,
and **the row reports FAILED precisely because it MATCHED**. That inverts a case table rather
than merely breaking it, and a case table is the thing other guards are proven by.

Two of them are CI *decision* scripts rather than reporters:

- `ci_test_tier.sh` decides quick/full/reuse. A false RED there changes **what CI runs**.
  #3273 failed on `Decide the test tier (BSE-17, PMAT-1077)` earlier this evening; I have not
  established that it was this mechanism, and I am not claiming it was.
- `ci_gpu_touched.sh` decides whether the GPU lane fires.

## Why I stopped

Nineteen mechanical edits across eight CI-critical scripts, at 22:00, with sixteen entries
mid-drain, is how a cleanup becomes an outage. The three fixed tonight (#3134's
`ci_measured.sh`, and both rows in `check_roadmap_sorted_test.sh`) were fixed because they were
**blocking** — this set is not.

Each file ships its own `--self-test`, so the sweep is verifiable one file at a time, and the
baseline is shrink-only so it can be taken in any order without stranding anyone.

PR #3285, opened 2026-09-14
fix(nightly): guards-nightly has been RED naming a step that runs every night

noahgift:
Five scheduled lanes are red (`guards-nightly`, `Coverage Nightly`, `Silicon Nightly`,
`examples-nightly`, `CUDA Nightly`). This is the one whose cause is entirely **inside the
guard** rather than in the thing it guards.

```
FAIL: manifest names a step this workflow does not run:
      Test-tier decision case table (BSE-17): quick/full/reuse, drift is ENV
```

The step is there — `guards-nightly.yml:158`:

```yaml
- name: "Test-tier decision case table (BSE-17): quick/full/reuse, drift is ENV"
```

The inline check matched the **bare** spelling:

```bash
grep -Fq -- "- name: $name" .github/workflows/guards-nightly.yml
```

A step name containing a colon-space **must be quoted in YAML**, and this one contains
`(BSE-17): quick`. So the matcher looked for `- name: Test-tier…` while the file said
`- name: "Test-tier…`, and the lane has been red since that name was quoted — reporting a step
that runs on every nightly as one the workflow does not run.

A guard that names a real step as missing is worse than no guard: it teaches the reader to skip
the lane.

## Extracted, not patched in place

Logic inside a `run:` block cannot be unit-tested and cannot carry a case table — which is why
a one-character class of defect survived in it. `scripts/check_guards_nightly_manifest.sh`
matches on the **parsed** name, so all three YAML quotings of one string are one step.

```
bare name parsed                              ok
DOUBLE-quoted name unquoted (the defect)      ok
single-quoted name unquoted                   ok
trailing whitespace stripped                  ok
manifest fully covered            -> exit 0
a genuinely absent step           -> exit 1   <- still fails; not made permissive
vacuity floor, <5 entries parsed  -> exit 1
```

Against the real tree: `ok  7 manifest entries, every one a step in this workflow`. Mutation:
appending one bogus entry turns it red again, so the matcher was widened without being disarmed.

## The self-test found a defect in its own harness first

Invoking `step_names` through `bash -c 'source …'` re-ran this file's own dispatch and returned
nothing, so four rows failed while the parser was correct. **A broken harness reporting a
working parser as broken is the same error being fixed, in the other direction** — worth
recording because the row that matters most is the one most easily faked, either way.

## Also

`check_roadmap_sorted.sh` caught the roadmap entry at a stale slot: main gained PMAT-3228/3229
while this branch was being written, so "after 3222" was no longer sorted. Re-inserted among
its numeric peers.

`guard_tree.sh --dry-run` reports the new guard `wired-elsewhere guards-nightly.yml`;
`check_guards_are_wired.sh` exits 0. bashrs 7.4.1: 0 errors.

The other four red lanes are separate causes and are **not** touched here: `Coverage Nightly`
(llvm-cov), `Silicon Nightly` (the x86_64-cpu SIMD axis genuinely failing), `examples-nightly`
(an example failing), `CUDA Nightly` (PP-26 batch-invariance, a known divergence).

no-close: a CI lane fix with no filed issue; the four remaining red nightlies are separate
causes and need their own triage.

ont-delta: none — a CI guard fix; it adds no entity, shape, verdict reason or resolves target.


noahgift:

PR #3267, opened 2026-09-14
fix(zram): the cuda feature says why it cannot build, instead of a cascade

noahgift:
## What

`aprender-zram-core --features cuda` and `aprender-zram --features cuda` — the last two red pairs in the 430-pair matrix (#3262) — now report **one named error** saying exactly what is missing, instead of a cascade of unresolved imports.

They stay red. That is the honest state, and this PR explains why.

## Three verified gaps

Each read directly out of the tree, not inferred:

1. `trueno_gpu::driver::PinnedBuffer` and `GpuBuffer::copy_{from,to}_pinned_async` do not exist and never did in this monorepo — `git log -S PinnedBuffer -- crates/aprender-gpu/` is empty. Only synchronous pageable copies exist.
2. `src/gpu/batch/mod.rs:610` launches the entry point `lz4_compress_warp_shuffle` with 32 threads and `shared_mem: 0`. **No kernel of that name exists.** The one that does, `lz4_compress_warp`, is documented as 128 threads per block with shared memory.
3. `Lz4WarpDecompressKernel::build_ptx` is a **stub** whose entire body is `ctx.label("L_exit")`. There is no GPU LZ4 decompressor in this workspace at all.

So this is not one API away from working. It is wired to a kernel ABI that was removed at consolidation, and its decompressor does not exist.

## Decided by quorum, two rounds

**Round 1** (3 agy lanes) voted 2/3 to rewrite the path on the existing pageable API so it would compile. Facts 1–3 were verified *after* that vote.

**Round 2** put those facts to three fresh lanes. All three **overturned round 1** — unanimous that making a dead path compile in order to turn a matrix cell green is theater. Round 2 then split 2/1 between deleting the path and the option in this PR.

**I took the minority option**, on a fact neither deleting lane had: this crate's **default build is green**, and `gpu/batch` is live, compiling, *public* API that `gpu/hybrid.rs` consumes. Only the cuda-gated arms are dead. Both deleting lanes argued from "a dark, unused feature" — the module is neither. Deleting it would remove live public types (`GpuBatchCompressor`, `BatchResult`, `BatchDecompressResult`, `GpuBatchConfig`) for no gain, which neither lane actually argued for; one said outright that this option "best serves someone who later wants to finish the GPU path".

## How

[...]

Issue #3182, opened 2026-09-12
G3.EX: 38 of 44 non-pass rows were classifier gaps, not defects — feature refusals need a re-run, not a skip


noahgift:
## What the first full G3.EX sweep measured

981 example targets, `timeout 120`, on lambda-vector (x86_64, RTX 4090). Evidence: `evidence/dogfood/0.67.0/examples.tsv`.

| class | rows |
|---|---|
| `pass` | 740 |
| `needs-data` | 168 |
| `needs-args` | 29 |
| `fail` | 23 |
| `timeout` | 21 |

The 44 `fail` + `timeout` rows were then re-run one at a time with their own `required-features` and their logs kept. That is what this ticket is about: **of the 44, only 6 rows were genuine defects.** The rest were the classifier answering a question it was not built to answer.

| what the row actually was | rows | what the sweep called it |
|---|---|---|
| runs clean once the feature it names is enabled | 9 | `fail` |
| long-running by design (server, TUI, benchmark) | 20 | `timeout` |
| missing model/data, worded outside the regex | 3 | `fail` |
| missing CLI argument, worded outside the regex | 3 | `fail` |
| needs a TTY | 1 | `fail` |
| this box's glibc vs a prebuilt `ort` artifact | 1 | `fail` |
| a wall-clock perf gate failing under load | 1 | `fail` |
| **genuine defect** | **6** | `fail` |

## The three gaps, in order of how many rows they cost

**1. A feature the example names at RUNTIME, not in `required-features` (9 rows).** `cuda_training_benchmark` prints "This benchmark requires the 'cuda' feature" and exits 1; with `--features cuda` it exits 0. Same for `cublas_fp8_7b_reproducer` (cuda), `compressed_kv` and `compressed_index` (compression), `semantic_embeddings` (embeddings), `tensor_example` (tensor), `showcase_shell` (shell-autocomplete). The sweep already reads `required-features` from `cargo metadata`; these examples declare none and gate themselves in code instead.

The right behaviour is not a new skip class. It is a **re-run**: parse the feature the example names, run it again with that feature, and classify the SECOND run. A skip here would hide the case where the example is broken *with* its feature — which is exactly what `gpu_info` turned out to be (#3179).

[...]

PR #3288, opened 2026-09-14
fix(examples): an example built WITHOUT its feature is not a broken example — needs-feature

noahgift:
`examples-nightly` is one of **five red scheduled lanes**. Its verdict is currently not a
trustworthy signal about whether examples work — which is the exact property
APR-RELEASE-001 §4.2 and T-1 want to depend on.

## Run 34826552378, 980 example targets

```
summary pass=602 fail=22 timeout=6 needs-args=30 needs-hardware=2 needs-data=318
```

The 28 `fail`+`timeout` rows, classified by what the output actually says:

```
 6   the example is not built with the feature it needs   <- NOT broken
 2   needs a network peer (192.168.50.100:9000)           <- NOT broken
 1   needs a TTY ("No such device or address")            <- NOT broken
 6   timeout at 180 s (buggy_server, brick_computer ...)  <- servers by design?
13   remainder
```

**A quarter of the lane's red is a missing label, not a defect.** This PR takes the 6:

```
aprender-db::compressed_kv          This example requires the 'compression' feature.
aprender-rag::compressed_index      This example requires the 'compression' feature.
aprender-rag::semantic_embeddings   This example requires the 'embeddings' feature.
aprender-distribute::tensor_example Error: This example requires the 'tensor' feature.
```

plus cargo's own refusal, which phrases it differently:

```
target `gpu_info` in package `aprender-zram-core` requires the features: cuda
```

## Like `needs-hardware`, not like `needs-args`

A `needs-feature` row is a **SKIP that must be re-run WITH the feature before a release
verdict**. A green lane on a default-feature host must never be readable as "the compression
example works" — which is why the row cites the line that classified it.

This follows the precedent the file set for `needs-data` on 2026-09-12 (173 rows), and obeys
the rule its own header states: *"one turns a real defect into a skip, so each addition belongs
in the selftest."*

## The regex ships both polarities

A skip class can only ever turn a `fail` into a non-failure, so the dangerous direction is
tested explicitly:

```
MUST match      This example requires the 'compression' feature.
                Error: This example requires the 'tensor' feature.
                ... requires the features: cuda
MUST NOT match  thread 'main' panicked at src/main.rs:11:39
                error[E0432]: unresolved import `trueno_gpu::driver::PinnedBuffer`
                Error: Os { code: 6, kind: Uncategorized }
                assertion failed: left == right
                feature is required for this operation     <- the deciding near-miss
```

Mutation, both directions:

```
regex matching nothing       -> selftest RED
regex widened to "feature"   -> selftest RED
shipped regex                -> green
```

**The harness caught its own staleness**: a completeness assertion hardcoded *"one of the six
classes"* and went red on the seventh. Updated rather than relaxed — that assertion is what
stops a row being written with no class at all.

## Deliberately not swept in

The 2 network rows, the 1 TTY row, the 6 timeouts and the 13 remaining are separate causes and
deserve their own triage. One of the 13, `aprender-zram-core::gpu_info`, is **already owned by
#3267**, which found that `trueno_gpu::driver::PinnedBuffer` *"does not exist and never did in
this monorepo"*.

bashrs 7.4.1: 0 errors. selftest: all rows passed.

no-close: the measurement this implements is **#3182** (open, 0.68.0), which counted 9
of 44 non-pass rows as feature refusals miscalled `fail`. This PR adds the `needs-feature`
class for exactly those rows; #3182 also covers the 20 long-running-by-design `timeout`
rows and the needs-data surface, so it stays open. The four other red nightlies
(`Coverage Nightly`, `Silicon Nightly`, `CUDA Nightly`, and the rest of `examples-nightly`)
are separate causes.

ont-delta: none — a classification fix in a CI harness; it adds no entity, shape, verdict
reason or resolves target of its own.


noahgift:

Issue #3240, opened 2026-09-14
mdBook CI reports a vanished dep-info file as a book that does not build — cargo_classify.sh has known this signature since #2712 and no workflow step calls it


noahgift:
## The defect

`mdBook CI` runs three cargo-building steps as **bare workflow steps**. When cargo
dies for a HOST reason, the job reports a bare exit 101 on a check named
`Build Book` — indistinguishable from a book that does not build.

Measured 2026-09-14, run [34799968553](https://github.com/paiml/aprender/actions/runs/34799968553), step `CLI parity gate`:

```
error: could not parse/generate dep info at: target/release/deps/minijinja-652b98d7c28c791c.d
Caused by:
  No such file or directory (os error 2)
error: could not parse/generate dep info at: target/release/deps/arrow_ord-b90606e68d473618.d
error: could not parse/generate dep info at: target/release/deps/arrow_cast-c5fa33bc4fec0f56.d
error: failed to compile `apr-cli v0.67.0`
```

Three crates' dep-info files vanished over 35 s while cargo was writing them.
The same branch had been green on the same job 100 minutes earlier.

`could not parse/generate dep info` is **row C8 in `scripts/cargo_classify.sh`'s own
committed case table** — already classified ENV, already fixture-tested, sourced by
six `check_*.sh` guards. None of those six is a workflow step, so a cargo command
that runs as `run: cargo install …` has never had an arm at all.

## Why this is the incomplete half of a fix that already landed

`book.yml`'s bashrs step carries this comment, written when `cargo install bashrs`
was removed from it:

> it died twice on 2026-09-13 losing a dep-info file under /tmp

Same signature, same job, same week. The fix stopped at the step that hurt.
`cargo install --path crates/apr-cli` was left, and it is the one that died on
2026-09-14. Removing the install is not available there — apr-cli *is* the tree
under test — so the remaining move is to name the fault.

## Blast radius

8 of the last 40 `mdBook CI` runs failed. Each one reaches a reviewer as
`Build Book fail` with nothing to distinguish a host fault from a real one.

## Second defect in the same job, same class

```yaml
bashrs lint scripts/check_book_*.sh 2>&1 | tee /tmp/bashrs.log
! grep -qE "Summary: [1-9][0-9]* error" /tmp/bashrs.log
```

The default GH shell is `bash -e {0}` with **no pipefail**, so the pipeline's status
is `tee`'s, which is 0 whatever bashrs did. A crashed linter leaves a log with no
[...]

Issue #2822, opened 2026-08-31
CI jobs fail with cargo exit 101: target/debug/deps files vanish MID-BUILD (ENOENT, not ENOSPC — 884G free)


noahgift:
`workspace-test` and `guard-runner-labels` jobs are failing with cargo `exit code 101` whose root cause is **ENOENT on files inside `target/debug/deps/`, mid-build** — not a full disk, and not a code defect.

## Evidence

PR #2804, job `99507516986`, `guard-runner-labels` on `intel-clean-room-8`, step `aprender-profile lib tests`:

```
   Compiling crossterm v0.28.1
error: could not parse/generate dep info at:
  /home/noah/data/actions-runner-8/_work/aprender/aprender/target/debug/deps/proptest-2f6ffc38b471ed82.d

Caused by:
  No such file or directory (os error 2)
warning: build failed, waiting for other jobs to finish...
error: could not parse/generate dep info at: .../criterion-f848472e779efcd5.d
error: could not parse/generate dep info at: .../sqlparser-9d72f144111fd70c.d
error: could not parse/generate dep info at: .../parquet-10f6b7e1ab2343a0.d
```

Four crates fail simultaneously with `os error 2` while unrelated crates are still compiling. The check-run annotation is only `Process completed with exit code 101`.

## Why this is NOT the disk

Measured on `mac-server` at peak (12 live CI runs, 15 cargo processes, load 578):

| | |
|---|---|
| `/` | 2.6T used, **884G available**, 75% |
| inodes | **11%** used |
| total across all 16 runner `_work` | 244G |
| largest in-flight target dir | 6.3G |

There is no space or inode pressure, transient or otherwise.

**A caution for whoever picks this up:** the repo's recorded shorthand maps `could not parse/generate dep info` to ENOSPC. That mapping is over-general — the `Caused by:` line distinguishes them, and here it says `No such file or directory`, not `No space left on device`. Read the cause, not the headline. Separately, grepping job logs for the prose `no space left` matches a *guard fixture* (`ok C17 a test asserting the prose "No space left..." -> CODE`), which will produce a false positive; that fixture exists precisely to stop this misclassification.

## What is established vs. hypothesised

**Established:** the files were removed while cargo held them; the disk was not short; the failure is environmental, not a code defect in the PR.

**Not established:** what removed them. Candidates worth checking, in order:

[...]

Issue #2775, opened 2026-08-29
nextest reports a vanished test binary as a test failure, and cargo_classify.sh already knows that signature but is not wired to workspace-test


noahgift:
Two PRs failed `workspace-test` within the same window today, both reported as **test failures**, both actually artifacts vanishing mid-run:

**#2773** — nextest could not spawn the test binary:
```
error: [double-spawn] failed to exec "/workspace/target/debug/deps/batuta-f9953bd983f7eb94"
  No such file or directory (os error 2)
```
Reported as `TRY 3 FAIL` on four named tests in `stack::releaser_preflight::tests`. All 108 tests in that module **pass locally** on the same commit.

**#2742** — the linker could not find its inputs:
```
clang: error: no such file or directory: '/workspace/target/debug/deps/...rcgu.o'
clang: error: no such file or directory: '.../liblibc-b75774821d156ccb.rlib'
```
`libc` cannot fail to link for code reasons.

## The gap

`scripts/cargo_classify.sh` — this repo's own ENV/CODE classifier — **already lists `[double-spawn] failed to exec` among its ENV signatures**, alongside `could not execute process`, `No space left on device` and `could not parse/generate dep info`.

But it is not applied to the `workspace-test` job. nextest surfaces a vanished binary as a **test failure on a named test**, so the job fails as though the code is broken, and the classifier never runs on that output.

The practical cost is misdiagnosis. A named failing step plus a named failing test is the strongest "real defect" signal a reader has, and here it is wrong. I initially classified #2742 as a real defect for exactly that reason, and only the log content corrected it.

## Suggested

Run `cargo_classify.sh` over the `workspace-test` output too, and have an ENV verdict fail with a message that **says it is environmental** rather than naming tests. Per the standing rule in this repo, an ENV verdict must still exit non-zero — the point is not to pass, it is to stop attributing a host fault to the code.

Worth adding `no such file or directory: .*\.(o|rlib|rcgu\.o)` to the signature list while there; the #2742 linker form is not currently matched.

## Context, since it is probably the cause

There are **122 registered git worktrees** on the build host and multiple concurrent agents sharing one target directory. Earlier today 199 tracked files were deleted from one worktree by something outside the agent working in it. Disk pressure and cross-agent interference on the runners are the most likely source, and both failures are consistent with it.

noahgift:
Blocks PP-066 **C0** (spec §4). Also: F-28, G-7 (proposed, class parent).

[...]

PR #3289, opened 2026-09-14
fix(silicon): a vanished dep-info file was reported as a SIMD failure — the classifier existed and this lane did not call it

noahgift:
Silicon Nightly is red. Its failing step is named

```
SIMD and numeric paths (x86_64, widest vectors)
```

which, on an ML framework, reads as **arch-specific numerical incorrectness**. It is not.
3511 tests passed; the job died afterwards on:

```
error: could not parse/generate dep info at:
    .../target/release/deps/arrow_ipc-b83c0a68418506db.d
```

A dep-info file vanished while cargo was writing it. That is a **host fault**, it is already
classified, and it is already fixture-tested in this repository — #3241 built
`scripts/cargo_step.sh` for exactly this, after mdBook CI reported the same error as *"a book
that does not build"*. Its case table carries the row `W2  dep-info ENOENT is ENV`.

## Measured coverage of that classifier (main @ `a0634f7b3`)

```
cargo invocation lines across all workflows   75
routed through cargo_step.sh                   5   (book.yml, and only book.yml)
```

The instrument covers **7%** of the surface, and the lane that just died is in the other 93%.

`scripts/cargo_classify.sh` has said ENV-or-CODE since #2712/#2822 and six `check_*.sh` guards
source it — but a cargo command running as a **bare workflow step** has no arm at all. That is
the shape #3241's own header describes, fixed for one lane rather than for the class.

## What this changes

Both legs of `silicon-nightly` — `x86_64-cpu` and `aarch64-cuda-sm121`, four cargo invocations
— route through the classifier, with the self-test caller added to each job first: *a
classifier nothing proves is the same theater as no classifier* (book.yml's pattern, and what
`check_guards_are_wired.sh` requires of any script shipping a `--self-test`).

**It does not make the lane green, and must not.** From `cargo_step.sh`'s own contract:

> ENV still exits NON-ZERO — a gate that goes green on "we could not measure" is the defect it
> exists to name.

What changes is the **claim**: a host fault stops reading as a SIMD defect, so nobody spends a
morning on AVX-512 correctness because a file disappeared.

## A defect I introduced and caught before pushing

Worth recording because the fix looked right. The first pass used a naive string replace, and
the **second replacement matched the continuation line inside the wrapper the first had just
created** — double-wrapping the x86_64 leg while leaving aarch64 raw. Redone by line position.

The job-name derivation was wrong too, labelling the x86_64 steps `coverage:` (a real sibling
job in this file). Corrected after checking which job each line actually sits in rather than
trusting the first answer.

Guards: `check_workflow_cargo_packages.sh` 0, `check_workflow_env_defined.sh` 0,
`cargo_step.sh --self-test` 0. YAML parses, 4 jobs.

The other four red nightlies are separate causes: `guards-nightly` (#3285), `examples-nightly`
(#3288), `Coverage Nightly`, `CUDA Nightly` (PP-26).

no-close: the underlying class is **#2822** (open, 0.68.0) — `target/**/deps/*.d` files
vanishing mid-build with ENOENT on a shared runner work dir, the same signature and the
same crates (crossterm, arrow_*). This PR does NOT fix that; it stops silicon-nightly
reporting it as a SIMD failure, which is the #2775 / #3240 misattribution class. #2822
stays open for the root cause. The remaining red nightlies need their own triage.

An earlier revision said "no filed issue" — that was a triage miss.

ont-delta: none — a CI classification fix; it adds no entity, shape, verdict reason or resolves
target of its own.


noahgift:

PR #3116, opened 2026-09-11
fix(compute): nine GPU tests panicked without an adapter — the nightly coverage run was RED six days on an environment fact (PMAT-1106)

noahgift:
## What
`backends::gpu::device::backward::tests` called `GpuDevice::new().expect("GPU device")` nine times. Under `cargo llvm-cov` on yoga-build2 (docker runner, no GPU passthrough) two of them panicked and the whole nightly coverage run exited 2 — run 34575134766, and the run is RED six days running (the other half of that redness is the same class in aprender-cgp, #2848, pulled into 0.67.0).

A missing adapter is an environment fact, not a kernel defect. `device_or_skip()` prints `SKIP: no GPU adapter on this host (<err>)` and the test returns; the box that has an adapter stays the gate.

## Proven both ways (lambda-vector, `cargo test -p aprender-compute --lib --features gpu -- backends::gpu::device::backward --nocapture`)
| polarity | result |
|---|---|
| `VK_ICD_FILENAMES=/nonexistent` (no adapter) | 9 `SKIP:` lines, 9 passed in 0.00 s |
| real adapter | 0 `SKIP:` lines, 9 passed in 1.42 s |
| `cargo fmt --check`; `cargo clippy -p aprender-compute --all-targets -- -D warnings -A unused-variables` (CI's chain, default features) | clean, 0 errors |
| `cargo clippy -p aprender-compute --lib --tests --features gpu -- -D warnings` | 1 error, **pre-existing and not in this diff**: `len() == 0` at `backends/gpu/pool.rs:169`, a gpu-feature-only test target no gate lints today (same class as the aprender-gpu dark targets noted in memory) |

Falsifier of record: the next `coverage-nightly.yml` run (03:00 UTC) — with this and #2848 on main it must reach the report step and print a line-coverage figure instead of exiting 2 in `cargo test`.

Refs PMAT-1106, #2848, operator rule 2026-09-11 ("nightly we should have a low priority code coverage run with goal of eventually getting to 95% coverage").

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3290, opened 2026-09-14
fix(zram): a same-process ratio is load-immune but NOT instrumentation-immune

### LINKED ISSUE #3185: coverage-nightly has never completed on the yoga pool (4/4 red since 2026-09-09): each night a different unit test asserts the box, so the 95 % ratchet has no baseline
## What happened

`coverage-nightly.yml` (#3123: low-priority coverage run on yoga toward the 95 % target) concluded `failure` on every scheduled run since it moved to the yoga pool — 34325094997 (09-09), 34451014241 (09-10), 34575134766 (09-11), 34680781819 (09-12) — and on the 0.67.0 cut when dispatched by hand (34702490406, `yoga-build2`, 2026-09-12T15:31Z). Every time `make coverage` (Makefile:586) exits 2 because ONE unit test fails under llvm-cov, and the test is different each time:

- main, 09-12 schedule: `backends::gpu::device::backward::tests::test_silu_backward_at_zero` and `…_length_mismatch` panic at `crates/aprender-compute/src/backends/gpu/device/backward.rs:1103` / `:1120` (no adapter on the box — the class #3116 fixed for nine other tests).
- the cut aa61219ef (which contains #3116): those two pass; instead `benchmark::tests::test_decompression_faster_than_compression` panics at `crates/aprender-zram-core/src/benchmark.rs:522`: `Decompression 2.6 MB/s much slower than compression 11.4 MB/s` — a throughput ratio asserted inside a `--lib` unit test, measured under coverage instrumentation on a shared box.

So the coverage number for 0.67.0 does not exist, and §5 P3 of APR-RELEASE-001 ("target = measured baseline + 2 points per ratchet toward 95 %") cannot start.

## Why

Two classes, both known here: a test that asserts the hardware it runs on (#3129, #3135, #3184 are the same shape on the GPU jobs), and a wall-clock/throughput assertion in a unit test (the rule in memory and in #3131, which already moved one such benchmark behind `bench-gates`).

## Ask

1. `benchmark.rs:522` — move the ratio assertion behind the `bench-gates` feature (as #3131 did), or assert correctness (round-trip bytes) and *record* the throughput.
2. `backward.rs:1103/1120` — skip with a named reason when no adapter is present, counted as skipped (the #3116 treatment).
3. `make coverage` — run the suite `--no-fail-fast` so a single box-assertion does not zero the night's report, and print the coverage number even when a test fails (report-only is the job's declared contract).
4. Then record the first baseline in `docs/build-ledger/` per APR-RELEASE-001 §5 P3.

Refs #3123, #3116, #3131, #2307, PMAT-1098 (0.67.0 train).

noahgift:
Coverage Nightly has been red **every night**. #3227 fixed the test it was named for (a lock in
`apr_serve_tests.rs`) and landed 02:19Z — and the 08:18Z run on `a1a19c4fc`, which **contains
that fix**, failed on a different one:

```
thread 'benchmark::tests::test_decompression_faster_than_compression'
  panicked at crates/aprender-zram-core/src/benchmark.rs:522:9
test result: FAILED. 403 passed; 1 failed
```

## Why the obvious defence didn't work

This test is **already** a same-process ratio — the shape its neighbour
`test_f058_entropy_overhead_minimal` was rewritten into, whose comment explains why:

> load-immune because both measurements are taken in the same process, microseconds apart,
> under whatever load exists

**Load-immune is not instrumentation-immune, and they are different axes.** `cargo llvm-cov`'s
overhead scales with the number of instrumented branches *executed*, and the compress and
decompress paths do not execute the same number. Under coverage this ratio measures
instrumentation density, not the algorithm — and no amount of same-process-ness helps, because
the distortion isn't noise, it's systematic and asymmetric.

The ratio had already been widened once for exactly this reason — *"In debug builds with
coverage instrumentation, decompression may be slower"*, `0.25` — and the lane failed on it
anyway. Widening again is the treadmill f058's own comment names: *"the number it measures is
how loaded the machine is."*

## So: assert where the number means something, report where it doesn't

```
#[cfg(coverage)]      print both rates; assert only that BOTH paths produced a
                      rate — true under instrumentation, and a zero is still caught
#[cfg(not(coverage))] the 0.25 ratio, unchanged
```

`cargo llvm-cov` compiles with `--cfg=coverage` — verified, not assumed:

```
$ cargo llvm-cov show-env
__CARGO_LLVM_COV_RUSTC_WRAPPER_RUSTFLAGS='-Cinstrument-coverage--cfg=coverage'
```

A compile-time split, not a runtime guess at whether we're instrumented.

## Proven, because a passing test doesn't say which branch ran

```
without the cfg      0 "informational" lines  -> the ASSERT branch ran
with --cfg=coverage  1 "informational" line   -> the REPORT branch ran
mutation 0.25 -> 1000.0, no cfg               -> FAILED
```

The third row is the one that matters: **the assertion is still load-bearing where it's kept**,
so this narrows the claim rather than deleting the test.

## The distinction is the finding

Three patterns now live in this file, and they are not interchangeable:

| | defends against |
|---|---|
| f058's ratio rewrite | **load** |
| F053's report-don't-assert | an **absolute** bound being meaningless in a required check |
| this | a **ratio** being meaningless *only where the instrument distorts it* |

fmt clean, `clippy -D warnings` 0 errors, roadmap additive (`reserialised=0`).

Closes #3185

`Coverage Nightly` had two independent failures and #3185 documented BOTH: the
GPU-adapter pair (`test_silu_backward_at_zero` / `…_length_mismatch`, fixed by the
#3116 class) and this one. #3185 was closed `COMPLETED` on 2026-09-14T03:01Z with the
second still live, and the lane failed again 5 h 17 m later on exactly the test its
body names. Reopened with the evidence; this PR is what closes it.

An earlier revision of this body said "a red-nightly fix with no filed issue" — that
was a triage miss, not a fact.

ont-delta: none — a test-assertion fix; it adds no entity, shape, verdict reason or resolves
target of its own.


noahgift:

PR #1364, opened 2026-05-02
test(realizar): backend-dispatch-v1 — bind FALSIFY-BD-001..005 at PARTIAL_ALGORITHM_LEVEL

noahgift:
## Summary

Binds FALSIFY-BD-001..005 from `contracts/backend-dispatch-v1.yaml` at
`PARTIAL_ALGORITHM_LEVEL` via 5 verdict functions covering backend
dispatch correctness (31 tests).

| Gate | Predicate |
|------|-----------|
| FALSIFY-BD-001 | n ≥ 100k ⇔ chose GPU |
| FALSIFY-BD-002 | rep>0.3 OR uniq<10 ⇔ classified garbage |
| FALSIFY-BD-003 | abs(QK score) ≤ sqrt(head_dim) |
| FALSIFY-BD-004 | encode(decode(tokens)) == tokens |
| FALSIFY-BD-005 | SIMD == scalar bit-identical (tolerance 0.0) |

## Test plan

- [x] `cargo test -p aprender-core --lib format::bd_001_005` — 31/31 pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3297, opened 2026-09-15
roadmap fragments (§1): one file per ticket, so the merge path has no shared mutable file

### LINKED ISSUE #3296: roadmap fragments: one file per ticket so the merge path has no shared mutable file — Amdahl serial fraction is currently 1
## Diagnosis — queueing, not compute

Little's Law: 42 open ÷ 8.8 merges/day ≈ **4.8 days lead time**. Arrival exceeds drain,
and drain is capped at 1 because **every PR writes `docs/roadmaps/roadmap.yaml`**. That
makes the Amdahl serial fraction on the merge path **1**. 27 runners cannot help a queue
whose critical section is a shared file — measured at the time of writing, aprender had
4 runs in progress carrying **5 jobs** while the queue drained at 8.8/day.

## Why the current mitigation cannot work

`.gitattributes` declares `docs/roadmaps/roadmap.yaml merge=roadmap`. `roadmap` is a
**custom** driver registered per-invocation by `scripts/ci_resolve_dirty.sh`
(`git -c merge.roadmap.driver=…`), deliberately, so a `git config` write does not leak
into sibling worktrees. GitHub's merge-queue server has no such config, so **inside the
queue the driver is inert** and git falls back to a plain text merge.

Only git's **built-in** drivers work server-side. `union` is built-in — but it is correct
for `impl-estimates.jsonl` (line-delimited, order-insensitive; #3256) and **wrong** for
`roadmap.yaml`, which requires *sorted* insertion. Measured 2026-09-15 on the nine-PR
batch #3295: the textual merge placed `PMAT-3226` after `PMAT-3229` and
`check_roadmap_sorted.sh` went RED — *"prefix PMAT must be numerically ascending among
its own peers"* — while all nine inputs were individually sorted and green.

**Two files, two policies.** No single merge driver fixes both.

## The change: one fragment per ticket

    docs/roadmaps/entries/PMAT-NNNN.yaml     <- what a PR writes
    docs/roadmaps/roadmap.yaml               <- GENERATED on main, never hand-edited

Unique filename by construction ⟹ N PRs pairwise disjoint on the roadmap ⟹ **conflict
rate 0% by proof, not by luck**. The sort problem stops existing because there is no
shared file to sort — ordering becomes a property of the aggregator, which is total.

### Gates (base-owned — the producer is never the gate)

1. **Trailer ↔ filename parity**: a new `entries/<id>.yaml` must match the commit's
   `Pmat-Ticket: <id>` trailer.
2. **No hand edits**: a PR touching `roadmap.yaml` directly fails.
3. `check_roadmap_sorted.sh` runs **against the aggregate only**.
4. A `pv` contract that the aggregate's ordering is total, ids unique, every fragment
   present. Extends the existing `contracts/apr-roadmap-ids-unique-v1.yaml` and
   `contracts/apr-roadmap-additive-diff-v1.yaml` rather than duplicating them.

### What breaks, named up front

- `git blame docs/roadmaps/roadmap.yaml` stops being chronological; history moves to
  `git log docs/roadmaps/entries/`.
- Anything consuming the raw file keeps working **only because the aggregate stays
  committed on `main`** — that is deliberate, and is why the file is generated
  post-merge rather than gitignored.
- The open PRs that carry inline roadmap edits need a one-time conversion.

## Sequence

This lands **before** raising `max_entries_to_merge` (live ruleset `17836320`:
`max_entries_to_build=3`, `max_entries_to_merge=1`, `grouping_strategy=ALLGREEN`,
`merge_method=SQUASH`). Grouping 8 entries that all write `roadmap.yaml` amplifies the
collision instead of fixing it. Once fragments land, native queue batching gives one
commit per PR on `main` — which hand-squashing a batch does not.

## Falsifier

A merge group containing ≥ 2 roadmap-touching PRs builds green, and each merges as its
own commit. Today that group ejects.

Refs #3294 (five-whys), #3256 (`union` for the ledger).
Measured 2026-09-15 on `main` @ e6e13ad11.

noahgift:
§1 of the queue architecture. Removes the shared mutable file from the merge path.

## The mechanism

Every PR writes `docs/roadmaps/roadmap.yaml`, so the Amdahl serial fraction on the
merge path is **1**: N pull requests contend on one file however disjoint their code is.
That is why 27 runners running 5 jobs cannot raise a drain of 8.8/day.

`.gitattributes` declares `merge=roadmap`, but that driver is **custom** and registered
per-invocation by `ci_resolve_dirty.sh`, so GitHub's merge-queue server never runs it.
Measured on batch #3295: the textual merge placed `PMAT-3226` after `PMAT-3229` and
`check_roadmap_sorted.sh` went RED, while all nine inputs were individually sorted and
green.

## The change

A ticket writes `docs/roadmaps/entries/<ID>.yaml`. Unique filename by construction ⟹
PRs pairwise disjoint on the roadmap ⟹ **conflict rate 0 by proof, not by luck**.
`roadmap.yaml` is the aggregate, and ordering becomes the aggregator's job: an entry is
placed at the slot `check_roadmap_sorted.sh` demands, using `parse_id` imported from
`roadmap_merge.py` so the guard and the generator cannot drift.

## Base + fragments, not a full split — and the obstacle that decided it

The base keeps its bytes exactly. That is what row 1 proves, and it is what keeps this
diff additive.

It also sidesteps a real blocker found by measurement. `roadmap.yaml` carries
`created: &id001 …` on entry 620 (`APR-PERF-GATE-001`), aliased by **17 later entries**
— `roadmap_diff.py`'s own docstring names this as real data its block parser cannot
resolve. A full 878-way split would emit 17 fragments that are individually unparseable
YAML.

I tried inlining that anchor as a precondition. It is a pure re-serialisation, and
`check_roadmap_diff_additive.sh` correctly refuses it:

```
VIOLATION reserialised: id=PERF-010 (bytes differ, no field actually changed)
reserialised=18   PMAT-980 (#2874)
```

The guard is right — so the base is never rewritten and the question does not arise.
That commit was dropped rather than worked around.

## Verification

| check | result |
|---|---|
| aggregate with 0 fragments vs `roadmap.yaml` | **byte-identical**, 17,785 lines |
| `roadmap_fragments.py --selftest` | 9 rows, 0 red |
| ↳ includes a **mutation row**: forcing append-only placement must turn the sorted-slot row RED | caught |
| `check_roadmap_sorted.sh` on the aggregate | rc=0 |
| `check_roadmap_diff_additive.sh` (post-commit, reads HEAD) | rc=0 `added=1 lifecycle=0 reserialised=0 deleted=0` |

`PMAT-3296` lands after `PMAT-3229`, its sorted slot. This PR's own roadmap entry is the
first fragment, so the mechanism is dogfooded by the change that introduces it.

## Not in this PR — named, not hidden

- The **enforcement guard**: trailer ↔ filename parity, and refusing a PR that hand-edits
  `roadmap.yaml`. Until that lands this is opt-in, not structural.
- The post-merge regeneration workflow on `main`, and `make roadmap-aggregate`.
- Raising `max_entries_to_merge` (live ruleset `17836320` is `max_entries_to_build=3`,
  `max_entries_to_merge=1`). That is §2 and must follow this, not precede it.

Closes #3296

no-close: #2874 appears only inside quoted guard output (`PMAT-980 (#2874)`), the ticket
that authored the additive rule. This PR obeys that rule rather than resolving it — it is
cited as the authority for dropping the anchor-inlining commit, and stays open.

ont-delta: shape apr-roadmap-fragments-v1


noahgift:
Flagging before this lands, per the operator's ruling today: `scripts/lib/roadmap_fragments.py` is a **second writer** of `docs/roadmaps/roadmap.yaml`. APR-RELEASE-001 §1 (line 89) records `RoadmapServiceIo::save()` as the sole locked roadmap writer (#1364/#1368) — a fragment aggregator outside pmat is the same defect class in a different language.

pmat already has both halves: a per-ticket work store (`.pmat-work/{ledger.jsonl, <id>/contract.json}`) and a deterministic renderer (`pmat roadmap sync --dry-run`, content-hashed, 260 items on infra today). What it lacks is the path (`ROADMAP.yaml` at root vs `docs/roadmaps/roadmap.yaml`), a tracked store, and `work add` writing only the fragment — asked for in https://github.com/paiml/paiml-mcp-agent-toolkit/issues/1370. Suggest this PR keeps its contract (`contracts/apr-roadmap-fragments-v1.yaml`, the `entries/` layout, the parity gate from paiml/.github#73) and drops the Python aggregator in favour of `pmat roadmap sync --check` as the gate. infra#608 is scoped the same way: read-only checker until pmat lands the writer.

Issue #3315, opened 2026-09-15
0.68 scope: 14 anonymous obligations get ids, and all 25 kani harnesses cite one


noahgift:
0.68 scope (Qwen 3.5), and the first slice of #3314.

## Measurement at `origin/main` `10bd25400`

| contract | obligations | anonymous | kani | citing an id | dangling |
|---|---|---|---|---|---|
| `gated-delta-net-v1` | 5 | 0 | 7 | 2 | 5 cite prose |
| `qwen35-hybrid-forward-v1` | 7 | **7** | 9 | 0 | **2** |
| `qwen35-e2e-verification-v1` | 7 | **7** | 9 | 0 | **2** |

19 obligations, **14 anonymous**, and 4 kani harnesses pointing at ids that do not exist:

```
KANI-QHF-001  -> QHF-INV-001     (no such obligation)
KANI-QHF-002  -> QHF-INV-002     (no such obligation)
KANI-QE2E-001 -> QE2E-INV-001    (no such obligation)
KANI-QE2E-002 -> QE2E-ORD-001    (no such obligation)
```

`pv validate` reports `0 error(s), 0 warning(s)` for all three, and `pv proof-status` counts `9 kani` against `7 obligations` without noticing that two of the nine attach to nothing.

The remaining 21 harnesses reference obligations by their **`property` prose** — which resolves, but means a proof is bound to a sentence. Rewording the sentence silently unbinds the proof.

## What this ticket does

1. Give all 14 anonymous obligations ids, using the convention `gated-delta-net-v1` already established: `<PREFIX>-<TYPE>-<NNN>`, where `NNN` is the obligation's position in the list and `TYPE` comes from its own `type:` field (`invariant`→INV, `bound`→BND, `ordering`→ORD, `monotonicity`→MON, `conservation`→CON, `equivalence`→EQ).
2. Make **every** harness cite an id, including the 21 that cited prose.
3. Repoint the 4 dangling references at the obligation each harness's own `property` names.

Not in scope: the guard that keeps this from returning, and the other 810 contracts. Those are #3314.

PR #3320, opened 2026-09-15
contracts: 3,612 obligations had no id, so nothing could cite them

### LINKED ISSUE #3314: 824 of 876 contracts have anonymous obligations and 586 kani references point at nothing — pv validate says 0 errors
## Measurement

Scanned every `contracts/**/*.yaml` at `origin/main` `10bd25400` — 876 files carrying `proof_obligations` or `kani_harnesses`:

| | count | share |
|---|---|---|
| contracts with ≥1 **anonymous** obligation (no `id:`) | **824** | 94% |
| contracts with ≥1 **dangling** `kani_harnesses[].obligation` reference | **287** | 33% |
| dangling references in total | **586** | — |

A dangling reference is a `kani_harnesses[].obligation` value that matches neither a declared obligation `id` nor any obligation's `property` text — it names nothing in the file.

Worst offenders: `trueno-gpu/gemm-backward-tiled-v1.yaml` (10), `apr-architecture-schema-v1.yaml` (7), `cpp-type-preservation-v1.yaml` (7), `pmat/pmat-work-lifecycle-v1.yaml` (7), `apr-cli-v1.yaml` (6).

## Nothing reports it

`contracts/qwen35-hybrid-forward-v1.yaml` has **7 obligations, all anonymous**, and two kani harnesses pointing at `QHF-INV-001` / `QHF-INV-002`, which do not exist. `pv 0.65.2` says:

```
$ pv validate contracts/qwen35-hybrid-forward-v1.yaml
0 error(s), 0 warning(s)
Contract is valid.
```

`pv proof-status` is more honest but still does not mention the dangling refs:

```
qwen35-hybrid-forward-v1   L3   7 obligs  7 tests  9 kani  0 lean  0/6 bound
```

It counts 9 kani harnesses against 7 obligations without noticing that two of them attach to nothing.

## Why it matters

An obligation with no id cannot be cited — not by a harness, not by a test, not by a receipt, not by a commit. A kani harness pointing at a non-existent id is a proof attached to nothing, counted as if it were attached to something. Both directions inflate the same number: `9 kani` reads as coverage of `7 obligations` when at least two of those bindings are vacuous.

This reframes the Qwen 3.5 finding. "0/17 bound" in #3091 is not a Qwen problem; it is the visible corner of a repo-wide one. The Qwen 3.5 contracts are being fixed under the 0.68 scope (14 obligations across `qwen35-hybrid-forward-v1` and `qwen35-e2e-verification-v1`) — this ticket is the other 810.

## Fix

1. **A guard** — every `kani_harnesses[].obligation` must resolve to a declared obligation `id` in the same contract. This is a pure reference check, cheap, and needs no proving. Ships with a baseline of the 287, shrink-only, because it cannot be drained in one PR.
2. **`pv validate` must report it.** A contract whose harnesses cite ids that do not exist is not valid; saying `0 error(s)` is the same denominator problem as pmat#1369 (`pv validate` printing `0 errors` for a contract with 0 of 19 obligations bound).
3. **Obligation ids become mandatory** for new contracts, with the existing 824 baselined and drained by area.

Order matters: (1) is a mechanical check that can land now; (2) is upstream in `pv`; (3) is the ratchet that stops the class growing.

Refs #3091, pmat#1369

noahgift:
Ruling 3, first half. Naming lands here; `pv` strictness is the next PR and is this one's RED evidence.

## Why the order is inverted from the ruling

`cargo test -p aprender-contracts --test validate_contracts` is wired at `ci.yml:570` and walks every contract under `contracts/`. Arming `pv` first would red-line `main` on 3,612 rows. Same end state, no red main: **name first, strict second.**

## Measurement

**3,612 anonymous obligations across 823 contracts.** `pv validate` reported `0 error(s)` for every one.

## The generator, committed

`scripts/lib/obligation_ids.py` — with `--check` and `--selftest`.

```
<PREFIX>-<TYPE>-<NNN>
  NNN     the obligation's 1-based position
  TYPE    from its own `type:` field (28 distinct values; 79 obligations have no type → OBL)
  PREFIX  the prefix the file's own kani harnesses already cite, else the contract name's initials
```

The initials rule is not invented — it reproduces both prefixes a human chose here: `gated-delta-net-v1 → GDN`, `qwen35-hybrid-forward-v1 → QHF`.

## Constraints, each measured

**Never renames.** The 52 files whose obligations already carry ids are skipped whole — `REG-OB-001`, `PO-HEH-001`, `OBLIG-DATA-QUALITY-007-DEAD-OUTPUT-ROW`, none of which this rule would reproduce.

> already-named at origin/main: **52** · modified by this PR: **0**

**Globally unique from day one.** 111 initials-prefixes were shared across files. Ownership is resolved over the **whole corpus**, including the 52 untouched files — their existing ids claim `PO-` on 22 files and `OBLIG-` on 16. Owner = lexicographically first claimant; everyone else takes `sha256(own relative path)[:4]`. Keying the suffix to the file's **own** path rather than to the set of colliders is what keeps an id stable when an unrelated contract is added later.

| | |
|---|---|
| ids claimed by >1 file | **0** |
| ids duplicated within a file | **0** |
| distinct obligation ids | 3,773 |

**Idempotent.** A second pass names 0 and leaves the tree byte-identical. `--check` asserts it and exits non-zero otherwise.

## Two bugs the case table caught — both the same shape

A rule that silently did nothing.

1. **83 of 823 contracts indent their list items** (`  - type:`). A first cut matched only column 0, reported those files as named, and changed nothing. `--check` found **410 obligations still anonymous**. The indent is now read from the file's own first item, and the indented shape is a must-match row.
2. **Collision detection counted only computed prefixes**, so it could not see that an untouched file already owned one — minting `AL-BND-001` in two files. Caught by asserting global uniqueness, not by reading the code.

Neither was visible in review; both were visible to an assertion.

## Verification

| check | result |
|---|---|
| `--selftest` | **9/9** rows |
| `--check` | OK: every obligation carries an id |
| second pass on the named tree | byte-identical |
| YAML parse failures | **0** of 877 |
| `cargo test -p aprender-contracts --test validate_contracts` | **10 passed, 0 failed** |
| `pv validate` (sampled) | `0 error(s)` |

## Not in scope

The 14 obligations in `qwen35-hybrid-forward-v1` and `qwen35-e2e-verification-v1` — #3315/#3316 names those against the same files. Once it lands they are fully named and this generator skips them, so the exclusion decays to a no-op.

Refs #3314, #3315, pmat#1369

ont-delta: none this PR assigns identifiers to existing obligations and adds no obligation, equation or contract row

no-close: #3314 still needs its guard and the pv strictness pass; #3315 is closed by its own PR; pmat#1369 is an upstream pv defect this repo cannot close.


noahgift:
## Important: these ids are inert until the pv model carries them

Measured while scoping the follow-up PR. `ProofObligation` in `crates/aprender-contracts/src/schema/types.rs:523` has **no `id` field**:

```rust
#[derive(Debug, Clone, Default, Serialize, Deserialize)]
pub struct ProofObligation {
    #[serde(rename = "type", default)] pub obligation_type: ObligationType,
    #[serde(default, alias = "statement")] pub property: String,
    #[serde(default, alias = "verification")] pub formal: Option<String>,
    // tolerance, applies_to, lean, requires, applies_to_phase, parent_contract
}
```

No `deny_unknown_fields`, so `id:` is **silently dropped on parse**, and nothing in `validator.rs` reads an obligation id (the `.id` references there are `story.id`, a different type).

So as of this PR alone, the 3,612 ids are **decoration**: written to disk, ignored by every consumer. That is precisely the class this repo already has a rule about — *prove a consumer reads the key before counting it*, the same shape as `pv validate` accepting `entity:` by ignoring it.

**This PR is still correct and still necessary** — the ids have to exist on disk before anything can require them, and `--check` plus the idempotence contract are what make them trustworthy. But it must not be read as "obligations are now citable". They become citable in the next PR, which must:

1. add `pub id: Option<String>` to `ProofObligation` — *this* is what makes them real;
2. then error when it is `None`;
3. then error when a `kani_harnesses[].obligation` does not resolve to a declared id (586 such references across 287 contracts, #3314);
4. and report `N evaluated` rather than `0 errors`.

Step 1 is the one that turns this PR from a text edit into a fact.

## One hazard checked and cleared

`pv unlock` is the only command that writes a contract back (`serde_yaml::to_string`, `unlock.rs:21`). It round-trips through `serde_yaml::Value`, not the typed `Contract`, so unknown keys survive and it will **not** strip the ids. Verified rather than assumed — a typed round-trip there would have silently undone this PR file by file.

Refs #3314, pmat#1369

noahgift:
**This PR is what has been ejecting the merge queue today** (three ejections: 07:46:53, 09:05:20, 12:25:10, all by `github-merge-queue[bot]`), and it takes unrelated PRs batched behind it down with it. Details in #3362. One-line fix.

**Measured:** this PR edits `contracts/setfit-encoder-conformance-v1.yaml` (`+28 -14`) without regenerating the tolerances that are pinned to its hash.

```
contract @ this PR head 67359fead : fdfaeb5ce8fbd53d93e946169d5dca9dde352eeed06aced038ee20338b46f14d
CONTRACT_SHA256 in tolerances_generated.rs : 16a6591788a6c693ad3d08845a20267e31d4a86ee663310a943c841d9e7b2b93
```

`crates/aprender-core/tests/setfit_conformance.rs:1079` asserts those are equal, and the group build fails with `workspace-test` → `gate`.

**Why CI here is green.** The `Integration tests` step is gated `if: steps.tier.outputs.tier == 'full'`. PRs run the quick tier, so this target never runs on this page; `merge_group` runs the full tier. The PR is green and the group is red, every time.

**Fix, on this branch:**

```
APRENDER_REGEN_TOLERANCES=1 cargo test -p aprender-core --features setfit,conformance-fixtures \
  --test setfit_conformance conformance_tolerances_regenerate -- --ignored
```

then commit `crates/aprender-core/tests/setfit_conformance/tolerances_generated.rs`. Please don't re-arm auto-merge until that lands — I armed it at 12:04 before I understood this, which caused the third ejection. That one is mine.

PR #3041, opened 2026-09-07
feat(R-0b): backend resolution reads the registry — a forced accelerator never downgrades to cpu

noahgift:
Row **R-0b** (#3002, PP-066 #2873, claim 1) — backend resolution reads the registry. **Not armed**: `status: partial` (two cfg reads await the #3040 decomposition) and the base R-0a (#3004) is not on main yet — this PR is stacked and cannot merge before it.

## Ticket
PMAT-1073 (minted in #3025). Issue #3002 (inst:A). Blockers: R-0 (#3004), G-11 ✓, L0-1a (#3026).

## Claim
apr-cli turns a backend request into a `Selection` or a refusal by reading the registry, never `cfg!`; a forced accelerator (`--gpu`, `--backend cuda|wgpu`) that this host has not Ready refuses and is **never** downgraded to cpu.

## RED test
`crates/apr-cli/tests/backend_refusal_case_table.rs` (5 rows over R-0a's fixtures) — GREEN on the resolution; the falsifiers are the two mutations below.

## Acceptance (`.pr/R-0b/accept.sh`)
A2 case table 5/5, A3 guard self-test 3/3, A4 registry unit tests, A5 `pv validate` — GREEN. A1 (`--static` == 0) is **RED**: `dispatch.rs:173` and `lib.rs:230` still read cfg (residual; #3040).

## Mutation (measured this tree; CI pair owed once the PR runs)
- not-Ready branch of `resolve_in` → cpu: case table RED (`…never_downgrades`, `every_backend_value…`).
- restore `cfg!` at `accel.rs:28`: `check_backend_registry.sh --static` RED naming the line.

## Contract
`contracts/apr-backend-registry-v1.yaml` REG-OB-004 + REG-F-003 (extends R-0a's); `pv validate`: 0/0, valid.

## Quorum
Review-only (the design was judged with R-0a's 3-lane quorum, 2026-09-06); the resolution half is a diff review, owed before arm.

## Receipt
`docs/audits/impl-PMAT-1073-receipt.md` — `status: partial`.

## Writes
[...]

PR #3344, opened 2026-09-16
feat(registry): S3b — apr-cli call sites resolve through the backend registry, and a static guard keeps cfg! backend reads at zero

noahgift:
Second slice of the #3041 split, per `docs/audits/pr-3041-split-plan.md` (#3334). **Stacked on #3342 (S3a)** — until S3a merges, this PR's diff shows S3a's commits too.

S3a added the resolution library without changing any caller. **This is where behaviour moves**: twelve call sites stop reading `cfg!(feature = "cuda")` and start resolving through the registry, and a static guard keeps them that way.

## What changes, and what proves each one

| call site | before → after | covered by |
|---|---|---|
| `dispatch.rs` | inlined cuda check + `ensure_available` → `run_preflight` → `ensure_available_for` → `registry::announce` (FeatureDisabled 9 / BackendUnavailable 14, **never a cpu downgrade**) | 2 new `run_preflight` tests |
| `dispatch_analysis.rs` | `ensure_available(...)` → `ensure_available_for(...)` | **mutation-proven**: reverting this single site turns `every_accelerator_surface_calls_the_refusal` RED, naming `apr chat` |
| `accel.rs` | `cfg!(any(cuda,wgpu))` → `registry::build_has_accelerator()` | 3 new tests on the switched entry point, which had none |
| `serve/mod.rs` | cfg-gated guard → `ensure_accelerator_available_in(cfg, reg)`, fixture-driven | 4 tests, now over 4 fixture registries × 20 requests instead of the build's own cfg |
| `bench.rs`, `lib.rs`, `finetune.rs`, `handler_gpu_completion.rs`, `tests_offload_report_pp14.rs` | `cfg!` → `registry::compiled(kind)` / `compute_class()` | existing oracles, re-pointed at the registry |
| `handlers_include_01.rs`, `chat_load_tokenizers.rs`, `chat_generate_session_02.rs` | add `registry::parity_line(..)` at the CUDA loads (REG-15) | cuda-only paths — **named as gaps, not claimed as covered** |

## Two gates that could not fail, found and fixed here

- The three-surface source scan searched for `accel::ensure_available`, which is a **substring of the new `ensure_available_for`**. It passed identically before and after the very switch it was the only oracle for.
- The new guard's own `EXIT` trap referenced a function-`local` under `set -u`, so it exited 1 *after* printing `10/10 rows`.

## The static guard

[...]

Issue #3042, opened 2026-09-07
R-0b follow-up: a forced accelerator whose runtime attempt fails must refuse BEFORE generation in realizar (decompose try_wgpu_generate cog 49 / try_apr_wgpu_inference cog 69)


noahgift:
Measured on 2026-09-07 (PP-066 R-0b review, PR #3041): `apr run <qwen2.5-coder-0.5b q4_k_m> --gpu` on a non-cuda build printed `selected: wgpu device[0]=NVIDIA GeForce RTX 4090`, then `Backend: wgpu (Vulkan)`, then — only under `--verbose` — `Backend: CPU (wgpu unavailable: Format error: Unsupported quantization type 6 for WGPU dequant)`, ran the whole generation on CPU and exited 0. Without `--verbose` nothing said so.

Site: `crates/aprender-serve/src/infer/gguf_gpu_generate.rs:375-391` (`run_gguf_generate`): the `try_wgpu_generate` Err arm prints only `if config.verbose`, `log_cpu_backend` returns early when not verbose, and the function returns `Ok((tokens, false))`. The CUDA arm (`try_gguf_gpu_generate` Err → returned model → CPU) has the same shape.

R-0b (#3041) closes the honesty gap at the apr-cli boundary: after generation, a forced request whose `used_gpu == false` is refused with `BackendUnavailable` (exit 14) before any output, and a default fallback prints `selected: cpu (fallback …)` unconditionally. What remains is the WASTE: the CPU generation is spent before the refusal. The right fix is inside realizar — refuse in the Err arm when the request was forced (a `forced` flag reaching `run_gguf_generate`, e.g. a process-wide setter beside the parity tap, or a field on `InferenceConfig`), and print the `WGPU_FALLBACK_LOG_PREFIX` line unconditionally.

Blocked by the pre-commit complexity hook: `gguf_gpu_generate.rs` carries two pre-existing over-threshold functions, `try_wgpu_generate` (cognitive 49, line 65) and `try_apr_wgpu_inference` (69, line 447), and `infer/mod.rs` `include!`s the file (plus `inference_result.rs::validate_gpu_first_token`, 29). Any edit needs those decomposed in the same commit (`--no-verify` is forbidden). Same class as #3040.

noahgift:
[...]

PR #3349, opened 2026-09-16
feat(registry): S3c — the startup Selection reaches GET /v1/effective-config, and a forced accelerator that fell to cpu at RUNTIME is refused

noahgift:
Third slice of the #3041 split (`docs/audits/pr-3041-split-plan.md`, #3334). **Stacked on #3344 (S3b), which stacks on #3342 (S3a)** — until those merge, this PR's diff shows theirs too.

S3a added the library, S3b switched the call sites. S3c closes the loop at both ends: what the process resolved at **startup** becomes observable, and a forced accelerator that falls back at **runtime** is refused instead of quietly producing cpu output.

## Two adaptations, both to make a claim falsifiable

#3041 published the startup resolution and then asserted it **only from the route's own test, which supplied the value itself** — that cannot witness the serve gate publishing the wrong thing. So:

- `run()`'s mapping became a named seam, `serve::publish_backend_resolution`, read back field-for-field through `realizar::api::effective_config::backend_resolution()`. Publishing a constant `kind: "cpu"` turns that row RED; making `resolved_report()` return `None` turns the route row RED.
- The runtime-refusal wiring test lives **inside one of the files it scans**, so a literal needle would contain itself and pass with the call site deleted. Every needle is assembled by `format!` from pieces.

## Five mutations, each observed RED and restored byte-identically

| mutation | row that went red |
|---|---|
| delete the `after_generation` call in `run_entry` | surface row, naming `apr run` |
| restore chat's silent `on_cuda_load_error(…, false)` fallback | same row, naming `apr chat` |
| move the call below the `--stream` branch | the positional row ("before the first output branch") |
| publish a constant `kind: "cpu"` | the publish read-back row |
| `resolved_report() → None` | the effective-config route row |

## Scope discipline

Only the two `5b5116d6c` hunks were taken for `chat_generate_session_02.rs`; #3041's blob for that file also **reverts** main's later `ShardedSafeTensors` arm, which the plan marks "stale side, take main". The contract's effective-config and `after_generation` clauses are restored with obligation ids unchanged, and the mutation text was rewritten to the mutations actually observed here rather than the ones #3041 claimed.

No CI wiring is owed — both proving targets are `--lib`, already in `workspace-test`. No `surface_audit.csv` row cites any file this slice touches (0 of 838), so the ledger is untouched and G2.1 is unchanged.

Gate at the head commit, every leg measured with an unpiped exit code: `apr-cli --lib` 7255 passed / 0 failed, `aprender-serve --lib effective_config` 29 passed, clippy `-D warnings` clean on both crates, fmt clean, dogfood coverage PASS, `pv validate` 0 errors (built from the worktree — the PATH `pv` is two minor versions stale).

Refs #3041

no-close: #3041 stays open until S4 lands; it is then closed with a pointer, per the plan. The PRE-generation refusal stays a non-goal here and is #3042.

ont-delta: resolves REG-OB-004 — the effective-config half (the served process reports its startup Selection as `resolved`, and a forced accelerator that fell to cpu at runtime is refused before any output), discharged by REG-F-003's new effective-config and after_generation legs; obligation ids unchanged.


noahgift:

No activity since 2026-09-17 13:44Z; **2 release trains have been cut since** (v0.68.2, v0.68.1). Age is the only input to this sweep — it is not a judgement on the work, and a train label such as `pp-066/carry-*` is not an exemption.

noahgift:

noahgift:
Triage #4158 (needs-owner): DIRTY against main, last commit 8 days ago, no owner named: needs a rebase by an owner. The cop assigns an owner from the 0.70 scope.

noahgift:
closed by operator 1-day PR rule; reopen/rebase to resume.

No live owner session, red CI on the current head, not carried by main or any open integration PR. Linked issue: #3041 (CLOSED, 0.68.0) — S3c; 41% of its lines on main, the rest unlanded. Linked issues are not touched by this close.

Branch is NOT deleted: `PMAT-3041-s3c-effective-config` @ `dde5d35c4`.

PR #3472, opened 2026-09-17
feat(ont): ONT-2b — Σ with entity_types and extractors, and a symbol-level check on formal: (PMAT-3471)

### LINKED ISSUE #3471: ONT-2b: Σ — contracts/ontology.yaml with entity_types and extractors; symbol-level check on formal:
ONT-001 v4.4 §5 row **ONT-2b** — "Σ — `ontology.yaml` with `entity_types` and `extractors`; symbol-level check on `formal:`" · `depends_on ONT-1, ONT-6` (both merged and bound: ONT-1 `2fb79ff0b`, ONT-6 `2911bdcde`; carrier infra#660) · K̂ 90.

This is the first eligible row in §5 order now that ONT-6 is bound (infra `ac94de5a7`). It unblocks ONT-2c, ONT-4, ONT-4b, ONT-4c and ONT-4d, every one of which lists ONT-2b in `depends_on`.

## What the spec asks (§5 ONT-2b, verbatim)

- **RED:** undeclared role → exit 1; undeclared symbol in `formal:` without `prose` → exit 1; `entity.type` not in Σ `entity_types` → exit 1; an `entity_types` entry naming no extractor → exit 3; Σ key with no reader → exit 3; `not_expressible` without `reader` → exit 3; `extractors[]` entry without `reader` → exit 3.
- **Change:** Σ per §4.1; symbol-level parser only; `formal_prose` and `unanchored_but_bindable` baselines recorded.
- **Mutation:** drop a role the corpus uses → RED.
- **probe:** `tracked contracts/ontology.yaml && present '^entity_types:' … && present '^extractors:' … && "$PV" lint contracts/ --gate sigma --format json > s.json && jq -e '.gate=="sigma" and .verdict=="Pass"' && merged ONT-2b`

§4.1 gives Σ's shape: `concepts{}`, `roles{domain, range, symmetric?, acyclic?}`, `symbols[]`, `worlds{}`, `agents[]`, `entity_types[{name, extractor, implemented}]`, `extractors[{name, reader}]`, `not_expressible[{key, reader}]` (`subsumes[]` belongs to ONT-4d).

## The corpus this must pass over, measured on main (`2911bdcde`)

| fact | value |
|---|---|
| contracts | 1792 (`census.json.n_files`) |
| `by_anchoring` | `unanchored: 1792`, class 0, instance 0 — **no contract carries an `entity:` block yet** |
| `by_entity_type` | `{}` — empty for the same reason |
| files carrying `formal:` | **489** |
| `formal:` expressions | **2330** |
| `by_kind` | kernel 362, schema 766, registry 519, pattern 79, … |

So the two exit-1 rules that read contract content have very different weight today: `entity.type not in Σ` is vacuous until contracts gain `entity:` blocks (0 of 1792), while the symbol rule immediately faces 2330 expressions written over a wide vocabulary (`∩ ∅ ⊆ ∀ ∧ ≥`, `isFinite(...)`, `modifies(...)`, `preserves(...)`, `len(...)`, `‖·‖`, `ε`, `mod`).

That is what `formal_prose` is for: the expressions that are not symbol-checkable are COUNTED in a shrink-only baseline rather than fixed in this row — the same ratchet shape as `contracts_without_entity`. The gate has to come out `Pass` on today's corpus (the probe demands it) without that pass being vacuous.

## Open design points for the plan grill

1. **What makes an expression "prose"** — an explicit per-entry marker, or derived? The RED line says "undeclared symbol … **without `prose`**", which reads as an explicit opt-out; the baseline then counts them and may only fall.
2. **Symbol granularity.** "symbol-level parser only" (not a term parser): identifiers and operators, not arity or types. Where exactly is the boundary drawn, and what is deliberately NOT checked?
3. **Where the two exit-3 classes live.** `entity_types` entry naming no extractor, a Σ key with no reader, `not_expressible`/`extractors[]` without `reader` are all Σ-is-malformed errors (exit 3, `error:`), distinct from a corpus violation (exit 1, `reject:`) — ONT-6's vocabulary already draws that line.
4. **Which roles Σ declares** and how "undeclared role → exit 1" can bite while `relations{}` is not yet populated in any contract.

## Landing

ONT-001 §2 protocol: this issue → `pmat work add --github-issue N -t kind:code,orch:fable,orch-basis:state` → branch `PMAT-N-ont-2b-<slug>` → RED → GREEN → mutation → quorum → `pmat-merge --auto --squash`, then an Opus `kind:triage` carrier binds the row in infra's ONT-001 ledger.

noahgift:
Closes #3471 · ONT-001 v4.4 row **ONT-2b** (paiml/infra `docs/specifications/paiml-ontology.md` §5, §4.1, R-8) · ticket PMAT-3471 · receipt `docs/audits/impl-PMAT-3471-receipt.md`

ont-delta: type ONT-2b — `contracts/ontology.yaml` declares 7 entity types and 7 extractors, and `pv lint --gate sigma` is the reader that makes them load-bearing

ONT-6 is bound (infra#660), so this was the **first eligible row in §5 selector order**. It unblocks ONT-2c, ONT-4, ONT-4b, ONT-4c and ONT-4d, each of which lists ONT-2b in `depends_on`.

## What lands

- **Σ — `contracts/ontology.yaml`** (new): 9 concepts, 6 roles, **66 symbols**, 1 world, 3 agents, 7 `entity_types` each naming a declared extractor (`implemented: false` — honest until ONT-4b/4c build them), 3 `not_expressible` keys, and a `readers:` map claiming every populated key.
- **`ontology/sigma.rs`** (new): the §4.1 types, the loader, and the row's four malformed-Σ classes as one typed `SigmaError` — each **exit 3 `error:`**, because a malformed declaration is not the corpus's fault.
- **`lint/sigma_gate.rs`** (new): PV-ONT-001 `entity.type ∉ Σ`, PV-ONT-002 undeclared role, PV-ONT-003 undeclared operator glyph without `prose: true`, PV-ONT-004 a **rise** in the `formal_prose` debt. It reads RAW YAML: `entity:` and `relations:` are §4.2 additive keys the `Contract` struct does not carry, and serde drops what it does not know.
- **`lint/sigma_symbols.rs`** (new): the symbol scan and the explicit opt-out.
- **`pv lint --gate <name>`**: runs ONE gate and reports only it, mapping the verdict through ONT-6's lattice — Pass 0 · corpus violation 1 `reject:` · no Σ 2 `decline:` (R-2) · malformed Σ 3 `error:`. Unknown gate names are refused, naming what `--gate` runs alone.
- **`sigma` is armed** (R-8: computed everywhere, armed per repo): it runs as the 10th gate of every `pv lint`, and `contracts/lint-baseline.json` arms it — nine armed gates.
- **`contracts/ont-sigma-v1.yaml`** (new): `kind: pattern` deliberately — Σ is a declaration and a gate over it, not a math kernel, so `kind: kernel` would claim a proof ladder this row does not climb.

## What the rule actually checks, stated as what it does NOT

Σ owns the **operator vocabulary**. Applied identifiers are **not** checked here: 1325 of the corpus's 2303 `formal:` expressions call a name Σ does not declare (`encode`, `is_ok`, `Err`, `predict`, `softmax`) and those are **code** — ONT-3a's `extract:code` resolves them fail-closed. Grammar, parenthesis balance, arity, types, scope and truth are not checked, and a bare variable is not a symbol.

The vocabulary was **measured, then declared**: 79 expressions in 51 files carried an operator Σ lacked, and all but two glyphs were real notation (`² √ ∞ ⊥ ∘ ∋ ∌ ⊇ ⊄ − ↔ ⟺ ≺ ± ⟨ ⟩ ∇ ∂ ⌈ ⌉ ′ ₀ ₁ ₂ ⋃ ← ̄`). The two that were not — `—` and `§` — are prose punctuation, so the **19 entries carrying them in 14 contracts now say `prose: true`**. The opt-out is exercised on the real corpus, not only in a fixture.

`formal_prose` is recorded at the measured **1464 of 2308** and enforced by the gate that measures it (one implementation, and it is armed).

## Both rules that cannot fire today have fixtures

0 of 1792 contracts carry `entity:` and 0 carry `relations:`, so `tests/fixtures/ont/sigma-entity-type-unknown/` and `…/sigma-undeclared-role/` are the only witnesses those rules can fire at all — plus `sigma-ok`, `sigma-malformed`, `sigma-absent` and `sigma-prose-ratchet`.

## Evidence

| check | result |
|---|---|
| `pv lint contracts/ --gate sigma` | **Pass** — 1793 contracts, 2308 `formal:` expressions, 0 violations, `formal_prose` 1464 |
| `pv lint contracts/` (armed, 9 gates) | rc **0**; validate 1793 contracts / 0 errors |
| `cargo test -p aprender-contracts --lib` | **1579 passed** |
| `cargo test -p aprender-contracts-cli` | **249 passed** |
| clippy `-D warnings` | clean |
| `make contracts` | rc 0 (lint, census diff, README, provenance, 1579 engine tests) |
| infra `scripts/ont/done_when/ONT-2b.sh` | every conjunct rc 0 except `merged ONT-2b`, which the carrier binds |

Mutations, each run on the committed tree and restored: drop the role the corpus uses → 2 gate tests RED (the row's named mutation) · unmark one `prose: true` entry → corpus `Fail`, 1 violation · raise the prose debt above its baseline → PV-ONT-004.

## Two defects I caused, and how each was caught

1. **Σ was parsed as a member of its own corpus** — `pv lint contracts/` went 1792/0 → **1793/1**. Caught by the test asserting `--gate validate` still passes. Σ now sits in `NON_CONTRACT_FILENAMES` beside `binding.yaml`.
2. **A scripted edit that counted right and was wrong anyway** — the first `prose: true` inserter wrote the marker INSIDE multi-line quoted scalars (`formal: '… yields prose: true Commands::Serve{…}'`). The count was 14 both times. Reverted and redone scalar-aware, and the script now **asserts the result**: every touched file re-parses, every `prose` is a real boolean, and no `formal:` text swallowed the marker.

## Review

Plan grill (3 lanes, isolated clones, tree witness verified 3/3): **3/3 non-PASS**, and every finding was re-measured before being folded into plan v2 — the probe path that does not exist in this repo, a `| jq -e` acceptance that exits 0 on empty input, the row's named mutation being unfireable, and `formal_prose` having no ratchet. Diff quorum: **3/3 PASS** on `904cfb1f9`, `diff_sha256` re-derived after the mid-run fetch.

The delegate's isolation scan found that this review **wrote into the shared checkout** — three tracked files rewritten and restored byte-identical, 4305 `target/` files — which porcelain cannot see. Filed as paiml/paiml-implement#245. Every number above is the orchestrator's own re-run, not a lane's claim.


noahgift:
**What the two quorum rounds established, and what they did not.** Round 1 agreed 3/3 on `904cfb1f9`; it lapsed when the diff moved (the README-guard filter, the `equations` block, the origin/main merge). Round 2 agrees 3/3 on `90159253c`, artifact committed, `receipt-lint` rc 0, three distinct measured gemini models against a claude author.

Recorded because it is measurable: in round 2 every lane returned `findings: []` with `num_turns: 1` over a 155 KB file-mode prompt — **zero tool calls** — while two of them wrote that the receipt's claims were "independently verified". Those numbers (10/10, 15/15, 8/8, clippy clean) came from the receipt inside their own prompt. And **no lane mentioned any of the three changes that made round 1 lapse**: 0/3 for `check_readme_claims`, 0/3 for the equations/444 ceiling, 0/3 for the roadmap aggregate. The delegate anchor-checked both code changes itself and found them sound.

So the acceptance evidence here is the orchestrator's own re-runs plus CI: `ci / gate`, `workspace-test`, `guard-cargo` and `guard-tree` are all green on `cad0c77ab`; only the advisory `present` is red. Separately, the review again wrote into the shared checkout (one tracked file — this artifact — plus 655 `target/` files), which is filed as paiml/paiml-implement#245.

PR #3497, opened 2026-09-18
fix(guard): R-2 close-issue guard now fails closed on the #3400 landmine

alfredodeza:
Fixes #3400.

## What was happening

GitHub's closing-keyword parser reads `close:` immediately inside `no-close:`
— the hyphen counts as a word boundary to it, so the negating prefix is
invisible. `check_pr_closes_issue.sh` (the R-2 guard, wired into `ci.yml`)
had the identical blind spot in its own `CLOSE_RE`: a body line starting
with the negating marker and naming an issue still parsed as a real closing
reference, so the guard printed PASS instead of catching the danger.

Measured cost so far: an issue tracked in this repo's Qwen3.5 CPU work was
closed twice by this mechanism before #3400 was filed. #3400's own remedy
list asked for a guard change and nothing else landed on it. Today the same
class of line, on PR #3484, closed both the GPU sub-ticket and its parent
epic a third time — despite the body explicitly carrying the negating marker
for each.

## What changed

`scripts/check_pr_closes_issue.sh`:
- New `NOCLOSE_LANDMINE_RE`: any hyphen-prefixed closing keyword (`fix`,
  `close`, `resolve`, and their variants) immediately followed by a colon and
  an issue reference. Checked first, unconditionally — a landmine anywhere in
  the body fails the whole check, even alongside an otherwise-correct closing
  line elsewhere.
- The sanctioned "this stays open, here's why" marker is now `keep-open:`
  instead of the negating marker — a string neither GitHub's parser nor this
  guard's own `CLOSE_RE` reads as a closing keyword.
- Self-test: 9 → 19 cases. Three new RED rows prove the landmine is caught,
  including a mixed body that also correctly closes a different issue (the
  landmine still fails the whole body).

## Also fixed by hand (not in this diff)

Two open PRs (#3488, #3494) still carried the negating-marker line naming the
epic; edited their bodies in place to the new marker before they could merge
and reopen the same hole. The epic and the GPU sub-ticket, closed again
today by this mechanism, are reopened.

keep-open: #3477, #3090 — this PR lands the guard only; GPU device support
itself is unstarted, separate work, and stays open on milestone 0.68.2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3477, opened 2026-09-18
EPIC: release train 0.68.2 (interrupt) — Qwen3 CPU+GPU verification, P0


alfredodeza:
**Release train 0.68.2 (interrupt)** — Qwen3 CPU + GPU support, P0

**Why this exists / relation to the 06x cadence (`docs/specifications/06x-release-schedule.md`):** that spec's only
stated Qwen priorities are F/G — Qwen2.5-Coder-7B parity and its fine-tune story. Qwen3 was never an operator-declared
priority row, so under §1.1 ("scope that is not on the platform rides the next train") every Qwen3 ticket has drifted
forward release over release with no floor. This milestone is an explicit operator override of that drift, inserted
ahead of 0.69.0/0.70.0, per the 2026-09-18 decision: **the next release ships Qwen3 CPU+GPU verification, full stop.**

**Investigation — how Qwen3 support ended up NOT in v0.68.1 (summary; full trail in ticket comments):**

1. **Dense Qwen3 (`Qwen3ForCausalLM`) has no current proof it runs at all.** The only historical verifications
   (#479, #308, closed 2026-02/03) predate the entire monorepo consolidation and multiple forward-path bug classes
   found since (Q5_K layout fix, 8 Qwen3.5-forward defects). Nothing re-checked it. This wasn't "deprioritized" —
   it was **undiscovered**: #3413 (P0) was filed 2026-09-17, the day *after* the 0.68.0 train closed, while
   investigating the adjacent Qwen3.5 ticket. It could not have ridden 0.68.1 because the gap wasn't known to exist
   until 0.68.0 had already shipped. It landed on milestone 0.69.0 by default (the only open train when filed), not
   by any prioritization decision.

2. **Qwen3.5 CPU (#3091) DID get an explicit no-slip commitment for 0.68** (operator ruling 2026-09-15, row 68-Q1
   of the 0.68.0 epic #3079) and a PR landed (#3114/#3091 merge). But the fix was incomplete: `tensor_byte_size`
   only handles 11 quant-type ids, and real downloaded quantizations (UD-IQ2_XXS, UD-IQ4_XS) crash with an
   unhandled-type refusal. The ticket's true state was hidden by **three wrongful auto-closes** — GitHub misparsed
   the literal string "no-close: #3091" in unrelated PR bodies as a closing keyword (tracked separately as #3400,
   a real GitHub/tooling defect). Operator triage caught this 2026-09-17, reopened it, and rescoped it to depend on
[...]

Issue #3400, opened 2026-09-16
GitHub parses 'no-close: #N' as 'close: #N' — #3091 was closed twice by PRs whose bodies said no-close (D0 of the pmat 3.41.0 run); #3351 carries the same line for #3347


noahgift:
Root cause of #3091 being closed on 2026-09-15T17:49:05Z and 2026-09-16T06:50:47Z, found by the pmat 3.41.0 release run (D0).

## Measured

- Issue events: both `closed` events have `commit_id: null`, `performed_via_github_app: null`, actor the merger.
- GraphQL `ClosedEvent.closer`: `PullRequest 3323` and `PullRequest 3340`.
- GraphQL `closingIssuesReferences`: #3340 → [3091]; #3322 → [3091, 3321]; #3323 → [3091, 3303]. GitHub lists 3091 as a closing reference of each, although each body used `Refs #3091` and a line beginning `no-close: #3091 …`.
- No custom closer exists: pmat's `close_issue` has zero call sites, `pmat work sync` has no GitHub-side close action, paiml-implement's `mutate.sh close` refuses without a quorum citation and was never invoked for 3091, no workflow here carries a token other than `GITHUB_TOKEN`.

## Mechanism

GitHub's closing-keyword parser matches `close: #3091` inside `no-close: #3091` — the hyphen is a word boundary, so the negating prefix is invisible to it. The convention chosen to say "this PR does not close N" is, to the parser, the one word guaranteed to close N.

## Required

1. Stop writing `no-close:` (or any `<prefix>-close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved: #N`) in PR bodies and commit messages. A marker GitHub cannot parse, e.g. `keep-open: #N`, says the same thing.
2. PR #3351 (open) has `no-close: #3347 stays open …` in its body; on merge it will close #3347. Edit the line before merging.
3. pmat will ship a commit-msg / `make gate` refusal for this pattern and a `pmat-issue-closure-v1` contract in 3.41.0; aprender's own guard scripts can carry the same regex until the pin moves.

alfredodeza:
Recurred a third time today, 2026-09-18T18:13:34Z: PR #3484 merged carrying `no-close: #3090` and `no-close: #3477` and closed both anyway — same mechanism as the two #3091 closes this ticket documents. Two more open PRs (#3488, #3494) were found carrying the same marker for #3477 and hand-fixed before they could merge.

Required item 3 (guard fix) now landed: PR #3497.

[...]

PR #3499, opened 2026-09-18
fix(guard): #3400 landmine regex requires a colon CLOSE_RE doesn't require

alfredodeza:
Follow-up to #3497. Found by that PR's own independent review
(`evidence/pr-review/3497/`), finding 1: `NOCLOSE_LANDMINE_RE` required a
literal colon while `CLOSE_RE` treats the colon as optional (`:?`), so a
colon-less body bypassed the new check entirely and read as a *valid*
closing reference instead of the landmine it actually was — the same bug
class in a spelling one edit away from the one #3497 fixed.

## What changed

- `scripts/check_pr_closes_issue.sh`: made the colon optional in
  `NOCLOSE_LANDMINE_RE`, matching `CLOSE_RE` exactly. Self-test 19 → 20
  cases; the new row is a body reading a hyphenated keyword, a space, then
  the issue reference with no colon between them — it now fails as intended.
- `docs/specifications/APR-RELEASE-001-train-and-build-kaizen.md`: the R-2
  predicate row still documented the retired negating marker as sanctioned.
  Updated to name the replacement and the reason.

## Verification

```
$ bash scripts/check_pr_closes_issue.sh --self-test
self-test OK: 20 case(s).
```

And the exact shape the review found, confirmed caught:

```
$ printf 'no-fix (issue) #123 tracked elsewhere' | bash scripts/check_pr_closes_issue.sh
FAIL: hyphen-prefixed closing keyword (#3400 landmine) -- ...
```

keep-open: #3477, #3090 — unrelated; guard-correctness follow-up only, not
GPU/CPU Qwen3 work.


noahgift:
Triage from the release babysit session (aprender-ba), so this does not sit red silently while the 0.68.2 train runs:

**The red is CODE, not env.** `guard-tree` (run 35389817458, 2026-09-18T20:11:59Z) fails on the PR **body**, not the diff:

```
FAIL  no `ont-delta:` line. §11.1: a sweep PR closes with a delta,
FAILED: §11.1 — a sweep that ends in prose and nothing else is not finished.
```

`gate` is red only because `guard-tree` is. The body has no `ont-delta:` line today.

Two things to know when fixing it: (1) the body guard reads the **frozen** `pull_request` event payload, so editing the body alone never re-runs it — a push (even an empty commit) or close/reopen is needed after the edit; (2) the branch is 15 commits behind main, and main has since landed #3520 (guard_tree reads GitHub with the job token), so the next run also picks up that fix.

Auto-merge is already armed on 672ea4cf1; once `guard-tree` is green it merges on its own. I have not touched the branch.

noahgift:
Update: the §11.1 predicate was the reach-too-far, not this PR. #3536 (PMAT-3535) merged at 20:04Z: the spec now names only the release spec (`docs/specifications/APR-RELEASE-001-*.md`) as a sweep sink, a `*-MASTER.md` design spec is the finding's home and not a sink, and `check_pr_ont_delta.sh` got the predicate self-test it never had. **No body change needed** — a rebase/update onto main (ce343c76d or later) re-runs `guard-tree` with the fixed guard.

noahgift:
Correction to my note above — for **this** PR a body line **is** still required. The re-run on 4921df880 (job 105959120528, with the fixed guard from #3536) still fails §11.1 because this branch edits `docs/specifications/APR-RELEASE-001-train-and-build-kaizen.md`, and the **release spec stays a sweep sink** by design (the amendment only excluded `*-MASTER.md` design specs, which is why #3420/#3424 clear on a rebase and this one does not).

Fix: add one line to the PR body, e.g.

```
ont-delta: none — guard regex fix (#3400 landmine CLOSE_RE), no ontology finding
```

then push (an empty commit is enough; the guard reads the frozen event payload). Everything else on the run is green and auto-merge is armed, so it merges on its own after that. Sorry for the false "no body change needed".

noahgift:

noahgift:
closed by operator 1-day PR rule; reopen/rebase to resume.

No live owner session, red CI on the current head, not carried by main or any open integration PR. Linked issue: no issue link in the body; follow-up to #3497 / #3400 (landmine regex colon). Linked issues are not touched by this close.

Branch is NOT deleted: `fix/landmine-colon-optional` @ `28aef690f`.

